Issue #44, opened 2019-08-16
Do not hard-code OTP Path for code navigation


robertoaloi:

robertoaloi:
Also see #60.

PR #46, opened 2019-08-19
[#44] Resolve OTP Path dynamically

jfacorro:
This is one way of doing it. Not sure if it's the best way of doing it.

### DIFF
--- a/src/erlang_ls_code_navigation.erl
+++ b/src/erlang_ls_code_navigation.erl
@@ -18,11 +18,6 @@
 -include("erlang_ls.hrl").
 
 %%==============================================================================
-%% Macros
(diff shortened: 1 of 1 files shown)


robertoaloi:
The problem with this approach is that the OTP library would be the ones used by the server. This could be problematic if multiple projects use different versions of OTP. My idea was for users to be able to configure projects (a bit like in EDTS) and to therefore extract the OTP path from the initialization options or something like that.

robertoaloi:
Said that, the provided solution is still an improvement over the hard-coded path. Cherry-picked your commit, switched to use `code:root_dir/0` and added the "include" dirs, too. See #47

PR #48, opened 2019-08-26
Fix bug in OTP path resolution

jfacorro:


robertoaloi:
D'oh! I forgot to push that! Thanks for the help!

PR #50, opened 2019-08-27
Include lib/*/src and lib/*/includes for the app as well as for OTP

jfacorro:


src/erlang_ls_code_navigation.erl
```diff
@@ -123,21 +123,27 @@ definition({type_application, {Type, _}}) ->
...
   [ filename:join([RootPath, "src"])
   , filename:join([RootPath, "include"])
+  | paths(RootPath)
   ].
 
+-spec paths(binary() | string()) -> [string()].
+paths(Root) when is_binary(Root) ->
+  paths(erlang:binary_to_list(Root));
+paths(Root) ->
+  Sources = filename:join([Root, "lib", "*", "src"]),
```
robertoaloi: I am not sure I would go this way. I would rather start working on the "workspaces" that the client can configure. That would lead to a much cleaner and powerful implementation. In Emacs, workspaces could also be integrated with `projectile` somehow.

jfacorro: I think this is independent of workspaces.

robertoaloi: Well, not every application has a `lib` directories. Standard apps have only `src`, old rebar apps have `deps`, rebar3 ones have `_build/...`, umbrella apps have a `apps` folder, some legacy apps have `lib` and so on. I would like the user to be able to specify additional paths for a given application. It should also be possible to mark an app as a `rebar3` or `erlang.mk` app and then it should be possible to automatically infer paths from the `rebar.config` or equivalent.

PR #51, opened 2019-08-27
Handle usage of macros with arguments

jfacorro:


src/erlang_ls_tree.erl
```diff
@@ -192,7 +192,10 @@ points_of_interest(Tree, function, Extra) ->
   {F, A} = erl_syntax_lib:analyze_function(Tree),
   [erlang_ls_poi:poi(Tree, {function, {F, A}}, Extra)];
 points_of_interest(Tree, macro, Extra) ->
-  Macro = erl_syntax:variable_name(erl_syntax:macro_name(Tree)),
+  Macro = case erl_syntax:macro_name(Tree) of
```
robertoaloi: Would it be possible not to rely on the internal representation, but on some `erl_syntax` API, instead? E.g. `Name = erl_syntax:macro_name(Tree)` and `case erl_syntax:type(Name) of atom -> ...`. Also, we should add such a macro as part of the test file, so  we don't accidentally introduce a regression later on.

robertoaloi:
@jfacorro Please look at #52 for an alternative implementation.

jfacorro:
I'm going to close this PR in favor of #52. 😄

Issue #19, opened 2019-07-14
Shutdown Request


robertoaloi:
The shutdown request is sent from the client to the server. It asks the server to shut down, but to not exit (otherwise the response might not be delivered correctly to the client). There is a separate exit notification that asks the server to exit.  Clients must not send any notifications other than exit or requests to a server to which they have sent a shutdown requests. If a server receives requests after a shutdown request those requests should be errored with `InvalidRequest`.

Request:

* method: `shutdown`
* params: `void`

Response:

* result: `null`
* error: code and message set in case an exception happens during shutdown request.

PR #55, opened 2019-08-30
[#19] Basic implementation of shutdown and exit requests

jfacorro:
#19. This enables restarting the server from an LSP client (e.g. in `lsp-mode` with the command `lsp-workspace-restart`).


robertoaloi:
Thanks!

PR #74, opened 2019-09-30
List all subdirs

jfacorro:


src/erlang_ls_code_navigation.erl
```diff
@@ -121,34 +121,60 @@ definition({type_application, {Type, _}}) ->
...
+subdirs(Path) ->
+  subdirs(Path, []).
+
+-spec subdirs(string(), [string()]) -> [string()].
+subdirs(Path, Subdirs) ->
+  case file:list_dir(Path) of
+    {ok, Files} ->
+      Fold = fun(F, Acc) ->
+                 FullPath = filename:join([Path, F]),
+                 case filelib:is_dir(FullPath) of
```
robertoaloi: Travis doesn't like nested cases :)

jfacorro: Fixed now 😄

robertoaloi:
@jfacorro That could very well be the case. It's totally fine, we can still improve things later if we need to.

PR #78, opened 2019-10-06
Move protocol methods implementation to its own module

jfacorro:


src/erlang_ls_protocol_impl.erl
```diff
@@ -0,0 +1,196 @@
...
+            },
+  {response, Result}.
+
+-spec config_path(map()) -> erlang_ls_uri:path().
+config_path(#{<<"erlang">> := #{<<"config_path">> := ConfigPath}}) ->
+  ConfigPath;
+config_path(_) ->
+  ?DEFAULT_CONFIG_PATH.
+
+-spec consult_config(erlang_ls_uri:path()) -> map().
```
robertoaloi: Note (outside of the scope for this PR): we could move config handling to its own separate module.

src/erlang_ls_server.erl
```diff
@@ -126,92 +121,18 @@ handle_request(Socket, Request) ->
       send_notification(Socket, M, P)
   end.
 
+%% @doc Dispatch the handling of the method to erlang_ls_protocol_impl
 -spec handle_method(binary(), map()) ->
+  {response, map() | null} | {} | {notification, binary(), map()}.
+handle_method(Method, Params) ->
+  Function = method_to_function_name(Method),
+  try erlang_ls_protocol_impl:Function(Params)
```
robertoaloi: Nice, thanks. FYI I am working on a branch which will end up slightly modifying things around this area. The idea is to define a "provider" behaviour that different providers will implement. Providers will then be responsible for handling different sections of the protocol.

jfacorro: Do all providers have the same "interface"?

Issue #84, opened 2019-10-09
Implement POI Database


robertoaloi:
The list of Points of Interest is calculated multiple times. Instead, we should use a DB (DETS?) to be able to store pre-calculated POIs for unchanged files.

PR #88, opened 2019-10-13
[#84] points of interest table

jfacorro:
#84


src/erlang_ls_parser.erl
```diff
@@ -60,8 +60,8 @@ parse_extra(IoDevice, Extra, StartLocation) ->
       end;
     {eof, _} ->
       {ok, Extra};
-    {error, Error} ->
-      {error, Error}
+    {error, ErrorInfo, EndLocation} ->
```
jfacorro: This change is based on the error tuple that is returned from the call to `io:scan_erl_form/3`.

src/erlang_ls_tree.erl
```diff
@@ -101,7 +101,7 @@ points_of_interest(Tree, Extra) ->
   try points_of_interest(Tree, Type, Extra)
   catch
     Class:Reason:Stacktrace ->
-      lager:warning( "Could not analyze tree: ~p:~p ~P"
+      lager:warning( "Could not analyze tree: ~p:~p ~p"
```
jfacorro: `~P` seems to be an invalid format argument.

test/prop_statem.erl
```diff
@@ -97,7 +97,7 @@ did_open(Uri, LanguageId, Version, Text) ->
   erlang_ls_client:did_open(Uri, LanguageId, Version, Text).
 
 did_open_args(_S) ->
-  [erlang_ls_proper_gen:uri(), <<"erlang">>, 0, utf8()].
+  [erlang_ls_proper_gen:uri(), <<"erlang">>, 0, erlang_ls_proper_gen:tokens()].
```
jfacorro: Otherwise the parsing of the file fails.

src/erlang_ls_protocol_impl.erl
```diff
@@ -124,8 +124,9 @@ textdocument_didchange(Params) ->
   case ContentChanges of
     []                      -> ok;
     [#{<<"text">> := Text}] ->
-      {ok, Buffer} = erlang_ls_buffer_server:get_buffer(Uri),
-      ok = erlang_ls_buffer:set_text(Buffer, Text)
+      {ok, Buffer0} = erlang_ls_db:find(erlang_ls_files, Uri),
```
jfacorro: These three calls could be simplified with `erlang_ls_buffer:update(Uri, TexT)` if `erlang_ls_buffer` was the one writing to the table and in general the one that know what table should it use.

src/erlang_ls_tree.erl
robertoaloi: The depth argument was missing.

Issue #64, opened 2019-09-23
Add support for references


robertoaloi:

jfacorro:
Can we close this one?

robertoaloi:
Yes, specific missing functionalities will require their own individual tickets.

PR #91, opened 2019-10-17
[#64] References provider

jfacorro:
#64 

This PR creates a new references provider that uses a new indexing feature for the application's files. 

The supervision tree within this PR looks like this:

* erlang_ls_sup
  * erlang_ls_config
  * erlang_ls_index_sup
    * erlang_ls_index children (e.g. erlang_ls_references_index)
  * erlang_ls_provider_sup
    * erlang_ls_provider children (e.g. erlang_ls_definition_provider)

`erlang_ls_index:initiatlize/1` kicks off an indexing process after starting up all index processes. The `erlang_ls_index:index/1` function takes a document, stores it in the `documents` table and then sends it to each index for processing.


src/erlang_ls_code_navigation.erl
```diff
@@ -10,6 +10,11 @@
 %% API
 -export([ goto_definition/2 ]).
 
+-export([ app_path/0
```
jfacorro: These are exported so that the indexer process can acces them. The indexer currently only uses `app_path/0` though.

src/erlang_ls_document.erl
```diff
@@ -33,16 +38,28 @@
 create(Uri, Text) ->
   {ok, Tree, Extra} = erlang_ls_parser:parse(Text),
   AnnotatedTree = erlang_ls_tree:annotate(Tree, Extra),
-  POIs = erlang_ls_tree:points_of_interest(AnnotatedTree),
+  POIs = erlang_ls_poi:list(AnnotatedTree),
```
jfacorro: This was a bug introduced in a previous PR.

src/erlang_ls_index.erl
```diff
@@ -0,0 +1,95 @@
...
+%%==============================================================================
+
+-spec initialize(map()) -> ok.
+initialize(_Config) ->
+  %% Initialize all indexes
+  [  supervisor:start_child(erlang_ls_indexes_sup, [Index])
+     || Index <- ?INDEXES
+  ],
+  %% Start indexing process
+  erlang:spawn(fun indexer/0),
```
jfacorro: This can be improved,  but I think for now it works.

src/erlang_ls_parser.erl
```diff
@@ -4,17 +4,19 @@
...
-
 -spec parse(binary()) ->
    {ok, erlang_ls_tree:tree(), erlang_ls_tree:extra()} | {error, any()}.
 parse(Text) ->
   %% epp_dodger only works with source files,
   %% so let's use a temporary file.
-  ok = file:write_file(?TMP_PATH, Text),
-  parse_file(?TMP_PATH).
+  %% TODO: Ideally avoid writing to file at all (require epp changes)
+  TmpPath = tmp_path(),
```
jfacorro: This was necessary to avoid the indexing intefering with the same temp file that another file was using.

src/erlang_ls_provider.erl
```diff
@@ -2,69 +2,64 @@
...
+        , handle_call/3
+        , handle_cast/2
+        ]).
+
 -callback is_enabled()    -> boolean().
 -callback setup(config()) -> state().
 -callback teardown()      -> ok.
 
--export([ enabled_providers/0
-        , is_enabled/1
```
jfacorro: I removed from the export list most of these functions and had `initialize/1` use them internally instead.

robertoaloi:
Amazing contribution @jfacorro ! Thanks!

PR #99, opened 2019-10-19
Add code completion for exported functions.

robertoaloi:


src/erlang_ls_index.erl
```diff
@@ -33,8 +36,10 @@ initialize(_Config) ->
   [  supervisor:start_child(erlang_ls_indexes_sup, [Index])
      || Index <- ?INDEXES
   ],
-  %% Start indexing process
-  erlang:spawn(fun indexer/0),
+  %% TODO: This could be done asynchronously,
+  %%       but we need a way to know when indexing is done,
+  %%       or the tests will be flaky.
+  indexer(),
```
jfacorro: I don't like the fact that the indexing process is now synchronous, since that means that adding the deps and/or OTP will make everything unresponsive. But avoiding flakiness is a very good reason for doing so.

robertoaloi: The whole indexing procedure requires some thinking.

PR #101, opened 2019-10-19
[Fixes #97] Completion with parenthesis and number of arguments for exported functions

### LINKED ISSUE #97: Fix insertion of export entries from code completion
When choosing one of the export entries provided by the code completion feature, the label is in the format `function/arity`. When selected, the text inserted should instead be `function(`.

jfacorro:
#97


robertoaloi:
Nice! I create #112 as a follow-up.

PR #117, opened 2019-10-21
[Fixes #116] Allow non-whitespace chars just before module qualifier …

### LINKED ISSUE #116: Completion should work even when the prexif contains non whitespace chars
**Describe the bug**
When a line includes something other than whitespace characters in the line that are not the module name and the trigger `:`, the completion does not work.

**To Reproduce**
Write something like:
```
foo() ->
   X = erlang_ls_poi:
```
The options from completion will not show up.

**Expected behavior**
Exported functions from `erlang_ls_poi` (or whatever module typed in) should be returned.

**Actual behavior**
Options are not returned.

**Context**
 - `erlang_ls` version (tag/sha): `ba52457`
 - Editor used: Emacs
 - LSP client used: lsp-mode

jfacorro:
…(i.e. '(', '[', '{', ',' and ';') and also enything before that.

#116

### DIFF
--- a/src/erlang_ls_completion_provider.erl
+++ b/src/erlang_ls_completion_provider.erl
@@ -58,23 +58,24 @@ handle_request({completion, Params}, State) ->
 -spec find_completion(binary(), binary(), binary()) ->
    any().
 find_completion(Prefix, ?TRIGGER_CHARACTER, <<":">>) ->
-  Pattern = <<"\s+(?<MODULE>.*):">>,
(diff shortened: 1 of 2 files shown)


robertoaloi:
I would personally use a different approach. Something like:

```
S = " X = lists:".
{ok, Tokens, _} = erl_scan:string(S).
[{':',_}, Atom|_] = lists:reverse(Tokens).
erl_syntax:atom_value(Atom).
lists
```

PR #121, opened 2019-10-22
[#116] Refactor extraction of last token from text

### LINKED ISSUE #116: Completion should work even when the prexif contains non whitespace chars

robertoaloi:
### Description

* Move text manipulation functions to a separate module
* Use OTP library functions instead of custom pattern matching.

Fixes #116 .

### DIFF
--- a/src/erlang_ls_completion_provider.erl
+++ b/src/erlang_ls_completion_provider.erl
@@ -39,7 +39,8 @@ handle_request({completion, Params}, State) ->
    , <<"textDocument">> := #{<<"uri">> := Uri
                             }
    } = Params,
-  {ok, #{text := Text}} = erlang_ls_db:find(documents, Uri),
+  {ok, Document} = erlang_ls_db:find(documents, Uri),
+  Text = erlang_ls_document:text(Document),
   case maps:find(<<"context">>, Params) of
     {ok, Context} ->
       TriggerKind = maps:get(<<"triggerKind">>, Context),
--- a/src/erlang_ls_completion_provider.erl
+++ b/src/erlang_ls_completion_provider.erl
@@ -39,13 +39,14 @@ handle_request({completion, Params}, State) ->
    , <<"textDocument">> := #{<<"uri">> := Uri
                             }
    } = Params,
-  {ok, Document} = erlang_ls_db:find(documents, Uri),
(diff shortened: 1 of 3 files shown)


src/erlang_ls_completion_provider.erl
```diff
@@ -39,13 +39,14 @@ handle_request({completion, Params}, State) ->
    , <<"textDocument">> := #{<<"uri">> := Uri
                             }
    } = Params,
-  {ok, Document} = erlang_ls_db:find(documents, Uri),
+  {ok, #{text := Text}} = erlang_ls_db:find(documents, Uri),
```
jfacorro: I think this is breaking the opaqueness of the `document()` type. Why not use the `erlang_ls_document:text/1` function?

robertoaloi: Nice catch, will fix.

src/erlang_ls_completion_provider.erl
```diff
@@ -58,10 +60,10 @@ handle_request({completion, Params}, State) ->
 -spec find_completion(binary(), binary(), binary()) ->
    any().
 find_completion(Prefix, ?TRIGGER_CHARACTER, <<":">>) ->
-  {ok, Tokens, _} = erl_scan:string(binary_to_list(Prefix)),
-  [{':', _}, Token | _] = lists:reverse(Tokens),
-  case Token of
-    {atom, _, Module} ->
+  Token = erlang_ls_text:last_token(Prefix),
+  case erl_syntax:type(Token) of
```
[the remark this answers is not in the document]
robertoaloi: I see. I still don't like to expose the internal format of the atom-tuple, though. I will try/catch instead.

PR #124, opened 2019-10-23
[#123] Add support for textDocument/documentSymbol

### LINKED ISSUE #123: Implement support for the Document Symbols Request
The document symbol request is sent from the client to the server. The returned result for now can be a 
`SymbolInformation[]` which is a flat list of all symbols found in a given text document. See the protocol for details:

https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_documentSymbol

robertoaloi:
### Description

Add support for the document symbol provider. This makes it possible, in Emacs, to use the `imenu` command to navigate to functions.

Fixes #123 .

### DIFF
--- a/src/erlang_ls_document_symbol_provider.erl
+++ b/src/erlang_ls_document_symbol_provider.erl
@@ -1,0 +1,47 @@
+-module(erlang_ls_document_symbol_provider).
+
+-behaviour(erlang_ls_provider).
+
+-export([ handle_request/2
+        , is_enabled/0
+        , setup/1
+        , teardown/0
+        ]).
+
+-include("erlang_ls.hrl").
+
+%%==============================================================================
+%% erlang_ls_provider functions
+%%==============================================================================
+
+-spec is_enabled() -> boolean().
+is_enabled() ->
+  true.
+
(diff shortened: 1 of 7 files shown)


src/erlang_ls_document_symbol_provider.erl
```diff
@@ -0,0 +1,47 @@
...
+  case POIs of
+    [] -> {null, State};
+    _ ->
+      {[ #{ name => list_to_binary(io_lib:format("~p/~p", [F, A]))
+          , kind => ?SYMBOLKIND_FUNCTION
+          , location => #{ uri   => Uri
+                         , range => erlang_ls_protocol:range(Range)
+                         }
+          }
+         || #{ info := {function, {F, A}}
```
jfacorro: Is there any other kind of symbol we may want to return? Like `define`s for example, we could return them as `SYMBOLKIND_CONSTANT`. If you think it's worth it, we can create a separate issue and it can be left for another PR.

robertoaloi: Probably. For now I just wanted to replace the `imenu` usage I have with EDTS which only works with functions. We can add the rest separately.

PR #128, opened 2019-10-24
[#125] Cleanup the POI data structure, add points_of_interest/2 API.

### LINKED ISSUE #125: Add an API to restrict the list of POIs returned for a document to a set
Right now the `erlang_ls_document` module exposes an API to retrieve the list of POI for a given module. It should be possible to restrict that list to a given set of types.

robertoaloi:
### Description

* Remove the pointless `type` field from the POI map.
* Split the `info` field into separate `kind` and `data`
* Be able to retrieve only a subset of POIs

Fixes #125  .


jfacorro:
I like this.

PR #137, opened 2019-10-25
Make erlang_ls_protocol_impl:initialize/1 work with autozimu/LanguageClient-neovim

erszcz:
LanguageClient does not send the `initializationOptions` field of the config.
According to https://github.com/Microsoft/language-server-protocol/issues/567 heavy reliance on this field is discouraged.
This change allows to successfully connect and start using basic `erlang_ls` functionality in Vim, for example providing `erlc` warnings directly in the editor.

In the future I'd be interested in using Gradualizer with Vim through the language server - this is the first necessary step. [My current Gradualizer integration attempt is sufficient for Vim, but not reusable in other editors.](https://github.com/erszcz/neomake/blob/erlang-gradualizer-wip/autoload/neomake/makers/ft/erlang.vim)

### DIFF
--- a/src/erlang_ls_methods.erl
+++ b/src/erlang_ls_methods.erl
@@ -77,11 +77,11 @@ method_to_function_name(Method) ->
 -spec initialize(params(), state()) -> result().
 initialize(Params, State) ->
   #{ <<"rootUri">> := RootUri
-   , <<"initializationOptions">> := InitOptions
      %% TODO: Store client capabilities, use them in completion_provider
      %%       to verify when Context is present
    , <<"capabilities">> := _ClientCapabilities
    } = Params,
+  InitOptions = maps:get(<<"initializationOptions">>, Params, #{}),
   Config = erlang_ls_config:initialize(RootUri, InitOptions),
   ok     = erlang_ls_index:initialize(Config),
   ok     = erlang_ls_provider:initialize(Config),


src/erlang_ls_protocol_impl.erl
```diff
@@ -27,11 +27,11 @@
   {response, map()} | {} | {notification, binary(), map()}.
 initialize(Params) ->
   #{ <<"rootUri">> := RootUri
-   , <<"initializationOptions">> := InitOptions
      %% TODO: Store client capabilities, use them in completion_provider
      %%       to verify when Context is present
    , <<"capabilities">> := _ClientCapabilities
    } = Params,
+  InitOptions = maps:get(<<"initializationOptions">>, Params, []),
```
jfacorro: I think the default value should be an empty map `#{}` instead of the empty list. The reason why it's not failing is because there is a catch-all clause in `erlang_ls_config:config_path/1`, but from the spec the function expects a `map()`.

robertoaloi:
Good point, thanks for the fix. It's also nice that you're trying the server in VIM. If you manage to get it up and running, a _how-to-vim_ section in the README would be more than welcome!

jfacorro:
This conflict was my fault, I will make the update in this branch and merge if you want @erszcz.

jfacorro:
Ah, but I won't be able to update the branch in your fork though. 😞 

Sorry for the toubles.

erszcz:
I've rebased on current master (601c8c0). No worries :)

I'll be happy to write a few lines on getting up and running with Vim once I spend some more time using the language server. I still have to figure out how to configure code / include paths, etc.

PR #138, opened 2019-10-25
[Fixes #19] [Fixes #20] Shutdown request keep state

### LINKED ISSUE #19: Shutdown Request

### LINKED ISSUE #20: Exit Notification
A notification to ask the server to exit its process. The server should exit with success code 0 if the shutdown request has been received before; otherwise with error code 1.

Notification:

* method: `exit`
* params: `void`

jfacorro:
### Description

Keep track of the status of the server so we can respond accordingly to requests after `shutdown` and `exit` with the correct exit code, following the protocol specifications.

Fixes #19. Fixes #20.

### DIFF
--- a/src/erlang_ls_utils.erl
+++ b/src/erlang_ls_utils.erl
@@ -1,0 +1,7 @@
+-module(erlang_ls_utils).
+
+-export([halt/1]).
+
+-spec halt(integer()) -> no_return().
+halt(ExitCode) ->
+  erlang:halt(ExitCode).
--- a/test/prop_statem.erl
+++ b/test/prop_statem.erl
@@ -3,7 +3,6 @@
 %%==============================================================================
 -module(prop_statem).
 
-
 %%==============================================================================
 %% Behaviours
 %%==============================================================================
@@ -34,7 +33,8 @@
 %%==============================================================================
 initial_state() ->
   #{ connected => false
-   , documents   => []
+   , shutdown  => false
+   , documents => []
    }.
 
 %%==============================================================================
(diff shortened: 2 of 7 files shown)


src/erlang_ls_utils.erl
```diff
@@ -0,0 +1,7 @@
+-module(erlang_ls_utils).
```
robertoaloi: Do we really need this?

test/prop_statem.erl
```diff
@@ -155,6 +158,55 @@ did_close_post(_S, _Args, Res) ->
...
+  catch disconnect(),
+  S#{shutdown => false, connected => false}.
+
+exit_post(S, _Args, Res) ->
+  ExpectedExitCode = case maps:get(shutdown, S, false) of
+                       true  -> 0;
+                       false -> 1
+                     end,
+  wait_for(halt_called, 1000),
+  ?assert(meck:called(erlang_ls_utils, halt, [ExpectedExitCode])),
```
robertoaloi: I see now why you wanted a wrapper around the halt function.

jfacorro: Testing the usage of `erlang:exit/1` always ends up being a little convoluted.

robertoaloi:
This was long due. Thanks!

PR #140, opened 2019-10-25
[#139] Type specs on hover

### LINKED ISSUE #139: Show type specs on hover for fully qualified applications

robertoaloi:
### Description

This PR allows you to view type specs for fully qualified functions while hovering over a fully qualified application. The feature was already partially implemented, but it didn't use a provider and it was disabled.

Fixes #139 .


include/erlang_ls.hrl
```diff
@@ -469,6 +469,21 @@
                                , containerName => binary()
                                }.
 
+%%------------------------------------------------------------------------------
+%% Signatures
```
robertoaloi: This is actually independent from hovering. The reason from introducing this is that I started implementing the `signatureHelp` method, then realized the `hover` was already there (some overlap exists between the two methods). Since this could still turn useful in the future, leaving it for now.

src/erlang_ls_methods.erl
```diff
@@ -136,7 +136,7 @@ exit(_Params, State) ->
   erlang_ls_utils:halt(ExitCode).
 
 %%==============================================================================
-%% textdocument_didopen
+%% textDocument/didopen
```
robertoaloi: These were probably erroneously replaced some time ago as part of a search/replace which passed review.

src/erlang_ls_tree.erl
```diff
@@ -198,7 +198,9 @@ attribute(Tree, Extra) ->
     {spec, {spec, {{F, A}, _}}} ->
       SpecLocations = maps:get(spec_locations, Extra, []),
       Locations     = proplists:get_value({F, A}, SpecLocations),
-      [erlang_ls_poi:poi(Tree, type_application, {T, L}, Extra) || {T, L} <- Locations];
+      [ erlang_ls_poi:poi(Tree, spec, {{F, A}, Tree}, Extra)
```
robertoaloi: When encountering specs, we were only indexing the types and not the spec itself. This is not the case anymore. The reason for passing the whole `Tree` as the third argument to the `poi/4` is so that we can then use the convenient `erl_prettypr` module to pretty-print the spec. We could have passed the pretty printed version but that would have been worse performance-wise. Also, it could turn useful to get the whole raw attribute in the DB.

Issue #35, opened 2019-08-06
Make POI type opaque


robertoaloi:
Ensure _Point of Interest_ data structure are created via constructors and accessed only via getters.

jfacorro:
Based on the comment in #159, should we close this one.

PR #158, opened 2019-10-26
[#35] POI handling cleanup

robertoaloi:
### Description

This PR does not bring in functional changes, but cleans up things around POI management.

* Move `tree/0` type definition to the header file.
* Stop exposing the `Extra` outside of the parser.
* Merge `parser` and `tree` modules (now the `parse` API returns `{ok, POIs}` instead of `{ok, Tree, Extra}`.
* Rename the POI constructor to `new` for clarity (we should rename the `create` constructor for documents, too).
* Move range functions to a separate module.


robertoaloi:
Opened too early. Still not ready for review.

PR #160, opened 2019-10-26
[#143] Enforce 80 char limit via Elvis

### LINKED ISSUE #143: Enforce 80 chars limit via Elvis
The [Contributing Guidelines](https://github.com/erlang-ls/erlang_ls/blob/master/.github/CONTRIBUTING.md) states that a 80-char limit should be respected. Let's enforce it via elvis. Existing code exceeding the limit should be broken up into multiple lines.

robertoaloi:
Fixes #143 .

@jfacorro Not sure if there is a better way to install elvis (e.g. via a rebar3 plugin). Please advise.


robertoaloi:
Superseded by #161 . I just realized we were already running elvis via a compiler hook.

PR #173, opened 2019-10-29
[#172] Escript with random node name

### LINKED ISSUE #172: Random node name when run as an escript
In order to connect to the `erlang_ls` node through a remote shell it would be nice to assign a name to the Erlang node when the server is started as an escript.

jfacorro:
### Description

Escript with random node name.

Fixes #172.

### DIFF
--- a/src/erlang_ls.erl
+++ b/src/erlang_ls.erl
@@ -4,6 +4,15 @@
 
 -spec main([any()]) -> ok.
 main([Port]) ->
+  init_node_name(),
   application:set_env(erlang_ls, port, list_to_integer(Port)),
   application:ensure_all_started(erlang_ls),
+  lager:info("Started erlang_ls server ~p", [node()]),
   receive _ -> ok end.
+
+-spec init_node_name() -> ok.
+init_node_name() ->
+  Name     = "erlang_ls_" ++ integer_to_list(rand:uniform(16#FFFFFFFFF)),
+  NodeName = list_to_atom(Name),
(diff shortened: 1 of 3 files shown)


src/erlang_ls.erl
```diff
@@ -4,6 +4,15 @@
...
   application:set_env(erlang_ls, port, list_to_integer(Port)),
   application:ensure_all_started(erlang_ls),
+  lager:info("Started erlang_ls server ~p", [node()]),
   receive _ -> ok end.
+
+-spec init_node_name() -> ok.
+init_node_name() ->
+  Name     = "erlang_ls_" ++ integer_to_list(rand:uniform(16#FFFFFFFFF)),
+  NodeName = list_to_atom(Name),
+  net_kernel:start([NodeName, shortnames]),
```
robertoaloi: This requires EPMD to be already started or it will fail, right?

robertoaloi: See for example: https://github.com/robertoaloi/ansible-nodetool/blob/master/nodetool#L123

jfacorro: Yup, but AFAIK there is no way to ensure EPMD is started, other than doing some dirty hack with `os:cmd/1`. Right?

robertoaloi: Not that I know. RabbitMQ does the same: https://github.com/rabbitmq/rabbitmq-common/blob/master/src/rabbit_nodes_common.erl#L73

jfacorro: Updated the PR.

PR #193, opened 2019-11-01
[#112] Include function argument names (when available)

### LINKED ISSUE #112: Show arguments when selecting a function name from auto-completion
When selecting one of the exported functions from auto-complete (see #97), the names of the arguments should also be visible, instead of empty strings. For example, typing:

    lists:keystore(Key, N, TupleList1, NewTuple)

Would give us:

    keystore(Key, N, TupleList1, NewTuple)

Instead of just:

    keystore(,,,)

jfacorro:
### Description

Implements showing variable names when completing function calls. The variable names are extracted from the function definition and not from the specs, since in general specs don't include variables names. If the argument is not a variable but a pattern, then a default `ArgN` is included where `N` is an integer with the number of the argument.

Fixes #112.

### DIFF
--- a/include/erlang_ls.hrl
+++ b/include/erlang_ls.hrl
@@ -557,9 +557,10 @@
                    | type_definition
                    | variable.
 -type poi_range() :: #{ from := pos(), to := pos() }.
--type poi()       :: #{ kind  => poi_kind()
-                      , data  => any()
+-type poi()       :: #{ kind  := poi_kind()
+                      , data  := any()
                       , range := poi_range()
+                      , tree  := tree()
                       }.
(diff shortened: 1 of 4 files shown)


include/erlang_ls.hrl
```diff
@@ -557,9 +557,10 @@
                    | type_definition
                    | variable.
 -type poi_range() :: #{ from := pos(), to := pos() }.
--type poi()       :: #{ kind  => poi_kind()
-                      , data  => any()
+-type poi()       :: #{ kind  := poi_kind()
+                      , data  := any()
                       , range := poi_range()
+                      , tree  := tree()
```
robertoaloi: Why mandatory?

PR #197, opened 2019-11-02
[#107] Keywords completion

### LINKED ISSUE #107: Code completion for Erlang keywords
A handy list of reserved keywords is available at:

https://github.com/erlang/otp/blob/7cc8adca707e413c0449275f225567b097714318/lib/stdlib/src/erl_scan.erl#L1173

```
reserved_word('after') -> true;
reserved_word('begin') -> true;
reserved_word('case') -> true;
reserved_word('try') -> true;
reserved_word('cond') -> true;
reserved_word('catch') -> true;
reserved_word('andalso') -> true;
reserved_word('orelse') -> true;
reserved_word('end') -> true;
reserved_word('fun') -> true;
reserved_word('if') -> true;
reserved_word('let') -> true;
reserved_word('of') -> true;
reserved_word('receive') -> true;
reserved_word('when') -> true;
reserved_word('bnot') -> true;
reserved_word('not') -> true;
reserved_word('div') -> true;
reserved_word('rem') -> true;
reserved_word('band') -> true;
reserved_word('and') -> true;
reserved_word('bor') -> true;
reserved_word('bxor') -> true;
reserved_word('bsl') -> true;
reserved_word('bsr') -> true;
reserved_word('or') -> true;
reserved_word('xor') -> true;
reserved_word(_) -> false.
```

robertoaloi:
Fixes #107 .


src/erlang_ls_completion_provider.erl
```diff
@@ -208,3 +218,17 @@ include_file_macros(Kind, Name) ->
...
       []
   end.
+
+%%==============================================================================
+%% Keywords
+%%==============================================================================
+
+-spec keywords() -> [map()].
+keywords() ->
+  Keywords = [ 'after', 'and', 'andalso', 'band', 'begin', 'bnot', 'bor', 'bsl'
```
jfacorro: Why not have them as binaries in the list already?

src/erlang_ls_completion_provider.erl
```diff
@@ -91,18 +94,22 @@ find_completion( Prefix
       variables(Document);
     %% Check for "[...] atom"
     [{atom, _, _} | _] ->
-      modules(Prefix) ++ functions(Document, false);
+      modules() ++ functions(Document, false);
```
jfacorro: `++ keywords()` should be included here as well, right?

README.md
```diff
@@ -54,6 +54,10 @@ file (assuming the `erlang_ls` executable is available in your `PATH`:
     ;; Enable LSP automatically for Erlang files
     (add-hook 'erlang-mode-hook #'lsp)
 
+    ;; Override the key bindings from erlang-mode to use LSP for completion
+    (eval-after-load 'erlang
+      '(define-key erlang-mode-map (kbd "C-M-i") #'company-lsp))
```
jfacorro: Maybe it's worth adding in the comment or somewhere that there is another binding from `company-lsp` to trigger this which is closer to what other IDEs have, which is `C-space`.

src/erlang_ls_completion_provider.erl
robertoaloi: I was going to do that, but then decided not to for readability.

src/erlang_ls_completion_provider.erl
robertoaloi: Yes.

README.md
robertoaloi: Uhm, I have `C-space` bound to `set-mark-command`.

PR #201, opened 2019-11-03
[#16] STDIO support

### LINKED ISSUE #16: Add support for STDIO transport
Currently, the Language Server only supports TCP. The transport should be configurable and also support STDIO.

jfacorro:
### Description

Implement support for STDIO.

Fixes #16.


src/erlang_ls_client.erl
```diff
@@ -140,19 +144,21 @@ workspace_symbol(Query) ->
...
-init({Host, Port}) ->
+-spec init({tcp, hostname(), port_no()}) -> {ok, state()}.
+init({tcp, Host, Port}) ->
   Opts         = [binary, {active, once}, {packet, 0}],
   {ok, Socket} = gen_tcp:connect(Host, Port, Opts),
-  {ok, #state{socket = Socket}}.
+  {ok, #state{transport = tcp, connection = Socket}};
+init({stdio, IoDevice}) ->
+  Self = self(),
+  proc_lib:spawn_link(erlang_ls_stdio, loop, [[], IoDevice, Self, ?JSON_OPTS]),
```
jfacorro: This is for using the same implementation as the server.

src/erlang_ls_client.erl
```diff
@@ -268,59 +264,54 @@ handle_call({initialize, RootUri, InitOptions}, From, State) ->
...
+  RequestId = State#state.request_id,
   Method = <<"exit">>,
   Params = #{},
   Content = erlang_ls_protocol:request(RequestId, Method, Params),
-  gen_tcp:send(Socket, Content),
+  send(Content, State),
   {reply, ok, State}.
 
 -spec handle_cast(any(), state()) -> {noreply, state()}.
+handle_cast( {messages, Responses}
```
jfacorro: This is handling the messages that `erlang_ls_stdio:loop/4` function sends throuhg `erlang_ls_server:send_requests/2`.

test/erlang_ls_fake_stdio.erl
```diff
@@ -0,0 +1,111 @@
+-module(erlang_ls_fake_stdio).
```
jfacorro: This file was created to emulated the communication between client and server through stdio.

src/erlang_ls_client.erl
```diff
@@ -140,19 +144,21 @@ workspace_symbol(Query) ->
 %%==============================================================================
 %% gen_server Callback Functions
 %%==============================================================================
--spec init({hostname(), port_no()}) -> {ok, state()}.
-init({Host, Port}) ->
+-spec init({tcp, hostname(), port_no()}) -> {ok, state()}.
```
robertoaloi: Why not to define a `client` behaviour and have a `tcp_client` and a `stdio_client`?

test/erlang_ls_completion_SUITE.erl
```diff
@@ -38,27 +39,29 @@
 suite() ->
   [{timetrap, {seconds, 30}}].
 
+-spec all() -> [atom()].
+all() ->
+  [{group, tcp}, {group, stdio}].
```
robertoaloi: Nice!

test/prop_statem.erl
```diff
@@ -275,6 +274,11 @@ teardown(_) ->
 %%==============================================================================
 cleanup() ->
   catch disconnect(),
+  %% Restart the server, since though the client disconnects the
+  %% server keeps its state.
+  application:stop(ranch),
```
robertoaloi: We should add ranch to the `.app.src`.

test/erlang_ls_fake_stdio.erl
robertoaloi: Can you expand? Why can't we use the normal server/client for the tests (I guess intercepting all other output)?

jfacorro: Because the client needs a way to *write to the server's stdin* and a way to *read from the server's stdout*. This can't be achieved in a simple way that I know of.

src/erlang_ls_client.erl
jfacorro: Becuase it would have involved a lot more time than I had already invested 😄.

rebar.lock
```diff
@@ -4,7 +4,7 @@
  {<<"goldrush">>,{pkg,<<"goldrush">>,<<"0.1.9">>},1},
  {<<"jsx">>,{pkg,<<"jsx">>,<<"2.9.0">>},0},
  {<<"lager">>,{pkg,<<"lager">>,<<"3.6.8">>},0},
- {<<"ranch">>,{pkg,<<"ranch">>,<<"1.5.0">>},0},
+ {<<"ranch">>,{pkg,<<"ranch">>,<<"1.7.1">>},0},
```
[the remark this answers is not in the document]
jfacorro: While going through ranch's docs I found that `ranch:accept_ack/1` had been deprecated in `2.x`, `1.7.1` was the latest version for `1.x` and included the new non-deprecated function `ranch:handshake/1`.

test/erlang_ls_fake_stdio.erl
robertoaloi: Could a file be used as a IO device for the tests?

rebar.lock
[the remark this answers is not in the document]
robertoaloi: I see, thanks.

test/prop_statem.erl
jfacorro: It is added, but stopping an application doens't stop it's dependecy applications.

robertoaloi:
@jfacorro I tried this branch with IntelliJ (I will add instructions shortly, so you can reproduce it) but it didn't work. One obvious problem is that when not using debug mode, lager will print to stdout breaking the communication. But there must be something else. The client cannot communicate with the server for some reason. Did you try the stdio transport with a real client (e.g. the Emacs one)?

jfacorro:
@robertoaloi Yes, I tried it with `lsp-mode` but using the `lsp-stdio-connection` (see diff [here](https://github.com/emacs-lsp/lsp-mode/compare/master...jfacorro:lsp-erlang-stdio-support?expand=1)).

robertoaloi:
Just tried with the Emacs client (in debug mode) and it seems to work correctly, so something is weird with the IntelliJ LSP Server.

test/erlang_ls_fake_stdio.erl
jfacorro: I thought about that, but I couldn't think of a way to make it work with files. We would need two files `A` and `B`, and some code that would use:
- `A` as stdout for the server and `B` as stdin.
- `B` as stdout for the client and `A` as stdin.
This code would need to be a process that implements an IO server and uses `A` and `B` as described above.

robertoaloi: Let's live with this for now.

src/erlang_ls.erl
```diff
@@ -3,13 +3,30 @@
...
   lager:info("Started erlang_ls server ~p", [node()]),
   receive _ -> ok end.
 
+-spec parse_args([string()]) -> ok.
+parse_args([]) ->
+  ok;
+parse_args(["--stdio" | Rest]) ->
+  application:set_env(erlang_ls, transport, erlang_ls_stdio),
+  parse_args(Rest);
+parse_args(["--tcp" | Rest]) ->
```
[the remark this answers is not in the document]
jfacorro: Changed to `--transport tcp|stdio`

src/erlang_ls_app.erl
```diff
@@ -16,23 +16,13 @@
...
 -spec start(normal, any()) -> {ok, pid()}.
 start(_StartType, _StartArgs) ->
-  Port = application:get_env(erlang_ls, port, ?DEFAULT_PORT),
-  {ok, _} = ranch:start_listener( erlang_ls
-                                , ranch_tcp
-                                , [{port, Port}]
-                                , erlang_ls_server
-                                , []
-                                ),
+  Transport = application:get_env(erlang_ls, transport, erlang_ls_tcp),
```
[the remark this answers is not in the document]
jfacorro: Placed all default values in the `.app.src` and replaced all calls to `application:get_env/3` to `application:get_env/2`.

src/erlang_ls.erl
```diff
@@ -3,13 +3,30 @@
...
   init_node_name(is_debug()),
-  application:set_env(erlang_ls, port, list_to_integer(Port)),
   application:ensure_all_started(erlang_ls),
   lager:info("Started erlang_ls server ~p", [node()]),
   receive _ -> ok end.
 
+-spec parse_args([string()]) -> ok.
+parse_args([]) ->
+  ok;
+parse_args(["--stdio" | Rest]) ->
```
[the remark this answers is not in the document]
jfacorro: Updated.

robertoaloi:
Thank you @jfacorro 🙏

Issue #190, opened 2019-10-31
Add logs and track times for indexing


robertoaloi:
@jfacorro I already have a branch with this, so assigning it to myself.

jfacorro:
Fixed #205.

PR #205, opened 2019-11-07
[#190] Indexing refactoring

jfacorro:
### Description

This PR refactors the indexing process by adding a pool of workers, placing all indexing code in the same module, enabling atomic updates to entries and adding application variables to disable the indexing of dependencies and/or OTP.

Fixes # .


src/erlang_ls_db.erl
```diff
@@ -55,6 +55,18 @@ store(Table, Key, Value) ->
   true = ets:insert(Table, {Key, Value}),
   ok.
 
+-spec update(table(), key(), any(), any()) -> ok.
+update(Table, Key, Old, New) ->
+  Replace = [{{Key, Old}, [], [{const, {Key, New}}]}],
+  case ets:select_replace(Table, Replace) of
```
jfacorro: This is the new function that has been made available in a "recent" OTP release.

src/erlang_ls_parser.erl
```diff
@@ -266,12 +266,12 @@ attribute(Tree, Extra) ->
         {include, [String]} ->
           [erlang_ls_poi:new( Tree
                             , include
-                            , erl_syntax:string_literal(String)
+                            , erl_syntax:string_value(String)
```
jfacorro: `erl_syntax:string_literal/1` was returning `"\"foo\""`, I don't know how we got it working when the value was that 🤷‍♂ .

src/erlang_ls_server.erl
```diff
@@ -21,12 +21,16 @@
...
         ]).
 
-%% Notifications API
+%% API
 -export([ process_requests/2
         , set_connection/2
         , send_notification/3
         ]).
 
+%% Testing
```
jfacorro: I added this function to simplify the setup we had in the proper tests. Just resetting the internal state makes everything much much simpler.

src/erlang_ls_server.erl
```diff
@@ -51,9 +55,9 @@
 %%==============================================================================
 -spec start_link(module()) -> {ok, pid()}.
 start_link(Transport) ->
-  {ok, Server} = gen_server:start_link(?MODULE, Transport, []),
-  {ok, _} = Transport:start_listener(Server),
-  {ok, Server}.
+  {ok, Pid} = gen_server:start_link({local, ?MODULE}, ?MODULE, Transport, []),
```
jfacorro: There should only be a single `erlang_ls_server` at a time, so I added a name to it. This will enable to stop having to pass the `Pid` around for notifications, for example.

src/erlang_ls_utils.erl
```diff
@@ -12,16 +13,25 @@
...
 %% Look for a given module in the DB and return the respective
 %% `URI`. If the module is not in the DB, try to index it.
+
 -spec find_module(atom()) -> {ok, uri()} | {error, any()}.
 find_module(M) ->
-  case erlang_ls_db:find(completion_index, M) of
+  find_module(M, erl).
+
+-spec find_module(atom(), erl | hrl) -> {ok, uri()} | {error, any()}.
+find_module(M, Extension) ->
```
jfacorro: I modified this function so it could also be used for finding and indexing `hrl` files.

src/erlang_ls_utils.erl
```diff
@@ -12,16 +13,25 @@
...
+  case erlang_ls_db:find(modules, M) of
     {ok, Uri} ->
       {ok, Uri};
     {error, not_found} ->
-      FileName = atom_to_list(M) ++ ".erl",
-      erlang_ls_index:find_and_index_file(FileName)
+      FileName = atom_to_list(M) ++ extension(Extension),
+      erlang_ls_indexer:find_and_index_file(FileName, sync)
```
jfacorro: We need this indexing to be synchronous, since we rely on the file having been indexed after this function returns.

src/erlang_ls_indexer.erl
```diff
@@ -0,0 +1,210 @@
...
+-spec index_deps() -> any().
+index_deps() ->
+  case application:get_env(erlang_ls, index_deps) of
+    {ok, true}  -> gen_server:cast(?SERVER, {index_deps});
+    {ok, false} -> lager:info("Not indexing dependencies")
+  end.
+
+-spec index_otp() -> any().
+index_otp() ->
+  case application:get_env(erlang_ls, index_otp) of
```
jfacorro: These were added as application variables, I considered adding them as user configuration, but I thought that the users shouldn't care about this. @robertoaloi WDYT?

src/erlang_ls_indexer.erl
```diff
@@ -0,0 +1,210 @@
...
+%% Exported functions
+%%==============================================================================
+-spec find_and_index_file(string()) ->
+   {ok, uri()} | {error, any()}.
+find_and_index_file(FileName) ->
+  find_and_index_file(FileName, async).
+
+-spec find_and_index_file(string(), async | sync) ->
+   {ok, uri()} | {error, any()}.
+find_and_index_file(FileName, SyncAsync) ->
```
jfacorro: The `SyncAsync` argument is for being able to request one or the other as a client of the indexer. This is useful when we need "find and index" and then rely on the fact that the file has already been indexed.

src/erlang_ls_db.erl
robertoaloi: That is OTP 20.0

src/erlang_ls_indexer.erl
robertoaloi: It is fine as an application env for now.

src/erlang_ls_indexer.erl
```diff
@@ -0,0 +1,215 @@
...
+          end
+      end,
+  {Time, {Succeeded, Failed}} = timer:tc( filelib
+                                        , fold_files
+                                        , [ Dir
+                                          , ".*\\.[e,h]rl$"
+                                          , true
+                                          , F
+                                          , {0, 0} ]),
+  lager:info("Finished indexing directory. [dir=~s] [time=~p]"
```
robertoaloi: Are these times meaningful now that the indexing happens asynchronously?

src/erlang_ls_server.erl
robertoaloi: That was something I noticed as well. 👍

PR #208, opened 2019-11-07
[#203] Filter by prefix

### LINKED ISSUE #203: Restrict completion items based on prefix
When trying out the branch that indexes the application, the deps and OTP, the completion feature made typing anything extremely slow. This is in part because the completion currently returns all indexed modules, and when there are *a lot* this means the full list is returned on every completion.

Use the prefix of what has been typed so far to filter the items returned in the completion.

N/A

**Additional context**
N/A

jfacorro:
### Description

Restrict completion of modules based on prefix.

Fixes #203.

### DIFF
--- a/src/erlang_ls_completion_provider.erl
+++ b/src/erlang_ls_completion_provider.erl
@@ -83,13 +83,11 @@ find_completion( Prefix
     [{var, _, _} | _] ->
       variables(Document);
     %% Check for "[...] atom"
-    [{atom, _, _} | _] ->
-      modules() ++ functions(Document, false) ++ keywords();
+    [{atom, _, Name} | _] ->
+      NameBinary = atom_to_binary(Name, utf8),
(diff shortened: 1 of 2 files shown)


src/erlang_ls_completion_provider.erl
```diff
@@ -83,13 +83,11 @@ find_completion( Prefix
...
-      modules() ++ functions(Document, false) ++ keywords();
+    [{atom, _, Name} | _] ->
+      NameBinary = atom_to_binary(Name, utf8),
+      keywords() ++ modules(NameBinary) ++ functions(Document, false);
     _ ->
-      modules()
-        ++ functions(Document, false)
-        ++ variables(Document)
-        ++ keywords()
+      []
```
jfacorro: We shouldn't try to return anything when it's unclear what to return.

PR #219, opened 2019-11-10
[#147] Write logs under single location

### LINKED ISSUE #147: Logs should end up in /var/log/erlang_ls/PROJECT
Right now it's very confusing to locate logs for a specific run. This is due to the fact logs are sparse in different dirs.

Ideally, the path should be configurable by a global `~/.config/erlang_ls` file

jfacorro:
### Description

Write logs for all projects under a single root dir.

Fixes #147.

### DIFF
--- a/src/erlang_ls.erl
+++ b/src/erlang_ls.erl
@@ -2,14 +2,23 @@
 
 -export([ main/1 ]).
 
+-define(APP, erlang_ls).
+
 -spec main([any()]) -> ok.
 main(Args) ->
-  application:load(erlang_ls),
-  parse_args(Args),
-  init_node_name(is_debug()),
-  application:ensure_all_started(erlang_ls),
+  %% Initialization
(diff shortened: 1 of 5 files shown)


src/erlang_ls.erl
```diff
@@ -19,16 +28,54 @@ parse_args(["--transport", Name | Rest]) ->
...
+      , {level, info}
+      ]
+    }
+  ].
+
+-spec log_root() -> string().
+log_root() ->
+  {ok, CurrentDir} = file:get_cwd(),
+  Dirname = filename:basename(CurrentDir),
+  filename:join([ "/usr/local/var/log/erlang_ls", Dirname]).
```
robertoaloi: I would like this to be configurable. Also, Isn't `/var/log` enough?

jfacorro: Configurable through the application's env vars?

PR #221, opened 2019-11-11
[#102] Implement folding over files manually and skip symlinks

### LINKED ISSUE #102: Indexing should not process the same files twice
In a `rebar3` project the `_build` subdirectory includes a lot of symlinks. Since the indexing process  is currently not very involved (i.e. it looks for all `.erl` files in the directory tree), then all of the application's and dependencies' files are processed multiple times when the `_build` includes various subdirectories for different `rebar3` profiles (e.g. `default`, `test`, `debug`).

Even when there is only `default` then the application's files will get processed at least twice, since the `src`, `test` and `include` dirs are symlinked by `rebar3` in the `default` profile subdirectory:
```
$ ll _build/default/lib/erlang_ls/
total 0
drwxr-xr-x  32 juan.facorro  staff   1.0K Oct 18 22:41 ebin
lrwxr-xr-x   1 juan.facorro  staff    19B Oct 13 15:39 include -> ../../../../include
lrwxr-xr-x   1 juan.facorro  staff    16B Oct 13 15:39 priv -> ../../../../priv
lrwxr-xr-x   1 juan.facorro  staff    15B Oct 13 15:39 src -> ../../../../src
```

jfacorro:
### Description

Implement folding over files manually and skip symlinks

Fixes #102.

### DIFF
--- a/src/els_utils.erl
+++ b/src/els_utils.erl
@@ -3,22 +3,26 @@
 -export([ find_document/1
         , find_module/1
         , find_module/2
-        , include_filename/2
+        , fold_files/4
         , halt/1
         , start_epmd/0
         ]).
 
 -include("erlang_ls.hrl").
 
-%% @doc Look for a module in the DB.
-%%
-%% Look for a given module in the DB and return the respective
-%% `URI`. If the module is not in the DB, try to index it.
+%%==============================================================================
+%% File and module functions
+%%==============================================================================
 
+%% @equiv find_module(Module, erl)
 -spec find_module(atom()) -> {ok, uri()} | {error, any()}.
 find_module(M) ->
   find_module(M, erl).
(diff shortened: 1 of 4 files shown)


src/els_utils.erl
```diff
@@ -75,3 +86,41 @@ epmd_path() ->
...
+      case Filter(Path) of
+        true  -> F(Path, Acc);
+        false -> Acc
+      end;
+    false -> fold_files(F, Filter, Path, Acc)
+  end.
+
+-spec is_symlink(string()) -> boolean().
+is_symlink(Path) ->
+  case file:read_link(Path) of
```
robertoaloi: We could also consider `file:read_link_info/2` combined with the `raw` option.

PR #225, opened 2019-11-13
[#222] Don't write logs by default

### LINKED ISSUE #222: Logging should be disabled by default
Make logging to files opt-in. This will simplify the fact that we have to choose a default path for logs(#147) that is actually platform dependent (e.g. OS X, Linux and Windows have different "correct" places to store logs).

jfacorro:
### Description

Don't write logs by default, only when `--log-dir` option is provided.

Fixes #222.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -26,19 +26,6 @@
 
 {minimum_otp_vsn, "21.0"}.
 
-{relx, [ { release
-         , {erlang_ls, "0.0.1"}
-         , [ erlang_ls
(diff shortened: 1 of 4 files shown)


rebar.config
```diff
@@ -26,19 +26,6 @@
 
 {minimum_otp_vsn, "21.0"}.
 
-{relx, [ { release
```
robertoaloi: Thanks for the cleanup!

PR #238, opened 2019-11-17
[#227] Add VSCode setup

### LINKED ISSUE #227: Add VS Code setup instructions to README
Add setup instructions for VS Code on how to start with the Erlang Language Server.

Reference: https://github.com/erlang-ls/vscode

robertoaloi:
### Description

Fixes #227.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,6 +80,18 @@ To enable logging on the client-side, just:
 
     (setq lsp-log-io t)
 
+### VSCode Setup
+
+The Erlang Language Server is available in VSCode via a [dedicated
+extension](https://github.com/erlang-ls/vscode).
+
+To try it out, simply open VSCode and install the extension via the
+Marketplace:
+
+    Preferences > Extensions
+
+Look for the `erlang-ls` extension and install it. That's it.
+
 ### IntelliJ Setup
 
 > **WARNING**: The current version of the IntelliJ LSP plugin (1.5.4)


README.md
```diff
@@ -80,6 +80,18 @@ To enable logging on the client-side, just:
...
+
+The Erlang Language Server is available in VSCode via a [dedicated
+extension](https://github.com/erlang-ls/vscode).
+
+To try it out, simply open VSCode and install the extension via the
+Marketplace:
+
+    Preferences > Extensions
+
+Look for the `erlang-ls` extension and install it. That's it.
```
jfacorro: `erlang_ls` shouldn't be in the path?

robertoaloi: Not a VSCode expert, but I believe the escript is packaged as part of the extension and therefore a local path is used. But please try it and let me know if it does not work. The VSCode installation process is trivial.

jfacorro: I just tried it and it just works. Checking the process I see the path to an `erlang_ls` binary included in the VS Code:
```
$ ps aux | grep erlang_ls
juan.facorro     27945   0.1  6.3  6548380 1053796   ??  S    10:03AM   6:53.41 /Users/juan.facorro/.vscode/extensions/erlang-ls.erlang-ls-0.0.1/erlang_ls/_build/default/bin/erlang_ls -B -- -root /Users/juan.facorro/.erlang-installs/21.3 -progname erl -- -home /Users/juan.facorro -- -boot no_dot_erlang -noshell -escript main erlang_ls -pz erlang_ls/erlang_ls/ebin -run escript start -extra /Users/juan.facorro/.vscode/extensions/erlang-ls.erlang-ls-0.0.1/erlang_ls/_build/default/bin/erlang_ls --transport stdio
```
I guess something worth mentioning in general is that the Erlang/OTP version available in the `PATH` should be 21+?

robertoaloi: Added.

PR #243, opened 2019-11-19
[#242] Avoid writing to temporary file for parsing

### LINKED ISSUE #242: Don't create temp files for parsing
The parser currently needs to create a file whenever it parses text, it shouldn't need to.

jfacorro:
### Description

Avoid writing to temporary file for parsing.

Fixes #242.

### DIFF
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -18,39 +18,21 @@
 %%==============================================================================
 %% API
 %%==============================================================================
--spec parse(binary()) -> {ok, [poi()]} | {error, any()}.
+-spec parse(binary()) -> {ok, [poi()]}.
 parse(Text) ->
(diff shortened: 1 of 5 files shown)


src/els_parser.erl
```diff
@@ -18,39 +18,21 @@
...
-  %% so let's use a temporary file.
-  %% TODO: Ideally avoid writing to file at all (require epp changes)
-  TmpPath = tmp_path(),
-  try
-    ok = file:write_file(TmpPath, Text),
-    parse_file(TmpPath)
-  after
-    file:delete(TmpPath)
-  end.
+  Pid = els_io_string:new(Text),
```
robertoaloi: Surprising how an official behaviour does not exist for this...

PR #250, opened 2019-11-20
[#244] Exclude some OTP applications from indexing

### LINKED ISSUE #244: Be able to exclude OTP applications from indexing
Some of the OTP applications are heavy and rarely used by end users. A notable example is represented by the [megaco](http://erlang.org/doc/man/megaco.html) application. We could introduce a default exclusion list for OTP applications that are not part of the indexing. It should be possible to configure this exclusion list via the `erlang_ls.config` file at project and (maybe, not necessary for this ticket) at installation level (e.g. via a `~/.config/erlang_ls` file or similar.

References:

https://bugs.erlang.org/browse/ERL-1097

jfacorro:
### Description

Configuration option to exclude OTP application from the indexing process.

Fixes #244.

### DIFF
--- a/src/els_config.erl
+++ b/src/els_config.erl
@@ -22,6 +22,13 @@
 %% Macros
 %%==============================================================================
 -define(DEFAULT_CONFIG_PATH, "erlang_ls.config").
+-define( DEFAULT_EXCLUDED_OTP_APPS
+       , [ "megaco"
+         , "diameter"
+         , "snmp"
+         , "wx"
+         ]
+       ).
 -define(SERVER, ?MODULE).
 
 %% TODO: Refine names to avoid confusion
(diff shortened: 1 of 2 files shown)


src/els_config.erl
```diff
@@ -23,6 +23,13 @@
 %%==============================================================================
 -define(DEFAULT_CONFIG_PATH, "erlang_ls.config").
 -define(SERVER, ?MODULE).
+-define( DEEFAULT_EXCLUDED_OTP_APPS
```
robertoaloi: Typo: DEEFAULT

robertoaloi: Also, please keep them in lexicographical order.

src/els_config.erl
```diff
@@ -56,9 +65,17 @@ initialize(RootUri, Capabilities, InitOptions) ->
...
                                         ])),
-  OtpPath = maps:get("otp_path", Config, code:root_dir()),
-  DepsDirs = maps:get("deps_dirs", Config, []),
-  IncludeDirs = maps:get("include_dirs", Config, ["include"]),
+  OtpPath        = maps:get("otp_path", Config, code:root_dir()),
+  DepsDirs       = maps:get("deps_dirs", Config, []),
+  IncludeDirs    = maps:get("include_dirs", Config, ["include"]),
+  OtpAppsExclude = maps:get( "otp_apps_exclude"
+                           , Config
+                           , ?DEEFAULT_EXCLUDED_OTP_APPS
```
robertoaloi: And same typo here.

src/els_config.erl
jfacorro: 🤦‍♂ Fixed

PR #251, opened 2019-11-20
[#245] flush references db on update

### LINKED ISSUE #245: Old references are not flushed when re-indexing a file
**Describe the bug**

If a source file is opened, edited and saved, the previous references are not flushed, and we end up with multiple references for a given `poi`, some of which are stale.

**To Reproduce**

This bug is most clearly shown if using the document highlighting from #233 , which uses the `references` database.

Open a file, insert some blank lines between any two functions, and save it again.  The `references` database will now have two entries for each of the items indexed below the inserted lines.
 
**Expected behavior**

We only ever have a single reference location for a given usage.

**Actual behavior**

We end up with multiple entries, a new one being added each time we save the file with lines added or removed.

**Context**
 - `erlang_ls` version (tag/sha): f174bd2ecb6c972f65b6870cf8d24055c11b2854 
 - Editor used: emacs 26.3
 - LSP client used: lsp-mode

alanz:
### Description

When indexing a file, remove any references to the previous iteration of the file before adding the current ones.

Fixes #245.

### DIFF
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -2,6 +2,7 @@
 
 %% API
 -export([ find/2
+        , find_multi/2
         , keys/1
         , list/1
         , store/3
@@ -18,10 +19,10 @@
         ]).
 
 -define(SERVER, ?MODULE).
--define(TABLES, [ documents
-                , modules
-                , references
-                , signatures
+-define(TABLES, [ {documents,  set}
+                , {modules,    set}
+                , {references, duplicate_bag}
+                , {signatures, set}
                 ]).
 
 -type state() :: #{}.
@@ -41,6 +42,15 @@ find(Table, Key) ->
   case ets:lookup(Table, Key) of
     [] -> {error, not_found};
     [{Key, Value}] -> {ok, Value}
+  end.
+
+-spec find_multi(table(), key()) -> {ok, nonempty_list(any())}
+                                    | {error, not_found}.
+find_multi(Table, Key) ->
+  case ets:lookup(Table, Key) of
+    [] -> {error, not_found};
+    KVs -> {ok, KVs} %% Return as is, we need to iterate the result
+                     %% anyway, no point doing it twice.
   end.
 
 -spec keys(table()) -> [any()].
@@ -70,6 +80,8 @@ update(Table, Key, UpdateFun, Default) ->
       Replace = [{{Key, Old}, [], [{const, {Key, New}}]}],
       case ets:select_replace(Table, Replace) of
         1 -> ok;
+        %% AZ: The next line is a recursive call. If it is ever hit,
+        %%     will we have an infinite loop?
         0 -> update(Table, Key, UpdateFun, Default)
       end
   end.
@@ -81,8 +93,8 @@ delete(Table, Key) ->
 
 -spec flush_all_tables() -> ok.
 flush_all_tables() ->
-  [delete_table(Name) || Name <- ?TABLES],
-  [create_table(Name) || Name <- ?TABLES],
+  [delete_table(Details) || Details <- ?TABLES],
+  [create_table(Details) || Details <- ?TABLES],
   ok.
 
 %%==============================================================================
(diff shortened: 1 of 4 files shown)


src/els_indexer.erl
```diff
@@ -211,3 +212,28 @@ register_reference(Uri, #{id := {F, A}, range := Range}) ->
 add_reference(Key, Value) ->
   UpdateFun = fun(OldRefs) -> ordsets:add_element(Value, OldRefs) end,
   ok = els_db:update(references, Key, UpdateFun, ordsets:new()).
+
+%% @edoc Remove all references to a given uri()
+-spec purge_reference_uri(uri()) -> ok.
```
jfacorro: We are removing every object and reinserting all of the ones that are not pointing to the URI. This is a lot of work to do every time a file is reindexed (which currently happens on every file change).

An alternative would be to change the way we store the references. Instead of having a table of type `set` we could change `references` to be a `bag` or `duplicate_bag` table and then use something like `ets:select_delete/2` to remove all objects where the `uri` is the one we want to remove.

```
MatchSpec = ets:fun2ms(fun({K, #{uri := U}}) -> U =:= Uri end),
_DeletedCount = ets:select_delete(references, MatchSpec),
ok.
```

Changing the `references` table will require changes in the indexing here and in the `els_references_provider`.

WDYT?

alanz: I agree.  I see this current PR as a draft to expose the requirements and get some feedback.

src/els_db.erl
```diff
@@ -2,11 +2,13 @@
 
 %% API
 -export([ find/2
+        , find_multi/2
         , keys/1
         , list/1
         , store/3
         , update/4
         , delete/2
+        , flush_all_tables/0
```
jfacorro: This function is only used in tests, but I wonder if it is really needed since the `uri` used in `flush_references_db` test is defined by us and could be something that does not coincide with any of the other files used in the tests.

src/els_indexer.erl
```diff
@@ -204,7 +209,13 @@ register_reference(Uri, #{id := {F, A}, range := Range}) ->
...
--spec add_reference(any(), any()) -> ok.
+-spec add_reference(ref_key(), ref_value()) -> ok.
 add_reference(Key, Value) ->
-  UpdateFun = fun(OldRefs) -> ordsets:add_element(Value, OldRefs) end,
-  ok = els_db:update(references, Key, UpdateFun, ordsets:new()).
+  ok = els_db:store(references, Key, Value).
+
+%% @edoc Remove all references to a given uri()
+-spec purge_reference_uri(uri()) -> ok.
+purge_reference_uri(Uri) ->
```
jfacorro: Nit: Should this function be named in the plural (i.e. `purge_references_uri`)?

test/els_references_SUITE.erl
```diff
@@ -148,6 +149,24 @@ export_entry(Config) ->
   assert_locations(Locations, ExpectedLocations),
   ok.
 
+%% Issue #245
+-spec flush_references_db(config()) -> ok.
+flush_references_db(_Config) ->
```
jfacorro: Nit: Maybe this test case should be named `purge_references` since that's what the new feature is doing, and it's not so related to the DB itself IMHO.

jfacorro:
@alanz I added some comments, but they can be addressed in a different PR if you want and we can merge this one. Let me know what you think.

src/els_db.erl
alanz: True. It will slightly complicate the tests though, because `flush_references_db` test uses both

* `els_db:list(modules)`
* `els_db:list(references)`

which are otherwise full of random other things.

src/els_indexer.erl
alanz: I am renaming it to `purge_uri_references`.

src/els_db.erl
jfacorro: I was thinking that `?assertEqual([{foo,<<"file://tmp/foo.erl">>}],els_db:list(modules))` is maybe not so related to the test and `els_db:list(references)` could be replaced with a call to the new `els_db:find_multi/1`.

alanz:
I think (hope) this is ready to land.  Please do so, I do not have permission to.

jfacorro:
> I think (hope) this is ready to land. Please do so, I do not have permission to.

Awesome work! I think this is ready for merging. I have one small request though, would it be possible to rebase this branch off `master`, to avoid having the merge commit in the history?

robertoaloi:
Looks good to me as well. Please rebase and then we can merge it. Thanks for the help!

PR #254, opened 2019-11-20
Enable compression for ETS tables

jfacorro:
### Description

Enabling the compression for ETS tables reduces the memory usage quite significantly. The memory used when compressing is roughly 40% of the uncompressed usage.


jfacorro:
@seriyps FYI

PR #261, opened 2019-11-21
[#235] Single pass over forms

### LINKED ISSUE #235: Fork the epp_dodger module
Having a fork of the epp_dodger will allow us to:

* Hook into the internal parser, to avoid the double pass in `els_parser`
* Edit the dodger to operate on text, rather than on binaries
* Avoid the `apply` in `els_parser` due to Dialyzer spec issues

jfacorro:
### Description

Expose functions from `els_dodger` to be able to make a single pass over all the forms. This improves the performance a bit, although not as much as we would have expected.

Fixes #235.

The following are flame graphs captured by running `eflame:apply(els_indexer, index_dir, ["src"]).` and then generating the SVG with the following command:
```
export NAME=stacks; cat $NAME.out | sort | grep -v "<0.349.0>" > "$NAME.out.sort"; _build/debug/lib/eflame/stack_to_flame.sh < $NAME.out.sort > $NAME.svg; open -a Google\ Chrome $NAME.svg
```

#### Before
before

#### After
after

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -34,6 +34,7 @@
                    %% upstream version:
                    %% https://github.com/massemanet/redbug/pull/10
                  , [ {redbug, {git, "https://github.com/robertoaloi/redbug.git", {branch, "relax-beam-lib-error-matching"}}}
+                   , {eflame, {git, "https://github.com/jfacorro/eflame.git", {branch, "various.improvements"}}}
                    ]
                  }
                , { escript_emu_args
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -20,8 +20,8 @@
 %%==============================================================================
 -spec parse(binary()) -> {ok, [poi()]}.
 parse(Text) ->
-  Pid = els_io_string:new(Text),
-  parse_file(Pid).
+  IoDevice = els_io_string:new(Text),
+  parse_file(IoDevice).
 
 -spec parse_file(file:io_device()) -> {ok, [poi()]}.
 parse_file(IoDevice) ->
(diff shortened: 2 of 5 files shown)


rebar.config
```diff
@@ -34,6 +34,7 @@
                    %% upstream version:
                    %% https://github.com/massemanet/redbug/pull/10
                  , [ {redbug, {git, "https://github.com/robertoaloi/redbug.git", {branch, "relax-beam-lib-error-matching"}}}
+                   , {eflame, {git, "https://github.com/jfacorro/eflame.git", {branch, "various.improvements"}}}
```
robertoaloi: I would prefer to bring in the upstream version, if possible.
What are the included changes?
Are you pull-requesting them to upstream?

src/els_parser.erl
```diff
@@ -25,46 +25,47 @@ parse(Text) ->
...
-        {ok, Form} ->
-          POIs = find_attribute_pois(Form, Tokens),
-          parse_attribute_pois(IoDevice, POIs ++ Acc, EndLoc);
-        {error, _Error} ->
-          parse_attribute_pois(IoDevice, Acc, EndLoc)
+-spec parse_form(file:io_device(), any(), function(), [any()]) ->
+  {'ok', erl_syntax:forms() | none, integer()}
+  | {'eof', integer()}
+  | {'error', any(), integer()}.
+parse_form(Dev, L0, Parser, _Options) ->
```
robertoaloi: Can we please use the original naming for the variables? E.g. `IoDevice` instead of `Dev`, `StartLoc` instead of `L0`, `Tokens` instead of `Ts`, etc?

src/els_parser.erl
```diff
@@ -25,46 +25,47 @@ parse(Text) ->
...
+  case io:scan_erl_form(Dev, "", L0) of
+    {ok, Ts, L1} ->
+      case catch {ok, Parser(Ts, undefined)} of
```
robertoaloi: Please don't use the deprecated `catch`. Use `try/catch` instead.

src/els_parser.erl
```diff
@@ -25,46 +25,47 @@ parse(Text) ->
...
+%% Internal Functions
+%%==============================================================================
+
 -spec find_attribute_pois(erl_parse:abstract_form(), [erl_scan:token()]) ->
    [poi()].
 find_attribute_pois(Form, Tokens) ->
   case erl_syntax:type(Form) of
     attribute ->
-      case erl_syntax_lib:analyze_attribute(Form) of
+      case catch erl_syntax_lib:analyze_attribute(Form) of
```
robertoaloi: Same here. Please don't use the old-style `catch`.

src/els_parser.erl
```diff
@@ -25,46 +25,47 @@ parse(Text) ->
...
 
-%%==============================================================================
-%% Internal Functions
-%%==============================================================================
+-spec parse_form(file:io_device(), any(), [any()]) ->
+    {'ok', erl_syntax:forms()
+  | none, integer()}
+  | {'eof', integer()}
+  | {'error', any(), integer()}.
+parse_form(Dev, L0, Options) ->
```
robertoaloi: This seems a re-implementation of `els_dodger:parse_form/3`. Why not to expose that one, instead?

robertoaloi: Sorry, I was misreading the code. Ignore this comment.

robertoaloi: An alternative option would have been to add a callback function to the `Options` list which could have been executed from within the `els_dodger`, and have the `parse_form/4` return the final _accumulator_ together with the list of forms.

src/els_parser.erl
```diff
@@ -25,46 +25,47 @@ parse(Text) ->
...
       end;
-    {eof, _} ->
-      {ok, Acc};
-    {error, ErrorInfo, EndLoc} ->
-      lager:warning( "Could not parse extra information [end_loc=p] ~p"
-                   , [EndLoc, ErrorInfo]
-                   ),
-      {ok, Acc}
+    {error, _IoErr, _L1} = Err -> Err;
+    {error, _Reason} -> {eof, L0}; % This is probably encoding problem
```
robertoaloi: I know the (missing an article) comment comes from the dodger, but I would get rid of it and, instead, add a comment to the top of this function explaining this is a simplified version of the respective dodger function.

src/els_parser.erl
```diff
@@ -114,11 +115,8 @@ do_find_spec_points_of_interest(Tree, Acc) ->
 
 -spec points_of_interest(tree()) -> [poi()].
 points_of_interest(Tree) ->
-  lists:flatten(
-    erl_syntax_lib:fold(
-      fun(T, Acc) ->
-          [do_points_of_interest(T)|Acc]
-      end, [], Tree)).
+  FoldFun = fun(T, Acc) -> do_points_of_interest(T) ++ Acc end,
```
robertoaloi: Do we need the `++` in here?

src/els_parser.erl
```diff
@@ -25,46 +25,47 @@ parse(Text) ->
...
+          {ok, [], L1}
       end;
-    {eof, _} ->
-      {ok, Acc};
-    {error, ErrorInfo, EndLoc} ->
-      lager:warning( "Could not parse extra information [end_loc=p] ~p"
-                   , [EndLoc, ErrorInfo]
-                   ),
-      {ok, Acc}
+    {error, _IoErr, _L1} = Err -> Err;
```
robertoaloi: Considering we are pattern matching on `{ok, ...}` above, maybe we can avoid the defensive style altogether here.

src/els_parser.erl
```diff
@@ -25,46 +25,47 @@ parse(Text) ->
...
+-spec parse_form(file:io_device(), any(), function(), [any()]) ->
+  {'ok', erl_syntax:forms() | none, integer()}
+  | {'eof', integer()}
+  | {'error', any(), integer()}.
+parse_form(Dev, L0, Parser, _Options) ->
+  case io:scan_erl_form(Dev, "", L0) of
+    {ok, Ts, L1} ->
+      case catch {ok, Parser(Ts, undefined)} of
+        {ok, F} ->
+          POIs = [find_attribute_pois(F, Ts), points_of_interest(F)],
```
robertoaloi: Now that we have the single pass, it probably makes more sense to check here whether we are dealing with an attribute and execute the `find_attribute_pois` function only in that case.

rebar.config
jfacorro: This is the PR with all the [improvements](https://github.com/proger/eflame/pull/17).

src/els_parser.erl
jfacorro: The check is done in the function itself, adding it here would only make this function harder to read and would bring no benefit IMO.

src/els_parser.erl
jfacorro: That would have required a change in the implementation of dodger which I was trying to avoid.

src/els_parser.erl
jfacorro: The result from this function is processed by the code in `els_dodger` which is pattern matching and handling each of these results though.

src/els_parser.erl
jfacorro: Do you mean at the end as in `parse/1` or at the end of the fold? If it's the former then I would agree, if it's the latter, I think it's much clearer what's happening if we use the `++`.

PR #265, opened 2019-11-22
[#264] Handle generic errors to avoid crashing the els_server process

### LINKED ISSUE #264: Server crashes on provider's gen_server timeouts
**Describe the bug**
While the indexing is at its peak CPU consumption, the server's provider might take some time to respond to requests, this results in a timeout which the `els_server` doesn't handle.

**To Reproduce**
Open a (big) project (with a lot of source files and deps in it) and while the indexing is consuming a lot of CPU, try to go to the definition of some funciton, "hopefully" it will cause a gen_server timeout.

**Expected behavior**
Return a `null` result or a notification saying the server is currently busy.

**Actual behavior**
The els_server process crashes.

**Context**
 - `erlang_ls` version (tag/sha): `61aa263`
 - Editor used: Emacs
 - LSP client used: lsp-mode

jfacorro:
### Description

Handle generic errors to avoid crashing the els_server process

Fixed #264.

### DIFF
--- a/src/els_methods.erl
+++ b/src/els_methods.erl
@@ -39,7 +39,15 @@
 dispatch(Method, Params, State) ->
   Function = method_to_function_name(Method),
   try do_dispatch(Function, Params, State)
-  catch error:undef -> not_implemented_method(Method, State)
+  catch
+    error:undef ->
+      not_implemented_method(Method, State);
+    Type:Reason ->
+      lager:error("Unexpected error [type=~p] [error=~p]", [Type, Reason]),
+      Error = #{ code    => ?ERR_UNKNOWN_ERROR_CODE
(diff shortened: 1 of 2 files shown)


src/els_methods.erl
```diff
@@ -39,7 +39,9 @@
 dispatch(Method, Params, State) ->
   Function = method_to_function_name(Method),
   try do_dispatch(Function, Params, State)
-  catch error:undef -> not_implemented_method(Method, State)
+  catch
+    error:undef       -> not_implemented_method(Method, State);
+    _:_               -> {error, <<"Unexpected error">>, State}
```
robertoaloi: Shouldn't we log the error?

jfacorro: Added log and fixed error returned.

PR #267, opened 2019-11-22
[#235] Catch the correct type of exception

### LINKED ISSUE #235: Fork the epp_dodger module

jfacorro:
### Description

`erl_syntax_lib:analyze_attribute/1` generates a `throw` exception, not an `error`.

Fixes #235.


jfacorro:
This is a simple fix so I'm going to just merge it.

Issue #218, opened 2019-11-09
PropEr tests occasionally fail


robertoaloi:
```
[{set,{var,1},{call,prop_statem,connect,[]}},{set,{var,2},{call,prop_statem,disconnect,[]}},{set,{var,3},{call,prop_statem,connect,[]}},{set,{var,4},{call,prop_statem,did_open,[<<102,105,108,101,58,47,47,47,116,109,112,47,99,46,101,114,108>>,<<101,114,108,97,110,103>>,0,<<66,97,114,44,66,97,114,44,66,97,114,44,66,97,114,44,102,111,111,44,102,111,111,44,66,97,114,44,34,98,97,122,34,44,34,98,97,122,34,44,102,111,111,46>>]}},{set,{var,5},{call,prop_statem,did_open,[<<102,105,108,101,58,47,47,47,116,109,112,47,99,46,101,114,108>>,<<101,114,108,97,110,103>>,0,<<102,111,111,44,34,98,97,122,34,44,66,97,114,44,66,97,114,44,102,111,111,44,102,111,111,44,34,98,97,122,34,44,102,111,111,44,66,97,114,44,102,111,111,46>>]}},{set,{var,6},{call,prop_statem,initialize,[<<102,105,108,101,58,47,47,47,116,109,112>>,[]]}},{set,{var,7},{call,prop_statem,did_save,[<<102,105,108,101,58,47,47,47,116,109,112,47,98,46,101,114,108>>]}},{set,{var,8},{call,prop_statem,did_close,[<<102,105,108,101,58,47,47,47,116,109,112,47,97,46,101,114,108>>]}},{set,{var,9},{call,prop_statem,disconnect,[]}},{set,{var,10},{call,prop_statem,connect,[]}},{set,{var,11},{call,prop_statem,disconnect,[]}},{set,{var,12},{call,prop_statem,connect,[]}},{set,{var,13},{call,prop_statem,initialize,[<<102,105,108,101,58,47,47,47,116,109,112>>,[]]}},{set,{var,14},{call,prop_statem,did_save,[<<102,105,108,101,58,47,47,47,116,109,112,47,97,46,101,114,108>>]}}]
Annotated History: [{{prop_statem,connect,0},
                     {#{connected => false,documents => [],shutdown => false},
                      {ok,<0.3598.1>}}},
                    {{prop_statem,disconnect,0},
                     {#{connected => true,documents => [],shutdown => false},
                      ok}},
[...]

PR #268, opened 2019-11-24
Condense handle_call/3 and reduce duplication in parameter handling

### LINKED ISSUE #144: Remove code duplication from erlang_ls_client
The code for handling requests is full of duplicates. Extract common parts into functions.

liamjjmcnamara:
### Description

Should not change any functionality, just a deduplication of much of the handle_call clauses. Parameter construction is still a little unwieldy, and the Method lookup could be improved, but this seems a step in the right direction. 
 
Fixes #144.

### DIFF
--- a/src/els_client.erl
+++ b/src/els_client.erl
@@ -78,45 +78,46 @@
                 ) ->
   ok.
 completion(Uri, Line, Char, TriggerKind, TriggerCharacter) ->
+  Opts = {Uri, Line, Char, TriggerKind, TriggerCharacter},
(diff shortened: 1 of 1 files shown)


liamjjmcnamara:
Unclear to me why the OTP21 build fails (works on my machine), possibly just needs a kick?

jfacorro:
> Unclear to me why the OTP21 build fails (works on my machine), possibly just needs a kick?
Yeah, we have an issue about that here #218. I retried the build and it's now ✅.

PR #269, opened 2019-11-24
[#92] Add code completion for record names

### LINKED ISSUE #92: Implement code completion for record names
This should be triggered by pressing the '#' character.

robertoaloi:
### Description

Fixes #92 .

### DIFF
--- a/src/els_completion_provider.erl
+++ b/src/els_completion_provider.erl
@@ -67,7 +67,12 @@ find_completion( _Prefix
                , ?COMPLETION_TRIGGER_KIND_CHARACTER
                , #{trigger := <<"?">>, document := Document}
                ) ->
-  macros(Document);
+  definitions(Document, define);
+find_completion( _Prefix
+               , ?COMPLETION_TRIGGER_KIND_CHARACTER
+               , #{trigger := <<"#">>, document := Document}
+               ) ->
+  definitions(Document, record);
 find_completion( Prefix
                , ?COMPLETION_TRIGGER_KIND_INVOKED
                , #{document := Document}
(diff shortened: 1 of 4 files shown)


src/els_completion_provider.erl
```diff
@@ -67,7 +67,12 @@ find_completion( _Prefix
                , ?COMPLETION_TRIGGER_KIND_CHARACTER
                , #{trigger := <<"?">>, document := Document}
                ) ->
-  macros(Document);
+  definitions(Document, define);
```
robertoaloi: Completion for records and macros (and potentially other entities) works exactly in the same way, so I collapsed the logic for both into one, instead of having code duplication. We can probably extend the same strategy to other entities, too.

src/els_completion_provider.erl
```diff
@@ -247,3 +259,13 @@ to_binary(X) when is_atom(X) ->
...
+
+%%==============================================================================
+%% Maps definition types to completion item kinds
+%%==============================================================================
+
+-spec completion_item_kind(def_type()) -> completion_item_kind().
+completion_item_kind(define) ->
+  ?COMPLETION_ITEM_KIND_CONSTANT;
+completion_item_kind(record) ->
+  ?COMPLETION_ITEM_KIND_STRUCT.
```
robertoaloi: A `struct` seemed to be the best fit for records. For alternative item kinds, see the `CompletionItemKind ` namespace at https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/

src/els_parser.erl
```diff
@@ -216,7 +216,7 @@ attribute(Tree) ->
           []
       end;
     {record, {Record, _Fields}} ->
-      [els_poi:new(Pos, record, atom_to_list(Record))];
+      [els_poi:new(Pos, record, Record)];
```
robertoaloi: The three changes to the parser ensure that both records and macros are handled consistently, allowing to re-use the same piece of code for code completion.

PR #270, opened 2019-11-24
[#196] Completion returns function/arity when in export list

### LINKED ISSUE #196: Function names in the export entry list are not completed correctly
**Describe the bug**
Function names typed in the export entry list are not completed correctly.

**To Reproduce**
Add an entry to an export list. The list of completions appears showing functions in the format `name/arity`.

**Expected behavior**
When pressing ENTER `name/arity` is inserted.

**Actual behavior**
When pressing ENTER `name()` is inserted.

**Context**
 - `erlang_ls` version (tag/sha): 
 - Editor used: Emacs
 - LSP client used:

jfacorro:
### Description

Completion returns function/arity when in export list.

Fixes #196.

### DIFF
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -56,7 +56,7 @@ parse_form(IoDevice, Location0, Parser, _Options) ->
           {ok, POIs, Location1}
       catch
         _:_ ->
-          {ok, [], Location1}
+          {ok, find_attribute_tokens(Tokens), Location1}
       end;
     {error, _IoErr, _Location1} = Err -> Err;
     {error, _Reason} -> {eof, Location0};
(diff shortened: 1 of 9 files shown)


src/els_parser.erl
```diff
@@ -56,7 +56,7 @@ parse_form(IoDevice, Location0, Parser, _Options) ->
           {ok, POIs, Location1}
       catch
         _:_ ->
-          {ok, [], Location1}
+          {ok, find_attribute_tokens(Tokens), Location1}
```
jfacorro: This is to handle the case when there is a syntax error in the export list, because an entry is not yet written but we still need the information about the section as one of the POIs.

```
-export([ foo/1
        ,  b
        ]).
```
```

PR #271, opened 2019-11-24
[#204] Missing instructions Emacs

### LINKED ISSUE #204: Some missing instructions and considerations for Emacs
We should include the custom var configuration needed for Emac's `lsp-mode` when using `erlang_ls` (i.e. setting the path to the `erlang_ls` binary).

Also mention the fact that the OTP that needs to be in the path for `erlang_ls` is `20+`.

jfacorro:
### Description

Missing instructions Emacs

Fixes #204.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,6 +64,11 @@ file (assuming the `erlang_ls` executable is available in your `PATH`:
     ;; Enable LSP automatically for Erlang files
     (add-hook 'erlang-mode-hook #'lsp)
 
+    ;; TCP connection timeout
+    (setq lsp-tcp-connection-timeout 500)
+    ;; Set path to erlang_ls escript (unless it's in your path)
+    (setq lsp-erlang-server-path "/path/to/erlang_ls")
+
     ;; Override the key bindings from erlang-mode to use LSP for completion
     (eval-after-load 'erlang
       '(define-key erlang-mode-map (kbd "C-M-i") #'company-lsp))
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -79,6 +84,9 @@ To enable logging on the client-side, just:
 
     (setq lsp-log-io t)
 
+Remember that the Erlang Language Server requires Erlang/OTP 21 or
```
jfacorro: This is also in the VS Code section but it's easy to miss and very important, so it's worth repeating here.

README.md
```diff
@@ -64,6 +64,11 @@ file (assuming the `erlang_ls` executable is available in your `PATH`:
     ;; Enable LSP automatically for Erlang files
     (add-hook 'erlang-mode-hook #'lsp)
 
+    ;; TCP connection timeout
+    (setq lsp-tcp-connection-timeout 500)
```
alanz: Can we not set this timeout in `lsp-erlang`?

jfacorro: We could, although it's not really necessary to do so. 

I removed this option form the  document in a later PR [here](https://github.com/erlang-ls/erlang_ls/pull/273), since this is more specific to what the Emacs user wants as a timeout in general and not specific for `erlang-ls` to work.

README.md
robertoaloi: Created #275 to expand a bit on this.

PR #273, opened 2019-11-25
[#155] Describe how to configure helm-lsp for workspace symbols

jfacorro:
### Description

Introduce sections in Emacs configuration.

Fixes #155.

PR #276, opened 2019-11-26
[#155] Restore keybinding override

robertoaloi:
### Description

These instructions were removed in #273 . I think this may be needed by a fair amount of users, so I would like to keep it for reference.


robertoaloi:
In retrospection, it looks like the `helm` instructions make this change not necessary anymore, so declining this PR.

PR #279, opened 2019-11-26
[#278] Report the server version on startup.

### LINKED ISSUE #278: Display exact server version on startup

To help with bug reporting, it is good to be able to know exactly what `erlang_ls` server a client is talking to.  This information is available in the git repo.

When the server starts up, it should emit a `window/showMessage` with the exact version it is running.
This can be generated using something like https://github.com/soranoba/rebar3_git_vsn
This version info should also go into the log on startup, if it is enabled.

alanz:
### Description

Reports the server version as a `window/showMessage` once initialisation is complete.

Also logs it, if logging is enabled.

Fixes #278.

### DIFF
--- a/src/els_methods.erl
+++ b/src/els_methods.erl
@@ -129,7 +129,17 @@ initialize(Params, State) ->
 
 -spec initialized(params(), state()) -> result().
 initialized(_Params, State) ->
-  {noresponse, State#{status => initialized}}.
+  %% Report to the user the server version
+  {ok,     App} = application:get_application(),
+  {ok, Version} = application:get_key(App, vsn),
+  lager:info("initialized: [App=~p] [Version=~p]", [App, Version]),
+  BinVersion = list_to_binary(Version),
+  Message = <<"Erlang LS version: ", BinVersion/binary>>,
+  Method  = <<"window/showMessage">>,
+  Params  = #{ type    => ?MESSAGE_TYPE_INFO
+             , message => Message
+             },
+  {notification, Method, Params, State#{status => initialized}}.
 
 %%==============================================================================
 %% shutdown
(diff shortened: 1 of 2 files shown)


src/els_methods.erl
```diff
@@ -129,7 +129,17 @@ initialize(Params, State) ->
...
 -spec initialized(params(), state()) -> result().
 initialized(_Params, State) ->
-  {noresponse, State#{status => initialized}}.
+  %% Report to the user the server version
+  {ok,     App} = application:get_application(),
+  {ok, Version} = application:get_key(App, vsn),
+  lager:info("initialized: [App=~p] [Version=~p]", [App, Version]),
+  BinVersion = list_to_binary(Version),
+  Message = <<"Erlang LS version: ", BinVersion/binary>>,
+  Method1 = <<"window/showMessage">>,
```
jfacorro: Nit: Why `Method1` and not just `Method`? 😸

alanz: Copy/paste, will clean it up

alanz:
The git version is actually exactly what I am keen on.

This PR plays nice with https://github.com/erlang-ls/vscode/pull/13, so you can know what backend you are running.

src/els_methods.erl
```diff
@@ -129,7 +129,17 @@ initialize(Params, State) ->
...
 -spec initialized(params(), state()) -> result().
 initialized(_Params, State) ->
-  {noresponse, State#{status => initialized}}.
+  %% Report to the user the server version
+  {ok,     App} = application:get_application(),
+  {ok, Version} = application:get_key(App, vsn),
+  lager:info("initialized: [App=~p] [Version=~p]", [App, Version]),
+  BinVersion = list_to_binary(Version),
+  Message = <<"Erlang LS version: ", BinVersion/binary>>,
+  Method  = <<"window/showMessage">>,
```
robertoaloi: There's also a `logMessage` method. Maybe that's less invasive than this one?

alanz: I debated doing that, but figured we can use this for now and switch if it becomes a problem.

I like seeing what version I am running as a message, because when working on the server its something you want to be sure of while testing.

PR #280, opened 2019-11-26
[#239] Resolve line length for compiler diagnostics

### LINKED ISSUE #239: Column ranges for compiler diagnostics are not correct
The compiler diagnostics report a hard-coded 1-character range in the lines where the diagnostic happens. Whilst this suffices in Emacs, diagnostics look weird in VSCode. Instead, the whole line should be added to the range and not just the first character.

jfacorro:
### Description

Resolve line length for compiler diagnostics, fix calculation for application's range.

Related to #239.

### DIFF
--- a/src/els_compiler_diagnostics.erl
+++ b/src/els_compiler_diagnostics.erl
@@ -53,18 +53,22 @@ diagnostics(Uri) ->
 %%==============================================================================
 -spec diagnostics([compiler_msg()], severity()) -> [diagnostic()].
 diagnostics(List, Severity) ->
-  lists:flatten([[ diagnostic(Line, Module, Desc, Severity)
+  lists:flatten([[ diagnostic(location(Line), Module, Desc, Severity)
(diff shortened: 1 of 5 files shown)


alanz:
This is a lot of work for not much result.

From the LSP Spec

> Range
>A range in a text document expressed as (zero-based) start and end positions. A range is comparable to a selection in an editor. Therefore the end position is exclusive. If you want to specify a range that contains a line including the line ending character(s) then use an end position denoting the start of the next line. For example:

```typescript
{
    start: { line: 5, character: 23 },
    end : { line 6, character : 0 }
}
```

Perhaps we should do this instead?

jfacorro:
I hadn't seen that, I like it. I did try something similar based on the following comment in the specs for the `Position` interface:
> 	 If the character value is greater than the line length it defaults back to the
>	 line length.

But it didn't work for the `lsp-mode`. 

I will implement the approach you suggest.

PR #281, opened 2019-11-27
[#196] Handle multiple exports sections

jfacorro:
### Description

Handle multiple exports sections

Related to #196.

### DIFF
--- a/src/els_completion_provider.erl
+++ b/src/els_completion_provider.erl
@@ -178,10 +178,12 @@ snippet_function_call(Function, Args0) ->
 -spec is_exports_entry(els_document:document(), non_neg_integer()) -> boolean().
 is_exports_entry(Document, Line) ->
   case els_document:points_of_interest(Document, [exports]) of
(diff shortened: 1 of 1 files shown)


src/els_completion_provider.erl
```diff
@@ -178,10 +178,12 @@ snippet_function_call(Function, Args0) ->
 -spec is_exports_entry(els_document:document(), non_neg_integer()) -> boolean().
 is_exports_entry(Document, Line) ->
   case els_document:points_of_interest(Document, [exports]) of
-    [#{range := #{from := {From, _}, to := {To, _}}}] ->
-      (From =< Line) andalso (Line =< To);
-    _ ->
-      false
+    [] -> false;
+    Exports ->
+      IsInside = fun(#{range := #{from := {From, _}, to := {To, _}}}) ->
```
robertoaloi: Nit: there's some duplication with the code in `els_poi:match_pos/2`. Maybe that API could be used here.

PR #282, opened 2019-11-28
[#260] Elvis diagnostics

### LINKED ISSUE #260: Add elvis as a source of diagnostics.

I'm always frustrated when I can't run my tests because the compile aborts for some silly style error.

Elvis diagnostics reported in real time (or at least on save), so I can fix them as I work.

alanz:
Generates Elvis diagnostics when a file is saved, so long as an Elvis configuration exists.

### Description

Add [elvis](https://hex.pm/packages/elvis_core) as a diagnostics source.

Fixes #260 .

### DIFF
--- a/priv/code_navigation/elvis.config
+++ b/priv/code_navigation/elvis.config
@@ -1,0 +1,22 @@
+[{ elvis
+ , [
+    { config,
+      [ #{ dirs    => ["src"]
+         , filter  => "*.erl"
+         , ruleset => erl_files
+         , rules   => [ {elvis_style, invalid_dynamic_call }
+                      , {elvis_style, line_length, #{limit => 80, skip_comments => false}}
+                      ]
+         , ignore => []
+         }
+      , #{ dirs    => ["."]
+         , filter  => "rebar.config"
+         , ruleset => rebar_config
+         }
+      , #{ dirs    => ["."]
+         , filter  => "elvis.config"
+         , ruleset => elvis_config
+         }
+      ]}
+   ]}
+].
--- a/rebar.config
+++ b/rebar.config
@@ -15,6 +15,7 @@
        , {yamerl,      "0.7.0"}
        , {docsh,       "0.7.2"}
        , {worker_pool, "4.0.1"}
+       , {elvis,       "0.4.2", {pkg, elvis_core}}
        ]
 }.
 
@@ -56,7 +57,6 @@
                    , {proper,         "1.3.0"}
                    , {proper_contrib, "0.2.0"}
                    , {coveralls,      "2.0.1"}
-                   , {elvis_core,     "0.4.2"}
                    ]
                  }
                , { pre_hooks
--- a/rebar.lock
+++ b/rebar.lock
@@ -1,26 +1,34 @@
 {"1.1.0",
-[{<<"cowlib">>,{pkg,<<"cowlib">>,<<"2.3.0">>},0},
+[{<<"aleppo">>,{pkg,<<"inaka_aleppo">>,<<"1.1.1">>},2},
+ {<<"cowlib">>,{pkg,<<"cowlib">>,<<"2.3.0">>},0},
  {<<"docsh">>,{pkg,<<"docsh">>,<<"0.7.2">>},0},
+ {<<"elvis">>,{pkg,<<"elvis_core">>,<<"0.4.2">>},0},
  {<<"getopt">>,{pkg,<<"getopt">>,<<"1.0.1">>},2},
  {<<"goldrush">>,{pkg,<<"goldrush">>,<<"0.1.9">>},1},
  {<<"jsx">>,{pkg,<<"jsx">>,<<"2.9.0">>},0},
+ {<<"katana_code">>,{pkg,<<"katana_code">>,<<"0.1.2">>},1},
  {<<"lager">>,{pkg,<<"lager">>,<<"3.6.8">>},0},
  {<<"providers">>,{pkg,<<"providers">>,<<"1.8.1">>},1},
  {<<"ranch">>,{pkg,<<"ranch">>,<<"1.7.1">>},0},
  {<<"redbug">>,{pkg,<<"redbug">>,<<"1.2.1">>},0},
  {<<"worker_pool">>,{pkg,<<"worker_pool">>,<<"4.0.1">>},0},
- {<<"yamerl">>,{pkg,<<"yamerl">>,<<"0.7.0">>},0}]}.
+ {<<"yamerl">>,{pkg,<<"yamerl">>,<<"0.7.0">>},0},
+ {<<"zipper">>,{pkg,<<"zipper">>,<<"1.0.1">>},1}]}.
 [
 {pkg_hash,[
+ {<<"aleppo">>, <<"B0F7D05A118C5EBC3A735D68E92A7BD5837FE53455495EDE37CF0AA8BA4912E3">>},
  {<<"cowlib">>, <<"BBD58EF537904E4F7C1DD62E6AA8BC831C8183CE4EFA9BD1150164FE15BE4CAA">>},
  {<<"docsh">>, <<"F893D5317A0E14269DD7FE79CF95FB6B9BA23513DA0480EC6E77C73221CAE4F2">>},
+ {<<"elvis">>, <<"ED5D861A38493B7750427DAECFE0D844001E017C9DBC1F7F8DE302DB6226BC24">>},
  {<<"getopt">>, <<"C73A9FA687B217F2FF79F68A3B637711BB1936E712B521D8CE466B29CBF7808A">>},
  {<<"goldrush">>, <<"F06E5D5F1277DA5C413E84D5A2924174182FB108DABB39D5EC548B27424CD106">>},
  {<<"jsx">>, <<"D2F6E5F069C00266CAD52FB15D87C428579EA4D7D73A33669E12679E203329DD">>},
+ {<<"katana_code">>, <<"347D42D5E89AC6C4BF8D02B64633AC1BFE14B7724B5910260D39EFEAFE737F01">>},
  {<<"lager">>, <<"897EFC7679BB82383448646C96768CDC4E747464DD18B999C7AACA485686B0DA">>},
  {<<"providers">>, <<"70B4197869514344A8A60E2B2A4EF41CA03DEF43CFB1712ECF076A0F3C62F083">>},
  {<<"ranch">>, <<"6B1FAB51B49196860B733A49C07604465A47BDB78AA10C1C16A3D199F7F8C881">>},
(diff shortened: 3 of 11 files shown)


rebar.config
```diff
@@ -15,6 +15,7 @@
        , {yamerl,      "0.7.0"}
        , {docsh,       "0.7.2"}
        , {worker_pool, "4.0.1"}
+       , {elvis,       "0.4.2", {pkg, elvis_core}}
```
robertoaloi: Why not just declare the dependency as `elvis_core`? Also, this is already part of the `test` profile. Now that we include it everywhere we should probably remove it from there.

priv/code_navigation/elvis.config
```diff
@@ -0,0 +1,30 @@
+[{ elvis
+ , [
+    { config,
+      [ #{ dirs    => ["src"]
+         , filter  => "*.erl"
+         , ruleset => erl_files
+         , rules   => [ {elvis_style, invalid_dynamic_call, #{ignore => [ els_compiler_diagnostics
```
robertoaloi: I guess the ignore rules come from the main `elvis.config`. These do not really make sense in the `priv` sample config, so I would remove them.

src/els_elvis_diagnostics.erl
```diff
@@ -0,0 +1,83 @@
...
+-spec diagnostics(uri()) -> [diagnostic()].
+diagnostics(Uri) ->
+  RelFile = els_utils:project_relative(Uri),
+
+  %% Note: elvis_core:rock_this requires a file path relative to the
+  %%       project root, formatted as a string.
+  WS = try elvis_core:rock_this(RelFile)
+       catch Err ->
+           lager:info("Elvis error.[Err=~p] ", [Err]),
+           ok  %% TODO: return a diagnostic saying elvis failed?
```
robertoaloi: Regarding the TODO: I would log this as a `lager:warning` or `lager:error` and call it a day for now.

src/els_elvis_diagnostics.erl
```diff
@@ -0,0 +1,83 @@
...
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+-spec format_diagnostics(any()) -> [map()].
+format_diagnostics(#{file := Path, rules := Rules}) ->
+  R = format_rules(Path, Rules),
+  lists:flatten(R).
+
+
+%%% This section is based directly on elvis_result:print_rules
```
robertoaloi: @jfacorro Most of the complexity in this PR seems to come from this function and the relative path handling. Would it make sense to add those to elvis itself? I.e. allowing absolute paths and separating formatting from outputting?

src/els_text_synchronization.erl
```diff
@@ -19,9 +19,10 @@ did_save(Params, Server) ->
   Uri          = maps:get(<<"uri">>         , TextDocument),
   CDiagnostics = els_compiler_diagnostics:diagnostics(Uri),
   DDiagnostics = els_dialyzer_diagnostics:diagnostics(Uri),
+  EDiagnostics = els_elvis_diagnostics:diagnostics(Uri),
   Method = <<"textDocument/publishDiagnostics">>,
   Params1  = #{ uri => Uri
-              , diagnostics => CDiagnostics ++ DDiagnostics
+              , diagnostics => CDiagnostics ++ DDiagnostics ++ EDiagnostics
```
robertoaloi: Diagnostics are becoming complex. Should we allow users to customize their behaviour via the `erlang_ls.config` file? E.g. I may want to disable elvis, but keep compiler diagnostics.

priv/code_navigation/elvis.config
alanz: Yes, I just copied the main config over.

rebar.config
alanz: I copied this from the documentation, will simplify it.

src/els_text_synchronization.erl
alanz: I agree, for a follow-up PR.

And we should also consider what happens if we have a diagnostics source that takes longer to run, should we send some results when we have them, or only send them all when the slowest one finishes.

rebar.lock
```diff
@@ -1,26 +1,36 @@
 {"1.1.0",
-[{<<"cowlib">>,{pkg,<<"cowlib">>,<<"2.3.0">>},0},
+[{<<"aleppo">>,{pkg,<<"inaka_aleppo">>,<<"1.1.1">>},2},
+ {<<"cowlib">>,{pkg,<<"cowlib">>,<<"2.3.0">>},0},
  {<<"docsh">>,{pkg,<<"docsh">>,<<"0.7.2">>},0},
+ {<<"elvis">>,{pkg,<<"elvis_core">>,<<"0.4.2">>},0},
```
jfacorro: We should maybe remove this entry.

test/els_diagnostics_SUITE.erl
```diff
@@ -91,6 +92,41 @@ compiler(Config) ->
...
+      ?assert(maps:is_key(uri, Params)),
+      #{uri := Uri} = Params,
+      ?assert(maps:is_key(diagnostics, Params)),
+      #{diagnostics := Diagnostics} = Params,
+      ?assertEqual(2, length(Diagnostics)),
+      Warnings = [D || #{severity := ?DIAGNOSTIC_WARNING} = D <- Diagnostics],
+      Errors   = [D || #{severity := ?DIAGNOSTIC_ERROR}   = D <- Diagnostics],
+      ?assertEqual(2, length(Warnings)),
+      ?assertEqual(0, length(Errors)),
+      [#{range := WarningRange1}
```
jfacorro: Nit: missing space between `[` and `#`,  and also `,` and `#` in the next line.

src/els_elvis_diagnostics.erl
```diff
@@ -0,0 +1,83 @@
...
+%%==============================================================================
+-spec diagnostics(uri()) -> [diagnostic()].
+diagnostics(Uri) ->
+  RelFile = els_utils:project_relative(Uri),
+
+  %% Note: elvis_core:rock_this requires a file path relative to the
+  %%       project root, formatted as a string.
+  WS = try elvis_core:rock_this(RelFile)
+       catch Err ->
+           lager:warning("Elvis error.[Err=~p] ", [Err]),
```
jfacorro: Nit: `lager:warning/2` already returns `ok` if the logging goes fine.

src/els_elvis_diagnostics.erl
jfacorro: We would have to check how that impacts elvis code, since the reason I think the paths are relative is because it simplifies reporting the path when printing.

test/els_diagnostics_SUITE.erl
alanz: Can we configure elvis to complain about those?

src/els_elvis_diagnostics.erl
robertoaloi: @jfacorro Should we then proceed with this PR as it is and look at the elvis changes separately, then?

rebar.lock
robertoaloi: True.

rebar.lock
```diff
@@ -1,26 +1,36 @@
...
  {<<"worker_pool">>,{pkg,<<"worker_pool">>,<<"4.0.1">>},0},
- {<<"yamerl">>,{pkg,<<"yamerl">>,<<"0.7.0">>},0}]}.
+ {<<"yamerl">>,{pkg,<<"yamerl">>,<<"0.7.0">>},0},
+ {<<"zipper">>,{pkg,<<"zipper">>,<<"1.0.1">>},1}]}.
 [
 {pkg_hash,[
+ {<<"aleppo">>, <<"B0F7D05A118C5EBC3A735D68E92A7BD5837FE53455495EDE37CF0AA8BA4912E3">>},
  {<<"cowlib">>, <<"BBD58EF537904E4F7C1DD62E6AA8BC831C8183CE4EFA9BD1150164FE15BE4CAA">>},
  {<<"docsh">>, <<"F893D5317A0E14269DD7FE79CF95FB6B9BA23513DA0480EC6E77C73221CAE4F2">>},
+ {<<"elvis">>, <<"ED5D861A38493B7750427DAECFE0D844001E017C9DBC1F7F8DE302DB6226BC24">>},
```
robertoaloi: And this one I guess?

src/els_elvis_diagnostics.erl
```diff
@@ -0,0 +1,83 @@
...
+           ok
+       end,
+  case WS of
```
robertoaloi: @alanz Would you mind avoiding the `case` after `try`? We could simply `try ... of ... catch`.

src/els_utils.erl
```diff
@@ -66,6 +68,25 @@ start_epmd() ->
...
+  Path = els_uri:path(Uri),
+
+  RootDir = els_config:get(root_uri),
+  RootList = filename:split(els_uri:path(RootDir)),
+  PathList = filename:split(Path) ,
+  RelFileBin = filename:join(els_utils:strip_prefix(RootList, PathList)),
+  RelFile = binary_to_list(RelFileBin),
+  RelFile.
+
+%% @doc Strips the prefix in List1 from List2
```
robertoaloi: Isn't there an OTP API for this?

src/els_elvis_diagnostics.erl
jfacorro: Yes, I don't think we should wait on changes on elvis, if the code in this PR works.

src/els_utils.erl
alanz: I looked for one, without success. If there is one, I will gladly use it.

test/els_diagnostics_SUITE.erl
jfacorro: Space after commas should be the default, not sure why it wasn't detected. Spaces before `#`  wouldn't make much sense IMO.

rebar.lock
alanz: I get

```
===> Dialyzer starting, this may take a while...
===> Updating plt...
===> Could not find application: elvis
```
If I take it out.   Which probably means it needs to go back in the `rebar.config` too?

src/els_utils.erl
jfacorro: `string:prefix/2` removes the prefix if it is there otherwise returns `nomatch`.

alanz: Thanks. And I presume we don't have to worry about path separators, and platform issues? As we will always be working with the same file system while doing this.

jfacorro: That is a good point, but as you say both the file's URI and the root's URI have the same format since they are defined based on the same filesystem. We are using `els_uri:path/1` which just strips out leading `file://` prefix, so I think we can safely rely on `string:prefix/2` to work, unless I'm missing something 😄

PR #284, opened 2019-12-02
[#283] Don't fail with function name as macro

### LINKED ISSUE #283: Using macro to define function in name generates crash
**Describe the bug**
The parser can't handle code that uses a module in the name of a function definition.

**To Reproduce**
```
els_parser:parse(<<"?X() -> ok.">>).
** exception throw: syntax_error
     in function  erl_syntax_lib:analyze_function/1 (erl_syntax_lib.erl, line 1888)
     in call from els_parser:function/1 (/Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_parser.erl, line 251)
     in call from els_parser:'-points_of_interest/1-fun-0-'/2 (/Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_parser.erl, line 135)
     in call from els_parser:parse_form/4 (/Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_parser.erl, line 55)
     in call from els_dodger:parse/5 (/Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_dodger.erl, line 343)
     in call from els_parser:parse_file/1 (/Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_parser.erl, line 28)
```

**Expected behavior**
Ignore the unparsable form and index the rest of the document.

**Actual behavior**
The indexing of the document fails and it's not available for other calls.

**Context**
 - `erlang_ls` version (tag/sha): `8b239c7`
 - Editor used: Emacs
 - LSP client used: lsp-mode

jfacorro:
### Description

Re-add the `try...catch` for `syntax_error`s. 

Fixes #283.

### DIFF
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -138,16 +138,19 @@ points_of_interest(Tree) ->
 %% @edoc Return the list of points of interest for a given `Tree`.
 -spec do_points_of_interest(tree()) -> [poi()].
 do_points_of_interest(Tree) ->
-  case erl_syntax:type(Tree) of
-    application   -> application(Tree);
-    attribute     -> attribute(Tree);
-    function      -> function(Tree);
-    implicit_fun  -> implicit_fun(Tree);
-    macro         -> macro(Tree);
-    record_access -> record_access(Tree);
-    record_expr   -> record_expr(Tree);
-    variable      -> variable(Tree);
-    _             -> []
+  try
+    case erl_syntax:type(Tree) of
+      application   -> application(Tree);
+      attribute     -> attribute(Tree);
+      function      -> function(Tree);
+      implicit_fun  -> implicit_fun(Tree);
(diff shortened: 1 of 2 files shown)


src/els_parser.erl
```diff
@@ -138,16 +138,19 @@ points_of_interest(Tree) ->
...
+      attribute     -> attribute(Tree);
+      function      -> function(Tree);
+      implicit_fun  -> implicit_fun(Tree);
+      macro         -> macro(Tree);
+      record_access -> record_access(Tree);
+      record_expr   -> record_expr(Tree);
+      variable      -> variable(Tree);
+      _             -> []
+    end
+  catch throw:syntax_error -> []
```
robertoaloi: Should we log the event so we are aware?

jfacorro: We could show it as a warning.

PR #287, opened 2019-12-03
[#286] Do not pass els_server's PID around

### LINKED ISSUE #286: Do not pass els_server's PID around, since it's registered

robertoaloi:
### Description

Fixes #286 .

### DIFF
--- a/src/els_server.erl
+++ b/src/els_server.erl
@@ -22,7 +22,7 @@
         ]).
 
 %% API
--export([ process_requests/2
+-export([ process_requests/1
         , set_connection/1
         , send_notification/2
         ]).
@@ -64,9 +64,9 @@ start_link(Transport) ->
   {ok, _} = Transport:start_listener(Pid),
   {ok, Pid}.
 
--spec process_requests(pid(), [any()]) -> ok.
-process_requests(Server, Requests) ->
-  gen_server:cast(Server, {messages, Requests}).
+-spec process_requests([any()]) -> ok.
+process_requests(Requests) ->
+  gen_server:cast(?SERVER, {process_requests, Requests}).
 
 -spec set_connection(any()) -> ok.
 set_connection(Connection) ->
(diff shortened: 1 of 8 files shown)


src/els_server.erl
```diff
@@ -22,27 +22,32 @@
...
 %% Record Definitions
 %%==============================================================================
--record(state, { transport  :: module()
-               , connection :: any()
-               , buffer     :: binary()
-               , state      :: map()
+-record(state, { transport      :: module()
+               , connection     :: any()
+               , buffer         :: binary()
+               , internal_state :: map()
```
robertoaloi: Renamed to clarify the distinction between state and internal state. I will further cleanup this module in a following PR.

src/els_server.erl
```diff
@@ -55,60 +60,62 @@
...
+  {ok, Pid} = gen_server:start_link({local, ?SERVER}, ?MODULE, Transport, []),
   {ok, _} = Transport:start_listener(Pid),
   {ok, Pid}.
 
--spec process_requests(pid(), [any()]) -> ok.
-process_requests(Server, Requests) ->
-  gen_server:cast(Server, {messages, Requests}).
+-spec process_requests([any()]) -> ok.
+process_requests(Requests) ->
+  gen_server:cast(?SERVER, {process_requests, Requests}).
```
robertoaloi: Renamed for consistency.

src/els_stdio.erl
```diff
@@ -42,7 +42,7 @@ loop(Lines, IoDevice, Server, JsonOpts) ->
       %% Use file:read/2 since it reads bytes
       {ok, Payload} = file:read(IoDevice, Length),
       Request       = jsx:decode(Payload, JsonOpts),
-      els_server:process_requests(Server, [Request]),
+      gen_server:cast(Server, {process_requests, [Request]}),
```
robertoaloi: Well, this was reeeeally confusing (and I think it could have something to do with the occasional timeouts we see in CI). We were calling:

```
els_server:process_requests(Server, [Request]),
```

But this function is actually called in two separate scenarios in the case of a stdio server: sometimes it calls the `els_server` (when sending requests), sometimes the client (when sending responses). In the latter case, the `els_client` implements its own `handle_cast` to handle this request. These two things are too tightly coupled and I plan to restructure this part. For now, I'm making sure we don't call `els_server` but we make an explicit cast to lower confusion.

jfacorro: Nice! Doesn't the same apply in `els_tcp`, though? I think we are actually only using TCP connections in the proper testing.

robertoaloi: In `els_client`, there is a special function clause for stdio (see `handle_cast/2`). The requests over tcp are simply dropped.

jfacorro: I guess the current approach is a little hacking, my bad 😿 .

robertoaloi: No worries! Your contribution actually allowed us to integrate with clients not supporting TCP. We can still clean things up if necessary!

PR #291, opened 2019-12-03
[#289] Handle nomatch for string:prefix/2

### LINKED ISSUE #289: Elvis Diagnostics cause server crash
**Describe the bug**

```
{badarg,[{erlang,binary_to_list,[nomatch],[]},{els_utils,project_relative,1,[{file,"/.../erlang_ls/src/els_utils.erl"},{line,77}]},{els_elvis_diagnostics,diagnostics,1,[{file,"/.../erlang_ls/src/els_elvis_diagnostics.erl"},{line,26}]},{els_text_synchronization,did_save,2,[{file,"/.../erlang_ls/src/els_text_synchronization.erl"},{line,22}]}]}
```

jfacorro:
### Description

Handle `nomatch` result for `string:prefix/2`. Depending on what the client provides as a `root_uri`, the value might or might not include a trailing `/`, we should handle both cases.

Fixes #289.

### DIFF
--- a/src/els_utils.erl
+++ b/src/els_utils.erl
@@ -70,12 +70,10 @@ start_epmd() ->
 %% @doc Returns a project-relative file path for a given URI
 -spec project_relative(uri()) -> file:filename().
 project_relative(Uri) ->
-  Path = els_uri:path(Uri),
-  RootDir = els_config:get(root_uri),
-  RootPath = els_uri:path(RootDir),
-  RelativeBin = string:prefix(string:prefix(Path, RootPath), <<"/">>),
-  binary_to_list(RelativeBin).
-
+  RootUri     = els_config:get(root_uri),
+  RootUriSize = byte_size(RootUri),
+  <<RootUri:RootUriSize/binary, RelativePath/binary>> = Uri,
+  binary_to_list(string:trim(RelativePath, leading, [$/, $\\ ])).
 
 %%==============================================================================
 %% Internal functions


src/els_utils.erl
```diff
@@ -70,17 +70,23 @@ start_epmd() ->
...
 -spec project_relative(uri()) -> file:filename().
 project_relative(Uri) ->
-  Path = els_uri:path(Uri),
-  RootDir = els_config:get(root_uri),
-  RootPath = els_uri:path(RootDir),
-  RelativeBin = string:prefix(string:prefix(Path, RootPath), <<"/">>),
+  Path        = els_uri:path(Uri),
+  RootUri     = els_config:get(root_uri),
+  RootPath    = els_uri:path(RootUri),
+  RelativeBin = prefix(prefix(Path, RootPath), <<"/">>),
```
robertoaloi: Uhm, this will not work on Windows, will it?

alanz: My previous version had the benefit of using file-system aware split and join.

But a colleague told me Erlang works with `/` as a path separator internally, but then switches it when it hits the file system.  Which won't work if the URL comes from non-Erlang.

alanz: Maybe we should revert to my original version? We have an internal utility function now anyway, so no real saving.

And at least it intentionally operates on file paths, rather than strings.

jfacorro: But aren't URIs file system independent and always expressed using `/`?

robertoaloi: What about something like:

```
%% @doc Returns a project-relative file path for a given URI
-spec project_relative(uri()) -> file:filename().
project_relative(Uri) ->
  RootUri     = els_config:get(root_uri),
  RootUriSize = byte_size(RootUri),
  <<RootUri:RootUriSize/binary, RelativePath/binary>> = Uri,
  binary_to_list(filename:basename(RelativePath)).
```

robertoaloi: Except that does not seem to work, let me try again... :)

jfacorro:
I tried adding a build in a windows machine but AppVeyor only includes OTP 20 in their Windows images 😞.

src/els_utils.erl
jfacorro: > Maybe we should revert to my original version? We have an internal utility function now anyway, so no real saving.
> 
> And at least it intentionally operates on file paths, rather than strings.

That is a good point 😜. I'm open to using the original version.

robertoaloi: Btw, the fact this exists is not comforting...

https://github.com/erlang/rebar3/blob/master/src/rebar_file_utils.erl#L429

robertoaloi:
```
  RootUri     = els_config:get(root_uri),
  RootUriSize = byte_size(RootUri),
  <<RootUri:RootUriSize/binary, RelativePath/binary>> = Uri,
  binary_to_list(string:trim(RelativePath, leading, [$/, $\\])).
```

robertoaloi:
@jfacorro That would have actually been nice.

robertoaloi:
@alanz I proposed an alternative solution. Whatever you guys prefer, I'm happy with.

PR #293, opened 2019-12-03
[#292] Stricter operator_spaces checks for Elvis

### LINKED ISSUE #292: Refine operator_spaces rules
By default, Elvis looks for missing whitespaces only for a limited set of operators. Let's be a bit more strict.

robertoaloi:
### Description

I think @alanz raised this in one of his PRs and I believe it makes sense.

Fixes #292 .


robertoaloi:
@jfacorro Does this one require any modifications after your `elvis_core` bump?

jfacorro:
@robertoaloi No, it doesn't. The improvement should just work.

PR #297, opened 2019-12-04
[#296] Split client in tcp_client and stdio_client

### LINKED ISSUE #296: Separate TCP from STDIO client
At the moment, the `els_client` module contain the logic to support both transports. The module should be splitted in two.

robertoaloi:
### Description

Part of a long-due cleanup on how the two transports (tcp, stdio) are used by client and server.

Fixes #296 .

### DIFF
--- a/src/els_client.erl
+++ b/src/els_client.erl
@@ -3,10 +3,19 @@
 %%==============================================================================
 -module(els_client).
 
+-callback start_link(any())     -> {ok, pid()}.
+-callback send(pid(), iolist()) -> ok.
+
 %%==============================================================================
 %% Behaviours
 %%==============================================================================
 -behaviour(gen_server).
+%% gen_server callbacks
+-export([ init/1
+        , handle_call/3
+        , handle_cast/2
+        , code_change/3
+        ]).
 
 %%==============================================================================
 %% Exports
@@ -29,14 +38,7 @@
         , workspace_symbol/1
         ]).
 
-%% gen_server callbacks
--export([ init/1
-        , handle_call/3
-        , handle_cast/2
-        , handle_info/2
-        , terminate/2
-        , code_change/3
-        ]).
+-export([ handle_responses/1 ]).
 
 %%==============================================================================
 %% Includes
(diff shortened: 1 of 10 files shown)


src/els_client.erl
```diff
@@ -3,10 +3,19 @@
...
+
 %%==============================================================================
 %% Behaviours
 %%==============================================================================
 -behaviour(gen_server).
+%% gen_server callbacks
+-export([ init/1
+        , handle_call/3
+        , handle_cast/2
+        , code_change/3
```
robertoaloi: Removed `handle_info` since only relevant for TCP.

src/els_client.erl
```diff
@@ -29,14 +38,7 @@
...
 
-%% gen_server callbacks
--export([ init/1
-        , handle_call/3
-        , handle_cast/2
-        , handle_info/2
-        , terminate/2
-        , code_change/3
-        ]).
+-export([ handle_responses/1 ]).
```
robertoaloi: This is the callback which is used by the specific implementation when a message is available.

src/els_client.erl
```diff
@@ -47,25 +49,23 @@
...
 %%==============================================================================
 %% Record Definitions
 %%==============================================================================
--record(state, { transport :: tcp | stdio
-               , connection
-               , request_id = 1
-               , pending    = []
-               , buffer   = <<>>
+-record(state, { transport_cb     :: transport_cb()
+               , transport_server :: pid()
```
robertoaloi: This may change in the future, since the STDIO implementation is abusing this field to store the reference to the IODEVICE.

jfacorro: I'm not sure what this comment means 😄

robertoaloi: Yeah, not the clearest one. What I mean is that the field is named "transport_server" and it is supposed to store a Pid, but in the case of the stdio implementation we store the iodevice in there. And Dialyzer doesn't even seem to be able to figure out the discrepancy.

jfacorro: `iodevice()`s are `pid()`s AFAIK.

PR #298, opened 2019-12-05
[#015] Initial formatting support in place

### LINKED ISSUE #15: Code Formatting / Pretty Printing
It should be possible to format entire Erlang modules or given portions of code via the Erlang Language Server.

Several tools and APIs exist in the Erlang / Elixir ecosystem that should be considered, either for inspiration or as libraries. For example:

* `erl_tidy` (which seems to use `erl_prettypr`
* `Code.format_string` (Elixir)
* [erl_pp](https://github.com/erlang/otp/blob/master/lib/stdlib/src/erl_pp.erl)
* [erl_prettypr](http://erlang.org/doc/man/erl_prettypr.html)
* [io_lib_pretty](https://github.com/erlang/otp/blob/master/lib/stdlib/src/io_lib_pretty.erl) (apparently used by EDTS)
* [prettypr](http://erlang.org/doc/man/prettypr.html)

Ideally, some customizations should be possible. For example, one may prefer spaces over tabs or viceversa. Some organizations use a comma-first convention, whilst others don't.

Something that should also be investigated is how to treat code that does not compile (since some of the above tools work directly on the `.beam` files).

Useful References:
* The "Programming Elixir" book should have a section about pretty-printing
* [This talk](https://www.youtube.com/watch?v=x2ckfhqB9nA) by Jose Valim covers the code formatter in Elixir
* In ["The design of a pretty-print library"](http://belle.sourceforge.net/doc/hughes95design.pdf) John Hughes introduces an algebraic approach to pretty printing
* [This](https://github.com/fireflyc/erlang-vscode/commit/14f3a14af55258b9de83f7690b688cef063f97e2?diff=split
) is the PR where formatting has been included in `erlang-vscode`
* A [Medium Post](https://medium.com/blackode/code-formatter-the-big-feature-in-elixir-v1-6-0-f6572061a4ba) discussing the code formatter in Elixir 1.6.0
* The ["Strictly Pretty"](http://citeseerx.ist.psu.edu/viewdoc/summary?doi=10.1.1.34.2200) paper about pretty-printing
* [Demo usages](https://github.com/erlang/otp/blob/d6285b0a347b9489ce939511ee9a979acd868f71/lib/syntax_tools/examples/demo.erl) of the Erlang Syntax Tools libraries for pretty-printing

Useful snippets:
```erlang
c(test, [debug_info]).
{ok,{_,[{abstract_code,{_,AC}}]}} = beam_lib:chunks("test.beam",[abstract_code]),
io:put_chars(erl_prettypr:format(erl_syntax:form_list(AC))).
```

```
print(SyntaxTree, Paper, Ribbon) ->
    io:put_chars(erl_prettypr:format(SyntaxTree, [{paper, Paper},
						  {ribbon, Ribbon}])).
```

```
read(Name) ->
    {ok, Forms} = epp:parse_file(Name, [], []),
    Comments = erl_comment_scan:file(Name),
    erl_recomment:recomment_forms(Forms, Comments).
```

alanz:
### Description

Enables and routes  functions for document, and range and on-type formatting.

On-type formatting is not feasible without `textdocument/didChange` support in `erlang_ls`, but the plumbing is in place.

Uses `rebar3_format` as an initial provider for document formatting. This will be replaced by `erlfmt` once that is available, at the same time making the formatter used configurable.

Adresses #15.


rebar.config
```diff
@@ -16,7 +16,7 @@
        , {docsh,       "0.7.2"}
        , {worker_pool, "4.0.1"}
        , {elvis,       "0.4.3", {pkg, elvis_core}}
-       , {rebar3_format, "0.0.2"}
+       , {rebar3_format, {git, "https://github.com/alanz/rebar3_format.git", {branch, "master"}}}
```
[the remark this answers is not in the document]
alanz: The path construction is wrong too, in the dependency, and I need the second branch of the `case` statement, where it actually gets a value for the `output_dir` param.

See https://github.com/AdRoll/rebar3_format/pull/39/files

include/erlang_ls.hrl
```diff
@@ -545,6 +545,22 @@
                                   , active_parameters => number()
                                   }.
 
+%%------------------------------------------------------------------------------
+%% Formatting
+%%------------------------------------------------------------------------------
+-type formatting_options() :: #{ tabSize           := integer()
+                               , insertSpaces      := boolean()
+                               %% Spec says further properties must
```
robertoaloi: Could you convert this comment into actual type specs for the `furtherProperties`?

priv/code_navigation/src/diagnostics.new.erl
```diff
@@ -0,0 +1,7 @@
+-module(diagnostics).
```
robertoaloi: I would rename the `.new` in the filename into `_new` (also below).

rebar.config
```diff
@@ -16,6 +16,10 @@
        , {docsh,       "0.7.2"}
        , {worker_pool, "4.0.1"}
        , {elvis,       "0.4.3", {pkg, elvis_core}}
+       , {rebar3_format, { git, "https://github.com/alanz/rebar3_format.git"
```
robertoaloi: If your fixes are reviewed soon we can just wait for that to happen.
If not and we need a fork, I suggest we create forks under the `erlang-ls` organization, so we all have access.

src/els_formatting_provider.erl
```diff
@@ -0,0 +1,115 @@
...
+-include("erlang_ls.hrl").
+
+%%==============================================================================
+%% els_provider functions
+%%==============================================================================
+
+
+%% Keep the behaviour happy
+-spec is_enabled() -> boolean().
+is_enabled() -> is_enabled_document().
```
robertoaloi: Would it make sense for this to be an `orelse` between the three modes?

include/erlang_ls.hrl
alanz: I am not sure how it would be represented, to be honest.   I guess a list of (key,value) pairs, would sort of capture it.

I understand the intent to be that you can put additional formatting options into the structure, that are first-class fields.  i.e. they are not wrapped in any additional structure.

robertoaloi: Read the specifications myself. I would say we should remove the `furtherProperties` field altogether here, and add individual properties once handled.

src/els_formatting_provider.erl
alanz: Each of these flags is used to set a specific capability value back to the client during initialisation, so they need to be separate.

priv/code_navigation/src/diagnostics.new.erl
alanz: I used the `.` style to indicate that it is not a regular file, but is a different version of the `diagnostics.el` file,  which is used as input to the test for diff generation, between `diagnostics.erl` and `diagnostics.new.erl`.  So the module id should not change either.

src/els_methods.erl
```diff
@@ -100,16 +103,16 @@ initialize(Params, State) ->
   ok = els_provider:initialize(),
   Result =
     #{ capabilities =>
-         #{ hoverProvider => els_hover_provider:is_enabled()
-          , completionProvider =>
-              #{ resolveProvider => false
-               , triggerCharacters => [<<":">>, <<"#">>, <<"?">>]
-               }
-          , textDocumentSync =>
+         #{ textDocumentSync =>
```
robertoaloi: Any specific reason behind this change?

src/els_formatting_provider.erl
```diff
@@ -0,0 +1,115 @@
...
+-spec is_enabled_range() -> boolean().
+is_enabled_range() ->
+  false.
+
+%% NOTE: because erlang_ls does not send incremental document changes
+%%       via `textDocument/didChange`, this kind of formatting does not
+%%       make sense.
+-spec is_enabled_on_type() -> document_ontypeformatting_options().
+is_enabled_on_type() -> false.
+%% is_enabled_on_type() -> #{ <<"firstTriggerCharacter">> => "\n"
```
robertoaloi: Please remove the commented code.

src/erlang_ls.app.src
```diff
@@ -15,6 +15,9 @@
       , docsh
       , worker_pool
       , elvis
+      , ephemeral
```
robertoaloi: How come do we need the `rebar3_format` deps in here?

robertoaloi: Forget about my comment. I saw now the `els_text_edit` module.

include/erlang_ls.hrl
alanz: Agree, and leave the comment.

src/els_methods.erl
alanz: I re-ordered the capabilities to line up with the order in the spec. It is easier to see what is needed, and what is extra that way.  The re-ordering could arguably be done in a separate PR.

src/els_text_edit.erl
```diff
@@ -0,0 +1,71 @@
...
+
+-type diff() :: {del, [string()]} | {eq, [string()]} | {ins, [string()]}.
+
+-spec make_text_edits([diff()]) -> [text_edit()].
+make_text_edits(Diffs) ->
+    make_text_edits(Diffs, 0, []).
+
+-spec make_text_edits([diff()], number(), [text_edit()]) -> [text_edit()].
+make_text_edits([{eq, Data}|T], Line, Acc) ->
+    make_text_edits(T, Line + length(Data), Acc);
```
michalmuskala: Please make sure this works correctly with utf8 data. `lenth` on Erlang strings returns the number of code-points, which might or might not be what is expected here (vs byte size or grapheme count or utf-16 codepoint size).

src/els_text_edit.erl
```diff
@@ -0,0 +1,71 @@
...
+
+make_text_edits([{del, Del}, {ins, Ins}|T], Line, Acc) ->
+    Len = length(Del),
+    Pos1 = #{ line => Line,       character => 0 },
+    Pos2 = #{ line => Line + Len, character => 0 },
+    Edit = #{ range => #{ start => Pos1, 'end' => Pos2 }
+            , newText => list_to_binary(lists:concat(Ins))
```
michalmuskala: This might now work correctly with utf8 data. Is `Ins` a list of bytes or code-points? `list_to_binary` generally fails on elements `> 255`. Probably `unicode:characters_to_binary` would be better in here, and then `lists:concat` is not needed.

src/els_text_edit.erl
alanz: At this point the `Data` is a list of lines,  as we are operating at the whole line level.  We use the length of the list to advance the line counter.  So unicode encoding should not be a problem.

src/els_text_edit.erl
alanz: `lists:concat` is used to join up the lines again. Agree on using `unicode:characters_to_binary`.

rebar.config
alanz: I got it working with the unpatched version, so this is not needed.

src/els_client.erl
```diff
@@ -272,6 +296,13 @@ request_params({initialize, {RootUri, InitOptions}}) ->
    , <<"initializationOptions">> => InitOptions
    , <<"capabilities">> => #{ <<"textDocument">> => TextDocument }
    };
+request_params({ document_formatting
+               , {Uri, TabSize, InsertSpaces}}) ->
+  #{ textDocument => #{ uri => Uri }
+   , options      => #{ tabSize      => TabSize
+                      , insertSpaces => InsertSpaces
+                      }
```
alanz: Should this use the explicit params like this, or use a single parameter for `Options`, which is expected to be

```erlang
#{ tabSize      => TabSize
 , insertSpaces => InsertSpace
 }
```

include/erlang_ls.hrl
```diff
@@ -545,6 +545,21 @@
                                   , active_parameters => number()
                                   }.
 
+%%------------------------------------------------------------------------------
+%% Formatting
+%%------------------------------------------------------------------------------
+-type formatting_options() :: #{ tabSize           := integer()
```
robertoaloi: I'm trying this branch locally. Even if my client sends `2` as the `tabSize`, I get 4 spaces indentation...

src/els_formatting_provider.erl
```diff
@@ -0,0 +1,112 @@
...
+
+-spec format_document(uri(), els_document:document(), formatting_options())
+                     -> {ok, [text_edit()]}.
+format_document(Uri, _Document, Options) ->
+    lager:info("format_document: ~p", [{Uri, Options}]),
+    Path = els_uri:path(Uri),
+    Fun = fun(Dir) ->
+            RelPath = els_utils:project_relative(Uri),
+            OutFile = filename:join(Dir, RelPath),
+            Opts = #{output_dir => Dir},
```
robertoaloi: Shouldn't we pass to the formatter options such as `break_indent` and `sub_indent` calculated according to the client preferences?

robertoaloi:
@alanz What is `erlfmt` that you mention in your PR? Googling for it I could only find a 3-years old project and a tiny wrapper on top of `io_lib_pretty`.

src/els_formatting_provider.erl
alanz: I guess we could. I saw this as initial support, to be used as a basis for the coming `erlfmt` from @michalmuskala.  So it is bare bones.

I need to rebase for the new persistent db, will see about adding it at the same time.

alanz:
@robertoaloi `erlfmt` is a new formatter, which is being finalised at the moment, and wending its way through legal prior to being open sourced. Needs one more rubber stamp, I understand.

jfacorro:
@alanz Is that the one @michalmuskala mentioned he was working on?

src/els_client.erl
robertoaloi: I think this is fine for now.

src/els_formatting_provider.erl
robertoaloi: Cool. In Emacs formatting has to be triggered via a specific command. What about VSCode? If the formatting is always requested explicitly by the client, I'm fine with merging this (after the tab/space options are passed to the formatter). If they are triggered by things like `TAB` or happen automatically in VSCode I would lean towards merging it but disabling it by default (until the formatter is good enough).

src/els_methods.erl
robertoaloi: It's fine. I was just curious, since I couldn't see the reason.

test/els_formatter_SUITE.erl
```diff
@@ -0,0 +1,85 @@
...
+               range =>
+                   #{'end' => #{character => 0,line => 7},
+                     start => #{character => 0,line => 4}}},
+             #{newText => <<"main(X) -> X.\n">>,
+               range =>
+                   #{'end' => #{character => 0,line => 8},
+                     start => #{character => 0,line => 8}}}]
+             , Result)
+  catch _Err ->
+      file:set_cwd(Cwd)
```
robertoaloi: I assume you want to do this in all cases, no just when an exception is caught (i.e. in an `after` block).

test/els_formatter_SUITE.erl
```diff
@@ -0,0 +1,85 @@
...
+-spec end_per_testcase(atom(), config()) -> ok.
+end_per_testcase(TestCase, Config) ->
+  els_test_utils:end_per_testcase(TestCase, Config).
+
+%%==============================================================================
+%% Testcases
+%%==============================================================================
+-spec format_doc(config()) -> ok.
+format_doc(Config) ->
+  Cwd = file:get_cwd(),
```
robertoaloi: Why do we need to set the `cwd`?

alanz: Because the formatting library requires a relative path.  And it is reset at the end to make sure the surrounding context expectations don't go bad if the test fails.

test/els_formatter_SUITE.erl
alanz: true. Oops.

robertoaloi:
Nice one, thanks!

PR #303, opened 2019-12-05
[#302] Solve flakyness of els_document_highlight_SUITE

### LINKED ISSUE #302: els_document_highlight_SUITE is flaky
On Travis, the suite occasionally fails. Investigate why.

robertoaloi:
### Description

This is one way to solve things. Please advise if you think we should do something else.

Fixes #302 .

### DIFF
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -21,7 +21,7 @@
 -define(SERVER, ?MODULE).
 -define(TABLES, [ {documents,  set}
                 , {modules,    set}
-                , {references, duplicate_bag}
+                , {references, bag}
                 , {signatures, set}
                 ]).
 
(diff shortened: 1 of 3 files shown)


src/els_db.erl
```diff
@@ -63,7 +63,7 @@ list(Table) ->
 
 -spec store(table(), key(), any()) -> ok.
 store(Table, Key, Value) ->
-  true = ets:insert(Table, {Key, Value}),
+  ets:insert(Table, {Key, Value}),
```
jfacorro: Was this  pattern matching removed because with a `bag` table it's no longer always `true`?

alanz:
For reference [here](https://github.com/erlang-ls/erlang_ls/pull/251#discussion_r348462994) is the original discussion around the change which resulted in `duplicate_bag`.

alanz:
And looking at the spec, I see it says


> bag
> The table is a bag table, which can have many objects, but only one instance of each object, per key.
>
> duplicate_bag
> The table is a duplicate_bag table, which can have many objects, including multiple copies of the same object, per key.

And I presume `object` is the value stored, so `bag` is in fact the right choice.

src/els_db.erl
robertoaloi: That is correct.

PR #304, opened 2019-12-07
[#151] Persistent DB

### LINKED ISSUE #151: A persistent DB
Indexing is an expensive operation. We probably want to persist the indexed data on disk across sessions.

robertoaloi:
### Description

A port of the current solution to Mnesia. Sorry about the huge size, but it wasn't something easy to do in small chunks.
With this PR, a database instance is created per project/otp version.
The aim of this PR is to gain persistency whilst not introducing any regression.

Closes #151 .


rebar.config
```diff
@@ -16,10 +16,12 @@
        , {docsh,       "0.7.2"}
        , {worker_pool, "4.0.1"}
        , {elvis,       "0.4.3", {pkg, elvis_core}}
+       , {uuid,        "1.8.0", {pkg, uuid_erl}}
```
robertoaloi: This has been included to produce unique ids, not used with the current schema.

rebar.config
```diff
@@ -90,3 +92,5 @@
 %% Disable warning_as_errors for redbug to avoid deprecation warnings.
 {overrides, [ {del, redbug, [{erl_opts, [warnings_as_errors]}]}
             ]}.
+
+{dist_node, [ {sname, 'erlang_ls'} ]}.
```
robertoaloi: Most rebar3 commands (e.g. `ct` will use this to start the node in distributed mode).

src/els_completion_provider.erl
```diff
@@ -258,6 +260,7 @@ keywords() ->
 %% Filter by prefix
 %%==============================================================================
 
+%% TODO: Implement as select
```
robertoaloi: Most of the filtering/selecting in the providers can be implemented as Mnesia queries.

src/els_db.erl
```diff
@@ -1,133 +1,93 @@
...
+  create_tables(),
+  application:stop(mnesia).
+
+-spec create_tables() -> ok.
+create_tables() ->
+  [els_db_table:create(T) || T <- ?TABLES],
   ok.
 
--spec update(table(), key(), function(), any()) -> ok.
-update(Table, Key, UpdateFun, Default) ->
```
robertoaloi: It turns out this function was unused?

src/els_db.erl
```diff
@@ -1,133 +1,93 @@
...
-  end.
-
--spec delete(table(), key()) -> ok.
+-spec delete(atom(), any()) -> ok | {error, any()}.
 delete(Table, Key) ->
-  true = ets:delete(Table, Key),
+  F = fun() -> mnesia:delete({Table, Key}) end,
+  transaction(F).
+
+-spec lookup(atom(), any()) -> {ok, [tuple()]} | {error, any()}.
```
robertoaloi: We probably don't need to run everything as a transaction.

src/els_document.erl
```diff
@@ -11,6 +11,7 @@
 -export([ create/2
         , uri/1
         , text/1
+        , md5/1
```
robertoaloi: Used to check whether the version of the document is already indexed.

src/els_utils.erl
```diff
@@ -41,13 +41,14 @@ find_module(M, Extension) ->
...
   {ok, els_document:document()} | {error, any()}.
 find_document(Uri) ->
-  case els_db:find(documents, Uri) of
-    {ok, Document} ->
+  case els_dt_documents:lookup(Uri) of
+    {ok, [#{document := Document}]} ->
       {ok, Document};
-    {error, not_found} ->
+    {ok, []} ->
+      %% TODO: Maybe we should not index this
```
robertoaloi: I am often getting timeouts when using this branch and navigating towards a OTP module during indexing.

src/erlang_ls.erl
```diff
@@ -12,6 +12,7 @@ main(Args) ->
   ok = parse_args(Args),
   ok = init_node_name(is_debug()),
   ok = lager_config(),
+  els_db:install(),
```
robertoaloi: The DB is currently recreated on startup. Obviously this needs to be fixed.

src/erlang_ls.app.src
```diff
@@ -29,6 +30,7 @@
         %% Logging
       , {logging_enabled, false}
       , {log_dir, "/usr/local/var/log/erlang_ls"}
+      , {db_dir, "/usr/local/var/cache/erlang_ls"}
```
alanz: Will this work on different operating systems?

robertoaloi: Fair point. The logging would also break, but that's disabled by default.
On the other hand, if we aim at supporting multiple platforms we should somehow add them to the CI pipelines.

rebar.config
jfacorro: Do we need this though?

src/els_db.erl
jfacorro: We stopped using it when we changes the `references` table to a `bag`.

src/els_dt_documents.erl
```diff
@@ -0,0 +1,83 @@
...
+%%==============================================================================
+%% Item Definition
+%%==============================================================================
+
+-record(els_dt_documents, { uri      :: uri()  | '_'
+                          , document :: els_document:document()
+                          }).
+-type els_dt_documents() :: #els_dt_documents{}.
+
+-type item() :: #{ uri      := uri()
```
jfacorro: The document already contains the `uri`, why not just have the following?
```
-type item() :: els_document:document().
```

src/els_dt_documents.erl
```diff
@@ -0,0 +1,83 @@
...
+%%==============================================================================
+%% API
+%%==============================================================================
+
+-spec from_map(item()) -> els_dt_documents().
+from_map(#{ uri := Uri, document := Document }) ->
+  #els_dt_documents{ uri = Uri, document = Document }.
+
+-spec to_map(els_dt_documents()) -> item().
+to_map(#els_dt_documents{ uri = Uri, document = Document }) ->
```
jfacorro: This could be renamed to `to_item` if the suggestion above is applied.

src/els_dt_documents.erl
```diff
@@ -0,0 +1,83 @@
+%%==============================================================================
+%% The 'documents' table
+%%==============================================================================
+
+-module(els_dt_documents).
+
```
jfacorro: Isn't the `-behaviour(els_db_table).` missing from all `*_dt_*` modules?

src/els_document_highlight_provider.erl
```diff
@@ -52,21 +52,22 @@ find_highlights(Uri, #{ kind := Kind
...
+          ],
       R
   end;
 find_highlights(_Uri, _POI) ->
   null.
 
 -spec document_highlight(poi_range()) -> map().
 document_highlight(Range) ->
-    %% TODO:AZ:range as displayed ends one col too early
+  %% TODO:AZ:range as displayed ends one col too early
```
jfacorro: This was probably fixed already by [this change](https://github.com/erlang-ls/erlang_ls/pull/280/files#diff-3f144025dae38c6a860fe792f9f86294R26).

src/els_dt_documents.erl
```diff
@@ -0,0 +1,83 @@
+%%==============================================================================
+%% The 'documents' table
+%%==============================================================================
+
+-module(els_dt_documents).
```
jfacorro: What does `dt` stand for here?

src/erlang_ls.erl
jfacorro: What happens in debug mode when the node's name is always different? Have you tried this branch out with different `erlang_ls` instances working at the same time?

src/erlang_ls.app.src
michalmuskala: [`filename:basedir/3`](http://erlang.org/doc/man/filename.html#basedir-3) might be useful to get correct temp or config directories on various systems

src/els_dt_documents.erl
robertoaloi: That's exactly where I'm heading to in the follow-up commits (not pushed yet).

src/els_code_navigation.erl
```diff
@@ -51,9 +51,8 @@ goto_definition(Uri, #{ kind := record_expr, id := Record }) ->
 goto_definition(_Uri, #{ kind := Kind, id := Include }
                ) when Kind =:= include;
                       Kind =:= include_lib ->
-  FileName = filename:basename(Include),
-  M = list_to_atom(FileName),
-  case els_utils:find_module(M) of
+  H = list_to_atom(filename:basename(Include, filename:extension(Include))),
```
robertoaloi: The "module" table is now gone, in favour of an indexed attribute on the "document" table. Now, header files are stored as "documents" as well and their id does not include the 'hrl' extension anymore, hence this change.

src/els_code_navigation.erl
```diff
@@ -51,9 +51,8 @@ goto_definition(Uri, #{ kind := record_expr, id := Record }) ->
...
+  case els_utils:find_header(H) of
```
robertoaloi: This is something that annoyed me for a long time. We were looking for modules named 'xxx.hrl'.

src/els_code_navigation.erl
```diff
@@ -72,38 +71,34 @@ goto_definition(_Filename, _) ->
 find([], _Kind, _Data) ->
   {error, not_found};
 find([Uri|Uris0], Kind, Data) ->
-  case els_db:find(documents, Uri) of
-    {ok, Document} ->
-      POIs = els_document:points_of_interest(Document, [Kind], Data),
-      case POIs of
+  case els_dt_document:lookup(Uri) of
```
robertoaloi: To support both sets and bags, the lookup API currently returns a list. We may want to change this later.

src/els_document_highlight_provider.erl
robertoaloi: Removed.

src/erlang_ls.app.src
robertoaloi: Nice! I didn't know about it.

robertoaloi: That's what happens when you tend to only support one system :p

src/erlang_ls.erl
robertoaloi: Things will probably explode. That's one of the reasons why this branch is not ready for merge :)

test/els_completion_SUITE.erl
```diff
@@ -88,10 +88,6 @@ atom_completions(Config) ->
                 , kind             => ?COMPLETION_ITEM_KIND_MODULE
                 , label            => <<"code_navigation_types">>
                 }
-             , #{ insertTextFormat => ?INSERT_TEXT_FORMAT_PLAIN_TEXT
```
robertoaloi: I think this was an actual bug. We were including header files as completions for atoms. This was due to the fact we were storing modules and headers together. With the new implementation this is not the case anymore.

test/els_completion_SUITE.erl
```diff
@@ -270,11 +266,11 @@ macros(Config) ->
 
   #{result := Completion1} =
     els_client:completion(Uri, 24, 1, TriggerKindChar, <<"?">>),
-  ?assertEqual(lists:sort(Expected), lists:sort(Completion1)),
+  [?assert(lists:member(E, Completion1)) || E <- Expected],
```
robertoaloi: Another spotted bug: for some reason, macros from included files were not returned in the completion list. This is now fixed.

rebar.config
robertoaloi: Ended up removing this, since not in scope.

rebar.config
robertoaloi: I don't recall why I had to add it. It was related to the Mnesia setup. Will check...

src/els_dt_documents.erl
robertoaloi: I ended up merging the `els_document` and `els_dt_document` entities into one. I also included the `kind` in the document and got rid of the `modules` table.

src/els_dt_documents.erl
robertoaloi: Sounds reasonable. Will do.

src/els_dt_documents.erl
robertoaloi: 🤦‍♂

src/els_dt_documents.erl
robertoaloi: Db Table. It's the name of the behaviour. I'm open for suggestion.

src/els_utils.erl
```diff
@@ -1,12 +1,11 @@
 -module(els_utils).
 
--export([ find_document/1
+-export([ find_header/1
         , find_module/1
-        , find_module/2
         , fold_files/4
+        , lookup_document/1
```
jfacorro: Why the difference in naming for this function? Why not `find_document/1`?

src/els_indexer.erl
```diff
@@ -67,20 +64,32 @@ index_file(Path, SyncAsync) ->
...
+%% TODO: Avoid nested transactions
+-spec index(els_dt_document:item()) -> ok.
+index(#{uri := Uri, text := Text} = Document) ->
+  MD5 = erlang:md5(Text),
+  case els_dt_document:lookup(Uri) of
+    {ok, [#{md5 := MD5}]} ->
+      %% The module is already indexed, no action needed.
+      ok;
+    _ ->
+      F = fun() ->
```
jfacorro: For readability, would it be possible to move this out into it's own function that takes arguments?

src/els_dt_documents.erl
jfacorro: I don't have any better suggestions to be honest 😜

src/els_db.erl
robertoaloi: Already done in subsequent commits.

src/els_utils.erl
robertoaloi: This was due to a bug in the db module. Fixed.

src/els_code_navigation.erl
robertoaloi: Fixed.

src/els_indexer.erl
robertoaloi: Yes, but I would leave this refactoring out of this PR. The indexer module needs some love.

rebar.config
```diff
@@ -78,6 +78,7 @@
 {dialyzer, [ {warnings, [ unknown
                         ]}
            , {plt_apps, all_deps}
+           , {plt_extra_apps, [mnesia]}
```
robertoaloi: Due to the Mnesia setup (the application needs to be stopped when creating the schema), mnesia is not part of the erlang_ls `apps`. Therefore, Dialyzer would not include it as a dependency automatically, resulting in "unknown functions" warnings.

src/els_code_navigation.erl
```diff
@@ -106,4 +101,6 @@ add_include_uri(#{ id := String }, Acc) ->
 beginning() ->
   #{range => #{from => {1, 1}, to => {1, 1}}}.
 
-%% TODO: Handle multiple header files with the same name?
+-spec filename_to_atom(string()) -> atom().
+filename_to_atom(FileName) ->
+  list_to_atom(filename:basename(FileName, filename:extension(FileName))).
```
robertoaloi: Yes, dinamically creating atoms. But we can't always ensure they always exist (and therefore use the `to_existing_atom` variant here. Hopefully, this is not going to be a problem...

src/els_utils.erl
robertoaloi: The idea was to distinguish between a lookup and a search operation.

robertoaloi:
Merged because:

* I got an approval \o/
* it will be easier for us to test it on a daily basis
* the PR was huge and would conflict with almost every other PR
* CI did not show any regressions
* we can still incrementally fix/clean it
* The VSCode extension (which seems to be the most used way to interact with the server) still uses an older version

Feel free to add after-merge comments and I will address them.

PR #308, opened 2019-12-12
[#274] Support multiple apps per project

### LINKED ISSUE #274: Rebar3 release project are not properly indexed
**Describe the bug**
Creating a rebar3 release project (rebar3 new release) does not get indexed. On this king of project, the server do not seems to be able to manage a simple find go to.

**To Reproduce**
1. Install lsp-mode for emacs with spacemacs
2. Create a release project with `rebar3 new release my_release`
3. Open the rebar.config and ensure `erlang-mode` is enable. This will enable the the lsp as well.
4. Open the file `apps/my_release/my_release_app.erl`
5. Try to `,gg` to your root supervisor.
6. You should get a no match found.
7. Visit the supervisor file and return to the app file.
8. Now the goto will work

**Expected behavior**
The goto should work when visiting the app file the first time.

**Actual behavior**
The file in a release project does not get indexed.

**Context**
 - `erlang_ls` version (tag/sha): daa347c6578825d691d185edceb951dc999a7b96
 - Editor used: emacs/spacemacs (dev, 8fcc9d849044dca6e69ff3f7bda89f82abf848b7)
 - LSP client used: lsp-mode

robertoaloi:
### Description

The language server can now support multiple applications for the same project (aka umbrella-style projects). A rebar3 release (as created by `rebar3 new release rebar3_release`) has been added to the `priv` directory, so the structure can be properly tested. A first testcase covering code navigation has been added to the suite. The PR also includes some refactoring to the `els_test_utils` module. That module can be further refactored, but that will be done in a separate PR.

Fixes #274 .


priv/rebar3_release/.gitignore
```diff
@@ -0,0 +1,19 @@
+.rebar3
```
robertoaloi: All these files are included verbatim from a newly rebar3 release created from scratch.

priv/rebar3_release/erlang_ls.config
```diff
@@ -0,0 +1,2 @@
+apps_dirs:
```
robertoaloi: This is all it's needed to support umbrella projects after this PR.

src/els_methods.erl
```diff
@@ -94,10 +94,11 @@ initialize(Params, State) ->
    } = Params,
   InitOptions = maps:get(<<"initializationOptions">>, Params, #{}),
   ok = els_config:initialize(RootUri, Capabilities, InitOptions),
+  DbDir = application:get_env(erlang_ls, db_dir, default_db_dir()),
```
robertoaloi: This is to avoid test suites polluting the cache directory. The DB will be created in the local test dir for CT suites.

test/els_test_utils.erl
```diff
@@ -147,6 +132,23 @@ end_per_testcase(_TestCase, Config) ->
   [application:stop(App) || App <- ?config(started, Config)],
   ok.
 
+-spec start(stdio | tcp) -> ok.
```
robertoaloi: Refactored so it can be re-used by other test suites.

priv/rebar3_release/erlang_ls.config
michalmuskala: How could this be used if the app dirs are directly in the project root?

jfacorro: I think the following should work, this is the deafult value though (see [here](https://github.com/erlang-ls/erlang_ls/pull/308/files#diff-a40e5ac4b7520587b5f49f27230ed126R72)):
```
apps_dirs:
  - "."
```

README.md
```diff
@@ -264,6 +264,7 @@ Currently, the following customizations are possible:
 |--------------------|-----------------------------------------------------------------------------------------|
 | otp\_path          | Path to the OTP installation                                                            |
 | deps\_dirs         | List of directories containing dependencies. It supports wildcards.                     |
+| apps\_dirs         | List of directories containing project applications. It supports wildcards.             |
```
jfacorro: Should we add what the default value is (i.e. `["."]`)?

priv/rebar3_release/erlang_ls.config
robertoaloi: Yes, what @jfacorro said :)

README.md
robertoaloi: I thought about it, but we don't do it for any other case. Let's improve the config docs separately from this PR.

Issue #309, opened 2019-12-12
DB initialization takes up to 10s


robertoaloi:
After we introduced Mnesia, DB initialization started to take a long time. While a time increase is expected, 10 seconds seem excessive. Check what's happening.

robertoaloi:
The delay comes from the `wait_for_tables` operation. Tables load in the following order: signatures, documents, references.

robertoaloi:
This is probably due to Mnesia repairing tables after a non-clean shutdown. This is superseded by #337 .

jfacorro:
I tried both #337 and #338 and neither separately or combined made a difference. I  think it would  make more sense to close those other two and add as tasks the things mentioned in #337 and #338.

PR #311, opened 2019-12-12
[#309] Increase DB wait_for_tables timeout to 20s until the root cause is identified

robertoaloi:
### Description

Avoid timeouts while #309 is properly investigated.

### DIFF
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -16,7 +16,7 @@
                 , els_dt_references
                 , els_dt_signatures
                 ]).
--define(TIMEOUT, 5000).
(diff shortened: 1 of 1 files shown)


src/els_db.erl
```diff
@@ -16,7 +16,7 @@
                 , els_dt_references
                 , els_dt_signatures
                 ]).
--define(TIMEOUT, 5000).
```
jfacorro: This might need to be even bigger. I bumped it to 20 seconds locally since I was still getting timeouts.

robertoaloi: Bumped to 20s for now.

Issue #108, opened 2019-10-20
Code Lens for including a header file when auto-completing an imported record name.


robertoaloi:
When pressing the '#' character, record names will be suggested for auto-complete (see #92 ). Whenever a record defined in a non-included header file is picked, a code lens should propose the user to include the header file.

alanz:
I think this is probably better done with a code action.

alanz:
The original intent of this issue can be achieved by returning a value for the `additionalTextEdits?: TextEdit[];` field of the returned `CompletionItem`.

plux:
Added a code action to accomplish this:
https://github.com/erlang-ls/erlang_ls/pull/1510

PR #314, opened 2019-12-12
[#108] Add code action support

alanz:
### Description

Add support for code actions, code lenses, and execute command.

This allows generation of quick fixes for problems.

Includes quick fix for unused variables, by prepending a `_`.

Provides infrastructure for # 108.


include/erlang_ls.hrl
```diff
@@ -149,10 +149,15 @@
 %%------------------------------------------------------------------------------
 %% Workspace Edit
 %%------------------------------------------------------------------------------
--type workspace_edit() :: #{ changes         => #{ binary() := [text_edit()]
-                                                 }
-                           , documentChanges => [text_document_edit()]
-                           }.
+
+-type document_change() :: [text_document_edit()].
+                         %% | [ text_document_edit()
```
robertoaloi: Please remove the commented text and create an issue so we can follow up.

include/erlang_ls.hrl
```diff
@@ -560,6 +562,29 @@
         , more_trigger_character  => string()
         }.
 
+%%------------------------------------------------------------------------------
+%% Code Actions
+%%------------------------------------------------------------------------------
+
+-define(CODE_ACTION_KIND_QUICKFIX, 1).
```
robertoaloi: Where is the `1` specified in the specs? I can't find it...

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,128 @@
...
+  lager:info("code_actions: [Context=~p]", [Context]),
+  #{ <<"start">> := #{ <<"character">> := _StartCol
+                     , <<"line">>      := StartLine }
+   , <<"end">> := _End
+   } = Range,
+  #{ <<"diagnostics">> := Diagnostics } = Context,
+  %% TODO: process optional "only" field in Context
+  Actions0 = [ make_code_action(Uri, C) || C <- Diagnostics],
+
+  Actions = case application:get_env(erlang_ls, test_code_action) of
```
robertoaloi: I would remove this from the current version.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,128 @@
...
+                                    , before => Before }])
+    }.
+
+-spec make_code_action(uri(), diagnostic()) -> [code_action()].
+make_code_action(Uri, #{ <<"message">> := Message
+                       , <<"range">> := Range } = _Diagnostic) ->
+    lager:info("make_code_action: [Message=~p]", [Message]),
+    %% processing messages of the type "type type_name() undefined".
+    %% Good resource:
+%% https://arifishaq.wordpress.com/2013/11/13/playing-with-regular-expressions/
```
robertoaloi: Nit: indentation

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,128 @@
...
+          els_utils:make_command( Title
+                                , <<"add-lines">>
+                                , [#{ uri => Uri
+                                    , lines => Lines
+                                    , before => Before }])
+    }.
+
+-spec make_code_action(uri(), diagnostic()) -> [code_action()].
+make_code_action(Uri, #{ <<"message">> := Message
+                       , <<"range">> := Range } = _Diagnostic) ->
```
robertoaloi: Using a regexp on the diagnostic message seems a fragile approach, since messages could vary between tools versions.
Diagnostics have a 'code' field which identify the diagnostic type. Maybe we could leverage that to have an exact match?

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,128 @@
...
+          end,
+    lager:info("undefined_type_action: [MFA=~p]", [MFA]),
+
+    Res = els_code_navigation:goto_definition(Uri
+                                             , #{ kind => type_application
+                                                , id => MFA
+                                                , data => undefined
+                                                , range => #{ from => {0, 0}
+                                                            , to   => {0, 0}}
+                                                }),
```
robertoaloi: This should use the `els_poi:new/X` API instead.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,128 @@
...
+    lager:info("undefined_type_action: [Res=~p]", [Res]),
+    [].
+
+-spec to_binary(any()) -> binary().
+to_binary({atom, 1, Atom}) -> Atom.
+     %% erlang:atom_to_binary(Atom, unicode).
```
robertoaloi: Should we use `erl_syntax:atom_value/1` instead?

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,128 @@
...
+
+-spec undefined_type_action(uri(), range(), binary()) -> [code_action()].
+undefined_type_action(Uri, _Range, UndefinedType) ->
+    lager:info("undefined_type_action: [UndefinedType=~p]", [UndefinedType]),
+    R = els_parser:parse(UndefinedType),
+    lager:info("undefined_type_action: [R=~p]", [R]),
+    {ok, Tokens, _} = erl_scan:string(binary_to_list(UndefinedType) ++ "."),
+    {ok, [{call, _, Name, Args}]} = erl_parse:parse_exprs(Tokens),
+    lager:info("undefined_type_action: [{Name,Args}=~p]", [{Name, Args}]),
+    MFA = case Name of
```
robertoaloi: We should probably move this to its own function and, if possible, use the `erl_syntax` API.

include/erlang_ls.hrl
alanz: I think it must have changed in the spec over time, the original value came in two years ago, via https://github.com/erlang-ls/erlang_ls/commit/525229bc70ddcf33d40c7631a460cabb3b2f370b

include/erlang_ls.hrl
alanz: https://github.com/erlang-ls/erlang_ls/issues/365

src/els_code_action_provider.erl
alanz: If I indent, elvis complains. Ideas?

src/els_code_action_provider.erl
alanz: I added it as a template to other devs for how to use this, and to ensure it works via tests.

Perhaps we should remove it as soon as we put in a real code action? If so, I will create an issue.

src/els_code_action_provider.erl
alanz: I understand the `code` field to be used as a reference back to the compiler spec in terms of the kind of error found, which means it would be a stretch to pack this info into it.

The fundamental problem is that we are working with what the compiler emits as a warning or error.  So we have to parse the error message some time, either when we generate the diagnostic, or when we generate the code action. The best approach is to modify the compiler to provide a structured diagnostic that we can use directly.

cc @michalmuskala  @ilya-klyuchnikov

src/els_code_action_provider.erl
alanz: I will move it to its own function, but I gave `erl_syntax` a quick scan and it is not clear to me what I should use.

src/els_code_action_provider.erl
michalmuskala: Sorry, I didn't see this comment before.

In general, it's possible to get structured output from the Erlang compiler. Calling the `compile:file/2` function with the `return` option, causes it to produce something like:
```erl
{error, 
	[{"test.erl", [{3,erl_lint,{undefined_function,{foo,1}}}]}],
	[]}
```
The text-based message might be recovered by calling, in this example, `erl_lint:format_error({undefined_function,{foo,1}})`. As far as I understand, this structured message should be treated as opaque and there are no guarantees about its stability, but in practice it seems to be stable enough. The Elixir compiler was matching on messages like this to emit proper error messages related to the original Elixir source before rewriting the linter pass and still matches on some of the messages coming from [deeper compiler passes](https://github.com/elixir-lang/elixir/blob/033c7387e25aa31fb4d1bf52d2b5cef6519e9df7/lib/elixir/src/elixir_erl_compiler.erl#L87-L150).

robertoaloi: @michalmuskala @alanz Compiler diagnostics use exactly this mechanism. See https://github.com/erlang-ls/erlang_ls/blob/634cc2ad63f8b1c5675737ea9b62764fe912630c/src/els_compiler_diagnostics.erl#L74 and https://github.com/erlang-ls/erlang_ls/blob/634cc2ad63f8b1c5675737ea9b62764fe912630c/src/els_compiler_diagnostics.erl#L137 (since the callback module can be something other than `erl_lint`).

alanz: Ok, so we are back to the problem of how to associate the diagnostic with the original compiler warning/error.  There is no "additional info" field in the `Diagnostic` type, and I think using the `code` field is inappropriate.

```typescript
/**
  * The diagnostic's code, which might appear in the user interface.
  */
	code?: number | string;
```

Perhaps we should store it, against a key made up of the diagnostic itself, or some key subset of its fields?

alanz: FYI: https://github.com/microsoft/language-server-protocol/issues/887

elvis.config
```diff
@@ -4,7 +4,7 @@
       [ #{ dirs    => ["src"]
          , filter  => "*.erl"
          , ruleset => erl_files
-         , rules   => [ {elvis_style, god_modules, #{ignore => [els_client]}}
+         , rules   => [ {elvis_style, god_modules, #{ignore => [els_client,els_methods]}}
```
robertoaloi: There's no space after the comma. Should we run elvis on the elvis.config file? :D

src/els_code_action_provider.erl
robertoaloi: Even if we find a way to reference an existing diagnostic, this would imply we need to store diagnostics on the server side, which would require many changes.
In 3.15 they added some (very limited) "tags" to the Diagnostic. That should probably be extended to include generic metadata. I agree `code` does not sound correct, but it feels like it's the only thing we have right now. Is there any documentation explaining the idea behind the `code` field? It's still not clear to me what it's supposed to represent.

For the time being, considering we don't have any real code action, what do you think about splitting this PR into parts (one for the code lenses support, one for the execute command), so we can easily merge those and hold on the code action part?

alanz: We don't have a code lens use case now either.

Perhaps I should split it into adding the machinery without the test, and keeping the test with its config in a separate branch as a model for when we do actually implement one.

I am not all that happy to have a feature in place that is noc "claimed" by a test though, the danger exists that it can rot.

src/els_code_action_provider.erl
alanz: Removed it.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,183 @@
...
+-include("erlang_ls.hrl").
+
+%%==============================================================================
+%% els_provider functions
+%%==============================================================================
+
+-spec is_enabled() -> boolean().
+is_enabled() -> true.
+
+%% /**
```
robertoaloi: Please use `@edoc` syntax for docs.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,183 @@
...
+   , <<"context">> := Context } = Params,
+  Result = code_actions(Uri, RangeLSP, Context),
+  {Result, State}.
+
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+
+
+%% result: (Command | CodeAction)[] | null where CodeAction is defined
```
robertoaloi: Same here. Please use `@edoc`. The specs should be used for documenting the types.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,183 @@
...
+%% as follows:
+-spec code_actions(uri(), range(), code_action_context()) -> [map()].
+code_actions(Uri, _Range, Context) ->
+  lager:info("code_actions: [Context=~p]", [Context]),
+  #{ <<"diagnostics">> := Diagnostics } = Context,
+  Actions0 = [ make_code_action(Uri, C) || C <- Diagnostics],
```
robertoaloi: Why `C`? Each entry is a diagnostic, so the name should reflect that.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,183 @@
...
+%%  * The server provides code actions. The `CodeActionOptions` return type is
+%%  * only valid if the client signals code action literal support via the
+%%  * property `textDocument.codeAction.codeActionLiteralSupport`.
+%%  */
+%% codeActionProvider?: boolean | CodeActionOptions;
+
+-spec options() -> boolean() | map().
+options() ->
+  Capabilities = els_config:get(capabilities),
+  lager:info("code_actions: [Capabilities=~p]", [Capabilities]),
```
robertoaloi: We are not using the capabilities, so I would remove this for now.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,183 @@
...
+  true.
+
+-spec handle_request(any(), els_provider:state()) ->
+  {any(), els_provider:state()}.
+handle_request({document_codeaction, Params}, State) ->
+  #{ <<"textDocument">> := #{ <<"uri">> := Uri}
```
robertoaloi: In maps, we tend to align key/value pairs for readability. As a side note, it would be incredibly cool if one could configure the formatter to do such a thing.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,183 @@
...
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+
+
+%% result: (Command | CodeAction)[] | null where CodeAction is defined
+%% as follows:
+-spec code_actions(uri(), range(), code_action_context()) -> [map()].
+code_actions(Uri, _Range, Context) ->
+  lager:info("code_actions: [Context=~p]", [Context]),
```
robertoaloi: Not sure if we need this level of debugging. Maybe drop them?

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,183 @@
...
+  #{ <<"diagnostics">> := Diagnostics } = Context,
+  Actions0 = [ make_code_action(Uri, C) || C <- Diagnostics],
+  Actions = lists:flatten(Actions0),
+  lager:info("code_actions: [Actions=~p]", [Actions]),
+  Actions.
+
+%% Note: if the start and end line of the range are the same, the line
```
robertoaloi: Please use `@edoc`, so the doc can be displayed by the server (e.g. on hover).

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,183 @@
...
+    lager:info("make_unused_variable_action: [UpdatedLine=~p]"
+              , [UpdatedLine]),
+    [ replace_lines_action( Uri
+                      , <<"Add '_' to '", UnusedVariable/binary, "'">>
+                      , list_to_binary(UpdatedLine)
+                      , Range)].
+
+%%------------------------------------------------------------------------------
+
+%% Based on https://stackoverflow.com/a/2185533
```
robertoaloi: `els_text` contains utilities to split binaries into lines. There we use `Lines = binary:split(Text, <<"\n">>, [global]),` for the split, which should be good enough. We can expose that function and call it from here.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,183 @@
...
+string2lines("\n" ++ Str, Acc) -> [lists:reverse(Acc) | string2lines(Str, [])];
+string2lines([H|T], Acc)       -> string2lines(T, [H|Acc]);
+string2lines([], Acc)          -> [lists:reverse(Acc)].
+
+%%------------------------------------------------------------------------------
+
+-spec undefined_type_action(uri(), range(), binary()) -> [code_action()].
+undefined_type_action(Uri, Range, Message) ->
+    %% processing messages of the type "type type_name() undefined".
+    %% TODO: this is a hot path, precompile the re's.
```
robertoaloi: We have a ticket about not having TODOs in the code. Let' s use GitHub issues for follow-up actions.

src/els_code_action_provider.erl
alanz: Note: the tag is `@doc`, not `@edoc`.  I am correcting this in the rest of  the app too.

src/els_code_action_provider.erl
alanz: Ok. Was useful when working on it.

src/els_code_action_provider.erl
alanz: cc @michalmuskala

alanz:
> @alanz I started reviewing this, but also realised that this PR includes too many things. I would prefer the providers to be included one at a time, that should simplify the review process a lot. What do you think?

I stripped out the code lens provider, it was just a stub anyway.

src/els_execute_command_provider.erl
```diff
@@ -0,0 +1,52 @@
...
+  lager:info("execute_command:replace-lines  [Arguments=~p]", [Arguments]),
+  Method = <<"workspace/applyEdit">>,
+  Params = #{ edit =>
+                  els_text_edit:edit_replace_text(Uri, Lines, LineFrom, LineTo)
+            },
+  lager:info("execute_command:replace-lines  [Params=~p]", [Params]),
+  els_server:send_request(Method, Params),
+  [];
+execute_command(Command, Arguments) ->
+  lager:info("execute_command: [Command=~p] [Arguments=~p]"
```
jfacorro: I think this message should be more in the lines of `"Unsupported command  [command=~p] [arguments=~p]"`.

src/els_execute_command_provider.erl
```diff
@@ -0,0 +1,52 @@
...
+               , [#{ <<"uri">>   := Uri
+                   , <<"lines">> := Lines
+                   , <<"from">>  := LineFrom
+                   , <<"to">>    := LineTo }] = Arguments) ->
+  lager:info("execute_command:replace-lines  [Arguments=~p]", [Arguments]),
+  Method = <<"workspace/applyEdit">>,
+  Params = #{ edit =>
+                  els_text_edit:edit_replace_text(Uri, Lines, LineFrom, LineTo)
+            },
+  lager:info("execute_command:replace-lines  [Params=~p]", [Params]),
```
jfacorro: I'm not sure if these log messages are necessary considering we have lowered the threshold and the logs now include `debug` as well.

src/els_execute_command_provider.erl
```diff
@@ -0,0 +1,52 @@
...
+%% Internal Functions
+%%==============================================================================
+
+-spec execute_command(string(), [any()]) -> [map()].
+execute_command(<<"replace-lines">>
+               , [#{ <<"uri">>   := Uri
+                   , <<"lines">> := Lines
+                   , <<"from">>  := LineFrom
+                   , <<"to">>    := LineTo }] = Arguments) ->
+  lager:info("execute_command:replace-lines  [Arguments=~p]", [Arguments]),
```
jfacorro: I'm not sure if these log messages are necessary considering we have lowered the threshold and the logs now include `debug` as well.

src/els_utils.erl
```diff
@@ -95,6 +96,13 @@ project_relative(Uri) ->
   <<RootUri:RootUriSize/binary, RelativePath/binary>> = Uri,
   binary_to_list(string:trim(RelativePath, leading, [$/, $\\ ])).
 
+-spec make_command(binary(), binary(), [any()]) -> command().
```
jfacorro: Shouldn't this be in `els_protocol`?

priv/code_navigation/src/code_navigation.erl
```diff
@@ -31,7 +31,7 @@ callback_a() ->
 function_c() ->
   code_navigation_extra:do(test),
   A = #record_a{},
-  _X = A#record_a.field_a,
+  XX = A#record_a.field_a,
```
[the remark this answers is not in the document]
alanz: Because existing tests are using that line, it makes the diff bigger.

robertoaloi:
@alanz The code action for unused variables does not seem to work as expected. I checked out this branch to test it locally and it looks like the underscore gets prepended to **all** variables in the ilne, which is incorrect. For example:

```
  case els_utils:find_module(M) of
    {ok, Uri}      -> X = find(Uri, function, {F, A});
    {error, Error} -> {error, Error}
  end;
```

This gets converted into:

```
    {ok, _Uri}      -> _X = find(_Uri, function, {_F, _A});
```

include/erlang_ls.hrl
```diff
@@ -149,10 +149,11 @@
...
 %% Workspace Edit
 %%------------------------------------------------------------------------------
--type workspace_edit() :: #{ changes         => #{ binary() := [text_edit()]
-                                                 }
-                           , documentChanges => [text_document_edit()]
-                           }.
+
+-type document_change() :: [text_document_edit()].
+
+-type workspace_edit() :: #{ changes         := map() }
```
robertoaloi: Looking at the protocol, both fields are marked as _optional_ (with `changes` being ignored if `documentChanges` is available), so `=>` should be used in the type definition.
The `changes` should be a struct of `uri() => [text_edit()]` pairs, so the original definition looks correct to me (where `binary()` could be replaced with a more accurate `uri()`.

robertoaloi:
@alanz What do we want to do with this PR? Do we want to bring in the boilerplate only for now? Do we want to decline and pause it for now? Complete it?

alanz:
> What do we want to do with this PR? Do we want to bring in the boilerplate only for now? Do we want to decline and pause it for now? Complete it?

I would like to complete it, as it now does at least one useful thing, to add an underscore to an unused variable.  I will only be able to get to it next week some time though.

It can be split into the boilerplate and the functionality, but my concern is the difficulty of putting in meaningful tests for the boilerplate stuff.

robertoaloi:
I just realized that my comments have been "pending" all of this time, since I never submitted the review (facepalm). Sorry about that. Ok, let's get in shape and merge it. I would not worry about testing the boilerplate for now.

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,114 @@
...
+is_enabled() -> true.
+
+%% @doc
+%%  The server provides code actions. The `CodeActionOptions' return type is
+%%  only valid if the client signals code action literal support via the
+%%  property `textDocument.codeAction.codeActionLiteralSupport'.
+%%
+%% `codeActionProvider?: boolean | CodeActionOptions;'
+
+-spec options() -> boolean() | map().
```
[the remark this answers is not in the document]
alanz: See #493

alanz:
> I would not worry about testing the boilerplate for now.

By having this as a single PR the boilerplate for code action and execute command does get tested, through being used for the code action included.  It is only not tested if the the PR is split into two.

Should it stay one, or split?

robertoaloi:
> Should it stay one, or split?

For this one I am ok to keep it one. As a side note, I suggest not force-pushing / rebasing after a review is open, since it makes it hard to follow incremental updates to the branch. Better to have commits on top of each other and eventually a final squash if the history is too messy.

src/els_execute_command_provider.erl
```diff
@@ -0,0 +1,50 @@
...
+%%==============================================================================
+%% els_provider functions
+%%==============================================================================
+
+-spec is_enabled() -> boolean().
+is_enabled() -> true.
+
+-spec options() -> map().
+options() ->
+  #{ commands => ["replace-lines"] }.
```
robertoaloi: This should be a binary, not a list.

src/els_execute_command_provider.erl
```diff
@@ -0,0 +1,50 @@
...
+  Arguments = maps:get(<<"arguments">>, Params, []),
+  Result = execute_command(Command, Arguments),
+  {Result, State}.
+
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+
+-spec execute_command(string(), [any()]) -> [map()].
+execute_command(<<"replace-lines">>
```
robertoaloi: @alanz I tried this locally, but I could not trigger it from Emacs. How should I test this?

src/els_code_action_provider.erl
```diff
@@ -0,0 +1,107 @@
...
+    case re:run(Message, "variable '(.*)' is unused"
+               , [{capture, all_but_first, binary}]) of
+        {match, [UnusedVariable]} ->
+            make_unused_variable_action(Uri, Range, UnusedVariable);
+        _ -> []
+    end.
+
+-spec make_unused_variable_action(uri(), range(), binary()) -> [map()].
+make_unused_variable_action(Uri, Range, UnusedVariable) ->
+    #{ <<"start">> := #{ <<"character">> := _StartCol
```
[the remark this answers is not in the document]
alanz: This was put together before the `.editorconfig` file was in place, it should be enforced/nudged via that now.

src/els_execute_command_provider.erl
alanz: It gets triggered via the code action. That is why this thing has to come in as a whole.   The process in general is

1.  Something generates diagnostics
2. The client requests code actions based on the current location in the file and the diagnostics that are active for that location.
3. Each code action provider examines the diagnostics (or other info) and generates a list of possible code actions. These can either be literal, where there is an embedded `ApplyEdit`, or contain a `command` and its parameters.
4. When the user selects a code action, if it has a literal edit, this is applied. Otherwise the `command` and it's parameters are send in an `executeCommand` request.
5. The response to the `executeCommand` is normally just empty, to indicate all went well.
6. The server then sends a `workspace/applyEdit` command to the client, with the changes derived from executing the command.

src/els_execute_command_provider.erl
alanz: It should be a list of binary

```typescript
export interface ExecuteCommandOptions extends WorkDoneProgressOptions {
	/**
	 * The commands to be executed on the server
	 */
	commands: string[]
}
```

alanz:
The remaining conflict is trivial

test/els_code_action_SUITE.erl
```diff
@@ -0,0 +1,90 @@
...
+  els_test_utils:end_per_testcase(TestCase, Config).
+
+%%==============================================================================
+%% Testcases
+%%==============================================================================
+-spec add_underscore_to_unused_var(config()) -> ok.
+add_underscore_to_unused_var(Config) ->
+  Uri = ?config(code_navigation_uri, Config),
+  Diag = #{<<"message">>  => <<"variable 'A' is unused">>
+          ,<<"range">>    =>
```
jfacorro: Nit (not for this PR): there should be a space after the comma in each comma-first line.

alanz: See #509

PR #321, opened 2019-12-13
[#312] Introduce module table for mapping module/header name to uri

### LINKED ISSUE #312: Listing all modules takes a lot of time
Getting all document where the `kind` field is `module`, takes a long time:
```
1> {Time, _} = timer:tc(mnesia, dirty_index_read, [els_dt_document, module, 4]), Time / 1000.
3259.71
```
The same is not true for the `header` kind:
```
2> f(), {Time, _} = timer:tc(mnesia, dirty_index_read, [els_dt_document, header, 4]), Time / 1000.
11.2
```

jfacorro:
### Description

Introduce module table for mapping module/header name to uri.

Fixes #312.

### DIFF
--- a/src/els_code_navigation.erl
+++ b/src/els_code_navigation.erl
@@ -48,10 +48,10 @@ goto_definition(Uri, #{ kind := record_access
   find(Uri, record, Record);
 goto_definition(Uri, #{ kind := record_expr, id := Record }) ->
   find(Uri, record, Record);
-goto_definition(_Uri, #{ kind := Kind, id := Include }
+goto_definition(_Uri, #{ kind := Kind, id := Id }
                ) when Kind =:= include;
                       Kind =:= include_lib ->
-  case els_utils:find_header(filename_to_atom(Include)) of
+  case els_utils:find_header(filename_to_atom(Id)) of
     {ok, Uri}      -> {ok, Uri, beginning()};
     {error, Error} -> {error, Error}
   end;
@@ -91,8 +91,8 @@ include_uris(Document) ->
   lists:foldl(fun add_include_uri/2, [], POIs).
 
 -spec add_include_uri(els_dt_document:item(), [uri()]) -> [uri()].
-add_include_uri(#{ id := String }, Acc) ->
-  case els_utils:find_header(filename_to_atom(String)) of
+add_include_uri(#{ id := Id }, Acc) ->
+  case els_utils:find_header(filename_to_atom(Id)) of
     {ok, Uri}       -> [Uri | Acc];
     {error, _Error} -> Acc
   end.
@@ -101,6 +101,6 @@ add_include_uri(#{ id := String }, Acc) ->
 beginning() ->
   #{range => #{from => {1, 1}, to => {1, 1}}}.
 
--spec filename_to_atom(string()) -> atom().
+-spec filename_to_atom(els_dt_document:id()) -> atom().
 filename_to_atom(FileName) ->
   list_to_atom(filename:basename(FileName, filename:extension(FileName))).
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -33,12 +33,12 @@ install(NodeName, BaseDir) ->
   lager:info("Creating DB [dir=~s]", [DbDir]),
   ok = filelib:ensure_dir(filename:join([DbDir, "dummy"])),
   ok = application:set_env(mnesia, dir, DbDir),
+  %% Avoid mnesia overload while indexing
+  ok = application:set_env(mnesia, dump_log_write_threshold, 50000),
   ensure_db().
 
 -spec ensure_db() -> ok.
 ensure_db() ->
-  %% Avoid mnesia overload while indexing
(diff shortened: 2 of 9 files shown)


src/els_workspace_symbol_provider.erl
```diff
@@ -30,17 +30,18 @@ handle_request({symbol, Params}, State) ->
 
 -spec modules(binary()) -> [symbol_information()].
 modules(Query) ->
-  {ok, All} = els_dt_document:find_by_kind(module),
+  {ok, All} = els_dt_document_index:find_by_kind(module),
+  Compare   = fun(#{id := X}, #{id := Y}) -> X < Y end,
+  AllSorted = lists:sort(Compare, All),
```
jfacorro: This sort add almost nothing to the time it takes to process this request.

src/els_db.erl
```diff
@@ -36,6 +37,8 @@ install(NodeName, BaseDir) ->
 
 -spec ensure_db() -> ok.
 ensure_db() ->
+  %% Avoid mnesia overload while indexing
+  application:set_env(mnesia, dump_log_write_threshold, 50000),
```
jfacorro: I added this to avoid all the noise in the logs.

src/els_code_navigation.erl
```diff
@@ -101,6 +101,6 @@ add_include_uri(#{ id := String }, Acc) ->
 beginning() ->
   #{range => #{from => {1, 1}, to => {1, 1}}}.
 
--spec filename_to_atom(string()) -> atom().
+-spec filename_to_atom(els_dt_document:id()) -> atom().
```
robertoaloi: How can this be valid? The `id` is defined as an `atom` and here we are expecting a `list`.

src/els_db.erl
robertoaloi: Should we move this line to the `install/2` function where we also specify the Mnesia `dir`? It feels more natural to do it there than in the function responsible to create the schema.

src/els_dt_document_index.erl
```diff
@@ -0,0 +1,97 @@
...
+
+-spec name() -> atom().
+name() -> ?MODULE.
+
+-spec opts() -> proplists:proplist().
+opts() ->
+  [ {attributes        , record_info(fields, els_dt_document_index)}
+  , {disc_copies       , [node()]}
+  , {index             , [#els_dt_document_index.kind]}
+  , {type              , bag}
```
robertoaloi: It seems like an overkill to have a bag to only support the fact the same id can be used for a module or a header. At this point, I would rather include the extension in the `id` (e.g. `file.hrl`) and use a `set`. Also this makes me think. Headers could indeed have the same `id` (independently of us including or not the extension), but different uris. In fact, whilst having two modules with clashing name is not typical, this can definitely happen for header files, sharing the same name but belonging to different apps. How do we plan to handle them?

src/els_indexer.erl
```diff
@@ -71,24 +71,30 @@ index(#{uri := Uri, text := Text} = Document) ->
...
-              ok = els_dt_references:delete_by_uri(Uri),
-              [register_reference(Uri, POI) || POI <- POIs],
-              ok
-          end,
+      F = fun() -> do_index(Document) end,
       els_db:transaction(F)
   end.
 
+-spec do_index(els_dt_document:item()) -> ok.
+do_index(#{uri := Uri, id := Module, kind := Kind} = Document) ->
```
robertoaloi: nit: The `Id` is not necessarily a `Module`.

src/els_workspace_symbol_provider.erl
robertoaloi: Did you consider to use an `ordered_set`?

src/els_code_navigation.erl
jfacorro: `filename:basename/2` and `filename:extension/1` work with type [`file:name_all()`](http://erlang.org/doc/man/file.html#type-name_all) which is defined as:
```
name_all() =
    string() | atom() | deep_list() | (RawFilename :: binary())
```

src/els_workspace_symbol_provider.erl
jfacorro: I hadn't 😄, but with the current approach the table needs to be a `bag` (ref. `bah` vs. `set` conversation in the other comment).

src/els_dt_document_index.erl
jfacorro: For headers with the same `id` and different `uri` there will be two different entries in the table since it's a `bag`. The same would happen with entries of type `other`.

> How do we plan to handle them?
We need to be mindful when using this index table of what we are trying to get and the fact that it's a table.

src/els_db.erl
jfacorro: Moved 😄

src/els_indexer.erl
jfacorro: Renamed the variable to `Id`.

> The Id is not necessarily a Module.

This was the case before the changes in this PR, right?

robertoaloi:
Let's merge this one, since it solves the current performance issues. We can refine it later.

PR #322, opened 2019-12-13
[#313] Index files with unknown extension and assign 'other' kind

### LINKED ISSUE #313: Error message when opening non erlang files
Whenever the client is working on a file that is not one of the three supported here:
```erlang
-spec new(uri(), binary()) -> item().
new(Uri, Text) ->
  Extension = filename:extension(Uri),
  Id         = binary_to_atom(filename:basename(Uri, Extension), utf8),
  case Extension of
    <<".erl">> ->
      new(Uri, Text, Id, module);
    <<".hrl">> ->
      new(Uri, Text, Id, header);
    <<".src">> ->
      new(Uri, Text, Id, src)
  end.
```
The following error is logged: 
```
2019-12-12 14:38:06.216 [error] <0.149.0>@els_methods:dispatch:47 Unexpected error [type=error] [error={case_clause,<<".spec">>}] [stack=[{els_dt_document,new,2,[{file,"/Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_dt_document.erl"},{line,154}]},{els_text_synchronization,did_o
pen,1,[{file,"/Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_text_synchronization.erl"},{line,13}]},{els_methods,textdocument_didopen,2,[{file,"/Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_methods.erl"},{line,174}]},{els_methods,dispatch,3,[{file,"/Users/juan.facorro/dev
/erlang-ls/erlang_ls/src/els_methods.erl"},{line,42}]},{els_server,handle_request,2,[{file,"/Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_server.erl"},{line,122}]},{lists,foldl,3,[{file,"lists.erl"},{line,1263}]},{els_server,handle_cast,2,[{file,"/Users/juan.facorro/dev/erlang
-ls/erlang_ls/src/els_server.erl"},{line,107}]},{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,637}]}]]
```

jfacorro:
### Description

Files like `rebar.config` or `application.app.src` should have the server's features available, so we should index files with unknown extension and assign kind `other` to them.

Fixes #313.

### DIFF
--- a/src/els_indexer.erl
+++ b/src/els_indexer.erl
@@ -187,16 +187,10 @@ terminate(_, _State) ->
 try_index_file(FullName, SyncAsync) ->
   try
     Uri = els_uri:uri(FullName),
-    Extension = filename:extension(Uri),
-    case els_dt_document:is_extension_supported(Extension) of
-      true ->
-        lager:debug("Indexing file. [filename=~s]", [FullName]),
-        {ok, Text} = file:read_file(FullName),
-        Document   = els_dt_document:new(Uri, Text),
-        ok         = index_document(Document, SyncAsync);
-      false ->
-        lager:debug("Skipping file. [filename=~s]", [FullName])
-    end
+    lager:debug("Indexing file. [filename=~s]", [FullName]),
+    {ok, Text} = file:read_file(FullName),
+    Document   = els_dt_document:new(Uri, Text),
+    ok         = index_document(Document, SyncAsync)
   catch Type:Reason:St ->
       lager:error("Error indexing file "
                   "[filename=~s] "
--- a/src/els_dt_document.erl
+++ b/src/els_dt_document.erl
@@ -25,7 +25,6 @@
         , pois/1
         , pois/2
         , get_element_at_pos/3
-        , is_extension_supported/1
         ]).
 
 %%==============================================================================
@@ -37,7 +36,7 @@
 %% Type Definitions
 %%==============================================================================
 -type id()   :: atom().
--type kind() :: module | header.
+-type kind() :: module | header | other.
 
 %%==============================================================================
 %% Item Definition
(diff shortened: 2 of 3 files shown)


src/els_indexer.erl
```diff
@@ -181,16 +181,10 @@ terminate(_, _State) ->
...
-    case els_dt_document:is_extension_supported(Extension) of
-      true ->
-        lager:debug("Indexing file. [filename=~s]", [FullName]),
-        {ok, Text} = file:read_file(FullName),
-        Document   = els_dt_document:new(Uri, Text),
-        ok         = index_document(Document, SyncAsync);
-      false ->
-        lager:debug("Skipping file. [filename=~s]", [FullName])
-    end
+    lager:debug("Indexing file. [filename=~s]", [FullName]),
```
robertoaloi: Why is this change being reverted? I don't think it's a good idea to index all files which sit under the project, especially if we don't know what they are. Opening this files, reading the content and attempting to parse them (to crash) sounds like a waste. If we want to support other extensions, we should just include them in my opinion. What features from the server do we use for `other` files?

jfacorro: During the initial indexing we are only processing files with extension `.hrl` and `.erl`, which will avoid indexing the bulk of all other types of files.

The crash is happening when we use the `els_dt_document:new/2` function, this happens in multiple places. I thought it would be more desirable to just handle the files that users mark as `erlang` type (i.e. in Emacs through specifying `erlang-mode`) than check the extension and type of file everytime we use the function above.

jfacorro:
> Why is this change being reverted?

Please see the reply to the other comment above.

src/els_indexer.erl
robertoaloi: Sorry, I missed the fact we have some additional filtering in the `index_dir` function. I thought we were trying to index all files.

src/els_completion_provider.erl
```diff
@@ -123,7 +123,7 @@ find_completion(_Prefix, _TriggerKind, _Opts) ->
   null.
 
 %%==============================================================================
-%% Modules
+%% modules
```
robertoaloi: Why? 🗡

jfacorro: I don't know!!! :disappointed:

jfacorro: I will revert this change and merge the PR.

alanz:
I see we now have a directory `test/els_indexer_SUITE_data` which has test data in it.

The balance of the test data is in the `priv` directory.

Is there a reason for using a different location?

robertoaloi:
@alanz Ideally all test data should sit in the standard `_data` dir like in this case. Unfortunately, `rebar3` tries to compile those files as well, though. So, in the case of the `code_navigation` application which has some intentional issues (e.g. non-existing behaviours, warnings, etc), compilation would fail. And that's why I had to move that application to the `priv`. This explanation is probably (hopefully!) part of the commit/pr where I introduced it.

PR #324, opened 2019-12-15
[#323] Limit the amount of workspace symbols returned

### LINKED ISSUE #323: Improve responsiveness for workspace symbols
Using the workspace symbols in a project where there are thousands of modules makes the interaction with this feature quite slow and awkard.

I'm pretty sure the slowness comes from the fact we are returning *all the modules* when the query is empty or just one or two letters. Limiting the amount of options shown will improve the experience a lot better.

jfacorro:
### Description

Limit the amount of workspace symbols returned.

Fixes #323.

### DIFF
--- a/src/els_workspace_symbol_provider.erl
+++ b/src/els_workspace_symbol_provider.erl
@@ -8,6 +8,8 @@
 
 -include("erlang_ls.hrl").
 
+-define(LIMIT, 100).
+
 %%==============================================================================
 %% els_provider functions
 %%==============================================================================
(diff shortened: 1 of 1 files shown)


src/els_workspace_symbol_provider.erl
```diff
@@ -35,17 +37,37 @@ modules(Query) ->
...
   {ok, RePattern} = re:compile(Query),
   ReOpts = [{capture, none}],
-  F = fun(#{id := Id}) ->
-          re:run(atom_to_binary(Id, utf8), RePattern, ReOpts) =:= match
+  F = fun(Name) ->
+          re:run(Name, RePattern, ReOpts) =:= match
       end,
-  lists:map(fun symbol_information/1, lists:filter(F, AllSorted)).
+  Filtered = filter_with_limit(AllSorted, F, ?LIMIT, []),
+  [symbol_information(X) || X <- lists:sort(Filtered)].
```
michalmuskala: Should this sort before filtering? Or is the order of items returned somehow relevant and sorted by something else and we're just re-sorting in here?

jfacorro: The sort that is happening on line `37` should already ensure a sorted result, this is a leftover after rebasing on `master`. If we just use `list:reverse/1` when returning from `filter_with_limit/4`, then the sorting is ensured.

Sorting the elements is nice since the client might not be sorting them, at least `lsp-mode` doesn't.

robertoaloi: I only have one concern with the limiting approach. From what I understand, the client sends an initial completion request, which can return many results. After that initial request, the list of candidates is narrowed down on the client side. So, in some cases, it would look like a certain symbol is undefined. This would be very problematic for certain kinds of completions (e.g. functions within a module), less problematic in this case case (workspace symbols). Are we ready to accept that behaviour?

jfacorro: For Emacs the implementation of `lsp-mode` a new request is sent for every character typed, which makes the experience very slow for the user when we return every single module, but limiting the results makes it extremely fast and responsive.

Maybe the main problem is that we are returning the whole list when the query is empty, but the [specs](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#workspace_symbol) define the query as a non-empty string:
```
/**
 * The parameters of a Workspace Symbol Request.
 */
interface WorkspaceSymbolParams {
	/**
	 * A non-empty query string
	 */
	query: string;
}
```
I can understand the limiting being a potential source of confusion for users if they are trying to get a list of all symbols that exist that match a (non-empty) query. I'm open to suggestions on how we can handle this while still making the feature usable.

robertoaloi: I think we should adhere to the specs and do not return anything in case an empty string is passed. If that's not enough to solve the issue, I'm ok with introducing a limit (ideally it should be configurable or possible to disable - but we don't have to do it now) as long as we inform the user. So maybe display a message at the same time saying the completion results have been limited?

alanz: Note: This method is not used for completion, it is used for searching for usages of a symbol across a project.

I wonder if we should not do some kind of "smart" limiting, by prioritising certain kinds of information.

jfacorro: > Note: This method is not used for completion, it is used for searching for usages of a symbol across a project.

That's a good point. The clients can implement this in diferent ways, and I guess in `lsp-mode` when using the command I'm using (i.e. `helm-lsp-workspace-symbol`) the implementation is an interactive one, where as you type you get the results. I haven't tried this in VSCode, though, the implementation in that client might work just fine when we return everything.

jfacorro:
After checking what is the behaviour in VSCode, I can see the unresponsiveness there as well. Not only that but VSCode returns the whole list of results as well when the query is empty.

Do we want to reject this PR and the ticket and just live with this? I really like the experience of listing the workspace symbols a lot better when it is being limited, but I can go either way.

robertoaloi:
I think we should favour the user experience and get rid of the un-responsiveness. So I'm all for including this one.

Issue #305, opened 2019-12-09
Add Windows platform to CI 


robertoaloi:
Some users have been reporting bugs in Windows, mostly due to path handling and path separators.
It shouldn't be a problem to support Windows in the server, but if we want to do so we need to add Windows as a platform to our CI pipelines to avoid regressions. I don't even know if this is a possibility in Travis or something else.

jfacorro:
[Clojure on the BEAM](https://github.com/clojerl/clojerl/blob/master/.appveyor.yml) uses [AppVeyor](https://ci.appveyor.com/project/jfacorro/clojerl-jqwev) but it seems that even though you can specify `21.1` the actual version of Erlang/OTP supported is [20.2](https://www.appveyor.com/docs/windows-images-software/#erlang). 😢

jfacorro:
I haven't used Circle CI but it seems like a possibility.

alanz:
I think CircleCI only supports Mac and Linux.  I know they are [investigating Azure](https://github.com/jneira/haskell-ide-engine/tree/hie-bios-azure) for haskell-ide-engine windows builds.

michalmuskala:
Elixir recently started using CirrusCI as one platform to do Linux, Windows, and FreeBSD.

Here's the [config for Linux + Windows](https://github.com/elixir-lang/elixir/blob/master/.cirrus.yml) and a PR for [FreeBSD](https://github.com/elixir-lang/elixir/pull/9599).

robertoaloi:
[GitHub actions](https://github.com/features/actions) seem to support Windows, so we may give them a try.

Issue #152, opened 2019-10-26
Explore GitHub Actions for CI


robertoaloi:
GitHub actions should be considered as an alternative CI method.

https://github.com/features/actions

PR #325, opened 2019-12-16
[#305] Build Workflow using GitHub Actions

robertoaloi:
### Description

Unfortunately, there's no Erlang installed by default on Windows or Mac systems and the existing `setup-erlang` actions do not work on Windows or Mac, so we would need to write one. For the time being, I'm simply adding a Linux-based workflow that we can keep in parallel to the Travis CI pipelines, to verify how both behave, which one is faster and so on. So, see this PR as a contribution to #152 which prepares #305.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,22 +1,31 @@
-name: Erlang CI
-
+name: Build
 on: [push]
-
 jobs:
-
   build:
-
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


robertoaloi:
Oh, and for the records the `container` approach is only available on Linux systems. And it wouldn't help anyway since we actually want to test on Windows and Mac themselves.

PR #328, opened 2019-12-17
[#327] Generate diagnostics when a file is opened

### LINKED ISSUE #327: Generate diagnostics when a file is opened

At the moment, diagnostics are only generated when a file is saved.

When a file is opened, the diagnostics should be provided.

alanz:
Fixes #327.

### DIFF
--- a/src/els_text_synchronization.erl
+++ b/src/els_text_synchronization.erl
@@ -1,5 +1,7 @@
 -module(els_text_synchronization).
 
+-include("erlang_ls.hrl").
+
 -export([ did_open/1
         , did_save/1
         , did_close/1
@@ -11,12 +13,24 @@ did_open(Params) ->
   Uri          = maps:get(<<"uri">>         , TextDocument),
   Text         = maps:get(<<"text">>        , TextDocument),
   Document     = els_dt_document:new(Uri, Text),
-  ok           = els_indexer:index(Document).
+  ok           = els_indexer:index(Document),
+  generate_diagnostics(Uri).
 
 -spec did_save(map()) -> ok.
 did_save(Params) ->
   TextDocument = maps:get(<<"textDocument">>, Params),
   Uri          = maps:get(<<"uri">>         , TextDocument),
+  generate_diagnostics(Uri).
+
+-spec did_close(map()) -> ok.
+did_close(_Params) -> ok.
+
+%%==============================================================================
+%% Internal functions
+%%==============================================================================
+
+-spec generate_diagnostics(uri()) -> ok.
+generate_diagnostics(Uri) ->
   CDiagnostics = els_compiler_diagnostics:diagnostics(Uri),
   DDiagnostics = els_dialyzer_diagnostics:diagnostics(Uri),
   EDiagnostics = els_elvis_diagnostics:diagnostics(Uri),
@@ -25,6 +39,3 @@ did_save(Params) ->
               , diagnostics => CDiagnostics ++ DDiagnostics ++ EDiagnostics
               },
   els_server:send_notification(Method, Params1).
-
--spec did_close(map()) -> ok.
-did_close(_Params) -> ok.
--- a/src/els_text_synchronization.erl
+++ b/src/els_text_synchronization.erl
@@ -7,6 +7,8 @@
         , did_close/1
         ]).
 
+-export([ generate_diagnostics/1 ]).
+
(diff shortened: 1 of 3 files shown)


alanz:
I am seeing something in my log that was not there before this commit

```
2019-12-17 16:13:13.510 [warning] <0.169.0> ** Undefined handle_info in els_server
** Unhandled message: {'EXIT',<0.657.0>,normal}
2019-12-17 16:13:41.781 [warning] <0.169.0> ** Undefined handle_info in els_server
** Unhandled message: {'EXIT',<0.693.0>,normal}
2019-12-17 16:15:31.815 [warning] <0.169.0> ** Undefined handle_info in els_server
** Unhandled message: {'EXIT',<0.695.0>,normal}
2019-12-17 16:15:33.291 [warning] <0.169.0> ** Undefined handle_info in els_server
** Unhandled message: {'EXIT',<0.696.0>,normal}
2019-12-17 16:22:01.893 [warning] <0.169.0>@els_methods:not_implemented_method:72 [Method not implemented] [method=$/cancelRequest]
2019-12-17 16:22:04.980 [warning] <0.169.0> ** Undefined handle_info in els_server
** Unhandled message: {'EXIT',<0.731.0>,normal}
2019-12-17 16:22:08.756 [warning] <0.169.0> ** Undefined handle_info in els_server
** Unhandled message: {'EXIT',<0.732.0>,normal}
```
Any idea where/how it could be coming about?

src/els_text_synchronization.erl
```diff
@@ -1,5 +1,7 @@
 -module(els_text_synchronization).
 
+-include("erlang_ls.hrl").
```
jfacorro: Why was this added?

jfacorro:
@alanz I'm not sure where those message are coming from, but the function `els_text_synchronization:did_save/1` is called with a spawn from the `els_methods:textdocument_didsave/2` function, I think we should do the same (at least for generating the diagnostics) when we call the `generate_diagnostics/1` function during `did_open/1`.

src/els_text_synchronization.erl
```diff
@@ -11,12 +13,24 @@ did_open(Params) ->
...
+
+-spec did_close(map()) -> ok.
+did_close(_Params) -> ok.
+
+%%==============================================================================
+%% Internal functions
+%%==============================================================================
+
+-spec generate_diagnostics(uri()) -> ok.
+generate_diagnostics(Uri) ->
```
jfacorro: I'm not sure where those message are coming from. But the function `els_text_synchronization:did_save/1` is spawned from the `els_methods:textdocument_didsave/2` function. I think we should do the same (at least for generating the diagnostics) when we call the `generate_diagnostics/1` function during `els_text_synchronization:did_open/1`.

robertoaloi: I agree. This should be done asynchronously.

src/els_text_synchronization.erl
alanz: Yes, I added it to fix compilation.

Aside: we should check that includes *are* actually needed.

src/els_text_synchronization.erl
alanz: Great. I thought it would make sense to do that, but wanted feedback.

alanz: Will do.

src/els_text_synchronization.erl
```diff
@@ -1,22 +1,39 @@
...
-  ok           = els_indexer:index(Document).
+  ok           = els_indexer:index(Document),
+  spawn (?MODULE, generate_diagnostics, [Uri]),
+  ok.
 
 -spec did_save(map()) -> ok.
 did_save(Params) ->
   TextDocument = maps:get(<<"textDocument">>, Params),
   Uri          = maps:get(<<"uri">>         , TextDocument),
+  generate_diagnostics(Uri).
```
jfacorro: Sorry I didn't mentioned it before but I think it would make more sense if we also used the spawn here instead of in `els_methods/didsave/2` function.

alanz:
One thing that worries me slightly is that we are not ensuring that we only have one compile running per file at a time.  Given the stateful interaction with the file system, this is probably something we should serialise. 

This will  become more of a concern when #329 gets resolved and we run dialyzer on save, as it potentially takes a long time to run too.

PR #334, opened 2019-12-19
[#329] Missing dialyzer diagnostics

### LINKED ISSUE #329: els_dialyzer_diagnostics does nothing
**Describe the bug**

`els_dialyzer_diagnostics:diagnostics` does not generate any diagnostics.

**To Reproduce**

Open a file with a dialyzer error, save it.

**Expected behavior**

The dialyzer error is reported as a diagnostic.

**Actual behavior**

No diagnostic is reported.

Note: changing the catch clause to log the error caught shows up the following

```
2019-12-17 16:31:32.933 [info] <0.453.0>@els_dialyzer_diagnostics:diagnostics:30 els_dialyzer: [Err={dialyzer_error,"Could not find the PLT: /Users/alanzimm/.dialyzer_plt\nUse the options:\n   --build_plt   to build a new PLT; or\n   --add_to_plt  to add to an existing PLT\n\nFor example, use a command like the following:\n   dialyzer --build_plt --apps erts kernel stdlib mnesia\nNote that building a PLT such as the above may take 20 mins or so\n\nIf you later need information about other applications, say crypto,\nyou can extend the PLT by the command:\n  dialyzer --add_to_plt --apps crypto\nFor applications that are not in Erlang/OTP use an absolute file name.\n"}]
```

**Context**
 - `erlang_ls` version (tag/sha): 84a9a0555aca07a18372b8ac7f376d9e86c50d8a
 - Editor used: emacs
 - LSP client used: lsp-mode

jfacorro:
### Description

Add a new `plt_path` configuration option. When it's not defined dialyzer diagnostics are not available.

I'm leaving the code lens approach we discussed for the future.

Fixes #329.

### DIFF
--- a/src/els_dialyzer_diagnostics.erl
+++ b/src/els_dialyzer_diagnostics.erl
@@ -25,8 +25,7 @@
 diagnostics(Uri) ->
   Path = els_uri:path(Uri),
   WS = try dialyzer:run([{files, [binary_to_list(Path)]}, {from, src_code}])
-       catch _:_ ->
-           []
+       catch _:_ -> []
        end,
   [diagnostic(W) || W <- WS].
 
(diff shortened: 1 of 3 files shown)


src/els_dialyzer_diagnostics.erl
```diff
@@ -24,19 +24,25 @@
...
-       end,
-  [diagnostic(W) || W <- WS].
+  case els_config:get(plt_path) of
+    undefined -> [];
+    DialyzerPltPath ->
+      WS = try dialyzer:run([ {files, [binary_to_list(Path)]}
+                            , {from, src_code}
+                            , {plts, [DialyzerPltPath]}
+                            ])
+           catch _:_ -> []
```
alanz: I think it is useful to log whatever the catch gets, even at debug level.

Issue #319, opened 2019-12-13
Wait for tables asynchronously


robertoaloi:
The user does not need to wait for tables. She should simply have a degraded service (no results or partial results) while the creation of the tables and the indexing are happening in the background.

jfacorro:
@robertoaloi Do we still want to spend time on trying this idea out?

robertoaloi:
If we can avoid the waiting time for the user on restart (and existing DB), I think we should do that.

alanz:
The "Wait for tables" step that blocks the response to the `initialize` message is one of the biggest blockers for us.

robertoaloi:
Some numbers from my experiments:

Code Base: 1.5 million lines of code

With empty tables it takes 400 ms to prepare tables.
Then it takes ~1m 40s to index the whole of OTP and the code base

With a Full DB, it takes ~25s to load tables and 10s to go through OTP and the same code base again, to figure out that everything is already indexed.

A Full DB means 4 tables:
Documents 600 MB
References 125 MB
Signatures 87 MB
Documents Index 1.2 MB

The table that takes the longest to load is the `references` table, which despite of the smaller size, has the biggest number of entries: 669.967

PR #342, opened 2019-12-20
[#319] Temporarily raise the timeout to infinity

robertoaloi:


onno-vos-dev:
Have a look at: https://github.com/erlang-ls/erlang_ls/pull/345 :) Seems like we ran into the same problem @robertoaloi  :wink:

PR #345, opened 2019-12-20
Add optional parameter to erlang_ls.config where the timeout to load mnesia tables can be specified

onno-vos-dev:
### Description

I've noticed that sometimes erlang_ls fails to start properly due to the `els_db:wait_for_tables` call timing out. The timeout is hardcoded to 20000 and cannot be overridden (as far as I could see?). This PR adds the option to set a longer timeout on a per project basis through the `erlang_ls.config`

### DIFF
--- a/README.md
+++ b/README.md
@@ -268,6 +268,7 @@ Currently, the following customizations are possible:
 | apps\_dirs         | List of directories containing project applications. It supports wildcards.                          |
 | include\_dirs      | List of directories provided to the compiler as include dirs. It supports wildcards.                 |
 | otp\_apps\_exclude | List of OTP applications that will not be indexed (default: megaco, diameter, snmp, wx)              |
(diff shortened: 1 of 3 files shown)


jfacorro:
First, thank you for taking the time to create the PR! ❤️ 

We are still fixing the issue regarding the DB taking too long to load, so I think I would rather prefer the PR where we raise this to `infinity` than adding a configuration for setting the timeout. This should be something the user does not need to think about.

onno-vos-dev:
Agreed @jfacorro I noticed while browsing through the recent issues that there's already a good bit of work in mind/ongoing to optimize this. :) 

Seeing as I adopted the use of erlang_ls today, I'll see where I can pitch in :+1: 

Great work on this project! :tada: <3

robertoaloi:
@onno-vos-dev Sorry about the conflict. Thanks for the contribution anyway!

PR #347, opened 2019-12-20
[#335] Use platform dependent log directory

### LINKED ISSUE #335: Log files are not produced in the correct directory
We should use:

```
filename:basedir(user_log, "erlang_ls").
```

To find the default location where to produce log files.

jfacorro:
### Description

Use platform dependent log directory.

Fixes #335.

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,8 @@ the Database Directory on your system:
 ### Logging
 
 When the escript is built using the `debug` profile as above, logging
-will be enabled and the logs will be written to `/usr/local/var/log/erlang_ls`.
+will be enabled and the logs will be written to your platform's log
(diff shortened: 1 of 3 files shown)


robertoaloi:
@alanz Why would you like to have logs in different dirs?

alanz:
I may be confused again, but I often end up with multiple servers running at the same time.
For some projects I want logging, others not. And each should log to a separate log file.

robertoaloi:
Sorry @alanz I totally misunderstood you. Yes, a log dir per project would be desirable, as long as the project logs will all be within the same folder. So, for example: `/whatever/logs/erlang_ls/{project_a,project_b}`. Will create an issue for that.

PR #349, opened 2019-12-20
[#333] Only process record access if the field is an atom

### LINKED ISSUE #333: Record field accessors assumed to be atoms
**Description**
Various things crash when editing a file with code like `Var#rec.?MACRO` as els_parser assumes the record field accessor should be an atom.

Error message:
```
 [error] <0.170.0>@els_methods:dispatch:50 Unexpected error [type=error] [error=badarg] 
[stack=[{erlang,atom_to_list,[{macro,{tree,variable,{attr,{12,9},[],none},'A'},none}],[]},
{erl_syntax,atom_name,1,[{file,"erl_syntax.erl"},{line,2038}]},
{els_parser,record_access,1,[{file,"…/src/els_parser.erl"},{line,291}]},
{els_parser,do_points_of_interest,1,[{file,"…/src/els_parser.erl"},{line,147}]},
{els_parser,'-points_of_interest/1-fun-0-',2,[{file,"…/src/els_parser.erl"},{line,134}]},
{erl_syntax_lib,fold_2,3,[{file,"erl_syntax_lib.erl"},{line,135}]},
{erl_syntax_lib,fold_1,3,[{file,"erl_syntax_lib.erl"},{line,130}]},
{erl_syntax_lib,fold,3,[{file,"erl_syntax_lib.erl"},{line,126}]}]]
```
**Context**
 - `erlang_ls` version (tag/sha): 84a9a05
 - Editor used: emacs
 - LSP client used: lsp-mode

jfacorro:
### Description

Only process record access if the field is an atom.

Fixes #333.


src/els_parser.erl
```diff
@@ -285,10 +285,11 @@ macro(Tree) ->
 -spec record_access(tree()) -> [poi()].
 record_access(Tree) ->
   RecordNode = erl_syntax:record_access_type(Tree),
-  case erl_syntax:type(RecordNode) of
-    atom ->
+  FieldNode = erl_syntax:record_access_field(Tree),
+  case {erl_syntax:type(RecordNode), erl_syntax:type(FieldNode)}  of
```
robertoaloi: I thought we wanted to support cases when a macro is used instead of an atom.
If I'm reading this correctly, we are explicitly ignoring all cases where the field name is an atom.
This could be problematic if, for example, in the future we want to find all accesses to a record field.
Isn't it possible to add support for macros instead?

jfacorro: > If I'm reading this correctly, we are explicitly ignoring all cases where the field name is an atom.

Did you mean we  are ignoring all cases where the field name is a *macro*?

> Isn't it possible to add support for macros instead?

It is possible but I don't know what you have in mind though? Are you proposing to:
1. Keep track of the values for all macros defined in the source (`.hrl` files included) and try to figure out the value of the macro used in the field?
2. Or register the record access with a dummy value for the field?

I would vote for option 2 since the option 1 get complicated very fast (i.e. what happens with macros with arguments?).

robertoaloi: I would also lean towards `2`.

PR #350, opened 2019-12-21
[#340] Report compiler messages for header files

### LINKED ISSUE #340: Report diagnostics for header files
**Describe the bug**
Diagnostics are not reported for `.hrl` filles.

**To Reproduce**
Create a module which includes a header file. Introduce a compilation error in the header file, the compilation error is not reported.

**Expected behavior**
The user can see diagnostics in header files.

This could be achieved by creating a fake module which includes the header file.

robertoaloi:
### Description

So far, we have not reported compiler messages for included files. This PR addresses this issue, by reporting them in the `.hrl` files. In the `.erl` file including a broken header the inclusion line is marked as erroneous and the compiler messages from the `.hrl` file are visible to the user on the same line.

Fixes #340.


src/els_compiler_diagnostics.erl
```diff
@@ -50,36 +61,116 @@ diagnostics(Uri) ->
...
+                , integer()) -> diagnostic().
+diagnostic(Path, Path, Range, _Document, Module, Desc, Severity) ->
+  %% The compiler message is related to the same .erl file, so
+  %% preserve the location information.
+  diagnostic(Range, Module, Desc, Severity);
+diagnostic(_Path, MessagePath, Range, Document, Module, Desc0, Severity) ->
+  #{from := {Line, _}} = Range,
+  InclusionRange = inclusion_range(MessagePath, Document),
+  %% The compiler message is related to an included file. Replace the
+  %% original location with the location of the file inclusion.
```
alanz: It is possible to return diagnostics for other files, which are then shown at the appropriate locations in the other file when opened.

So a general strategy should be to show a diagnostic in the .erl file at the location of the include, saying there are problems with the file, but then report diagnostics for the .hrl itself.

I have not run this code, so maybe that does in fact happen.

robertoaloi: Yes, this is exactly what it does.

PR #362, opened 2019-12-26
[#359] Folding provider

### LINKED ISSUE #359: Implement folding provider

robertoaloi:
### Description

The folding provider allows to collapse/expand parts of the code. Right now this is implemented only for functions.

Fixes #359 .

### DIFF
--- a/elvis.config
+++ b/elvis.config
@@ -11,7 +11,6 @@
                                                                         ]}}
                       , {elvis_style, line_length, #{limit => 80, skip_comments => false}}
                       , {elvis_style, operator_spaces, #{ rules => [ {right, ","}
-                                                                   , {right, "-"}
                                                                    , {left , "-"}
                                                                    , {right, "+"}
                                                                    , {left , "+"}
--- a/src/els_methods.erl
+++ b/src/els_methods.erl
@@ -43,6 +43,7 @@
 -spec dispatch(method_name(), params(), state()) -> result().
 dispatch(Method, Params, State) ->
   Function = method_to_function_name(Method),
+  lager:debug("Dispatching request [method=~p] [params=~p]", [Method, Params]),
   try do_dispatch(Function, Params, State)
   catch
     error:undef ->
--- a/src/els_methods.erl
+++ b/src/els_methods.erl
@@ -24,6 +24,7 @@
         , textdocument_formatting/2
         , textdocument_rangeformatting/2
         , textdocument_ontypeformatting/2
+        , textdocument_foldingrange/2
         , workspace_didchangewatchedfiles/2
         , workspace_symbol/2
         ]).
@@ -139,6 +140,8 @@ initialize(Params, State) ->
               els_formatting_provider:is_enabled_range()
           , documentOnTypeFormattingProvider =>
               els_formatting_provider:is_enabled_on_type()
+          , foldingRangeProvider =>
+              els_folding_range_provider:is_enabled()
           %% AZ: didchangewatchedfiles is not listed in
           %%     ServerCapabilities in the LSP spec.
           , didChangeWatchedFiles =>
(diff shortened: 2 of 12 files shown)


elvis.config
```diff
@@ -11,7 +11,6 @@
                                                                         ]}}
                       , {elvis_style, line_length, #{limit => 80, skip_comments => false}}
                       , {elvis_style, operator_spaces, #{ rules => [ {right, ","}
-                                                                   , {right, "-"}
```
robertoaloi: Elvis was requesting a space for negative numbers, which I don't want. Not sure if there's a way to enforce space only when the `-` is used as a binary operator and not as a unary one.

src/els_methods.erl
```diff
@@ -42,6 +43,7 @@
 -spec dispatch(method_name(), params(), state()) -> result().
 dispatch(Method, Params, State) ->
   Function = method_to_function_name(Method),
+  lager:debug("Dispatching request [method=~p] [params=~p]", [Method, Params]),
```
robertoaloi: We were not logging requests on the server side.

src/els_parser.erl
```diff
@@ -47,20 +47,22 @@ parse_form(IoDevice, Location, Options) ->
   {'ok', erl_syntax:forms() | none, integer()}
   | {'eof', integer()}
   | {'error', any(), integer()}.
-parse_form(IoDevice, Location0, Parser, _Options) ->
-  case io:scan_erl_form(IoDevice, "", Location0) of
-    {ok, Tokens, Location1} ->
+parse_form(IoDevice, StartLocation, Parser, _Options) ->
```
robertoaloi: Just renamed the variables for clarity.

src/els_parser.erl
```diff
@@ -47,20 +47,22 @@ parse_form(IoDevice, Location, Options) ->
...
-          POIs = [find_attribute_pois(F, Tokens), points_of_interest(F)],
-          {ok, POIs, Location1}
+          POIs = [ find_attribute_pois(F, Tokens)
+                 , points_of_interest(F, EndLocation)
+                 ],
+          {ok, POIs, EndLocation}
       catch
         _:_ ->
-          {ok, find_attribute_tokens(Tokens), Location1}
+          {ok, find_attribute_tokens(Tokens), EndLocation}
```
robertoaloi: I could have also passed the `StartLocation`, but I decided to go for the minimum possible change here.

src/els_parser.erl
```diff
@@ -73,15 +75,15 @@ find_attribute_pois(Form, Tokens) ->
           %% The first atom is the attribute name, so we skip it.
           [_|Atoms] = [T || {atom, _, _} = T <- Tokens],
           ExportEntries =
-            [ els_poi:new(Pos, export_entry, {F, A})
+            [ poi(Pos, export_entry, {F, A})
```
robertoaloi: The `els_poi:new/3` function takes a `location`. I believe it should take a range, instead. For now (to avoid a big refactoring) I'm simply adding a wrapper function, but this should be cleaned up. Also, the `els_range` module does not really make sense outside of the `els_poi` and the two modules could be merged.

src/els_server.erl
```diff
@@ -123,13 +123,13 @@ handle_request(Request, #state{internal_state = InternalState} = State0) ->
     {response, Result, NewInternalState} ->
       RequestId = maps:get(<<"id">>, Request),
       Response = els_protocol:response(RequestId, Result),
-      lager:debug("[SERVER] Sending response [response=~p]", [Response]),
+      lager:debug("[SERVER] Sending response [response=~s]", [Response]),
```
robertoaloi: Show the strings correctly in the log messages.

elvis.config
jfacorro: This is a bug, it shouldn't do that. 

There is a test case that checks for operators with one operand [here](https://github.com/inaka/elvis_core/pull/112/files#diff-1cb0d1d3d593a8deed4e80e20871c293R45). And the check is done [here](https://github.com/inaka/elvis_core/pull/112/files#diff-9d4ce82551a31e44e1d37fb5c5966363R292). 

Can you share the snippet of code that was getting the false positive?

src/els_provider.erl
```diff
@@ -24,7 +24,8 @@
                   | els_references_provider
                   | els_formatting_provider
                   | els_document_highlight_provider
-                  | els_workspace_symbol_provider.
+                  | els_workspace_symbol_provider
+                  | els_folding_range_provider.
```
jfacorro: Nit: alphabetical order 😜

elvis.config
robertoaloi: @jfacorro From `els_parser.erl`:

```
                      Range = #{ from => {StartLine, StartColumn}
                               , to   => {EndLine - 1, -1}
                               },
```

PR #364, opened 2019-12-27
[#363] Add implementation provider and support for gen_server callbacks

### LINKED ISSUE #363: Add implementation provider

robertoaloi:
### Description

A maybe opinionated way to use the _implementation_ provider to navigate to callback functions for OTP behaviours.

Something that I longed for a long time was the ability to navigate from a call such as `gen_server:call(...)` to the respective `handle_call`. To limit the magic happening, this is restricted to the same module.

Fixes #363.

### DIFF
--- a/src/els_implementation_provider.erl
+++ b/src/els_implementation_provider.erl
@@ -1,0 +1,76 @@
+-module(els_implementation_provider).
+
+-behaviour(els_provider).
+
+-include("erlang_ls.hrl").
+
+-export([ handle_request/2
+        , is_enabled/0
+        ]).
+
+%%==============================================================================
+%% els_provider functions
+%%==============================================================================
+
+-spec is_enabled() -> boolean().
+is_enabled() ->
+  true.
+
+-spec handle_request(tuple(), els_provider:state()) ->
+   {null | location(), els_provider:state()}.
+handle_request({implementation, Params}, State) ->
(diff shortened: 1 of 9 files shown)


src/els_implementation_provider.erl
```diff
@@ -0,0 +1,100 @@
...
+      [POI || #{id := {F, A}} = POI <- POIs, F =:= CF, A =:= CA];
+    undefined ->
+      []
+  end;
+implementation(_Document, _POI) ->
+  [].
+
+-spec callback(mfa()) -> {atom(), non_neg_integer()} | undefined.
+%% gen_event
+callback({gen_event, add_handler, 3})     -> {init, 1};
```
jfacorro: 😱

robertoaloi: Is the 😱because of the manually maintained list? That list comes from the OTP documentation and it only changes sporadically.

robertoaloi:
On the usefulness side, I would say it helps in most daily situations. Regarding the support for custom behaviours, something could be done to support those as well, but I suggest we include this as an experiment, we try it, we see how users react and if it creates any confusion. Then, we reconsider if we want to drop it or extend it.

src/els_implementation_provider.erl
jfacorro: Yeah, that was what the face was about 😜.

PR #376, opened 2020-01-08
[#375] Delete only references with the provided URI

### LINKED ISSUE #375: References deleted when file re-indexed
**Describe the bug**
The bug is that we are deleting the references of all functions when re-indexing a file [here](https://github.com/erlang-ls/erlang_ls/blob/master/src/els_indexer.erl#L94). 

We are doing this `els_db:delete(name(), Id)`  [here](https://github.com/erlang-ls/erlang_ls/blob/master/src/els_dt_references.erl#L86) which removes all references to that entry 🤦‍♂.
 

**To Reproduce**
In a fresh project (with no DB cache) open a file where there's an exported function that used in the same module *and* another module and then:
1. List references for that function. Check that all are there.
2. Modify the file and save it to force re-indexing.
3. List references for the same function. Observe the ones from the other modules are not there.

**Expected behavior**
References from modules with other URI shouldn't be deleted.

**Actual behavior**
Every reference is deleted.

**Context**
 - `erlang_ls` version (tag/sha): `634cc2a`
 - Editor used: Emacs
 - LSP client used: `lsp-mode`

jfacorro:
### Description

Delete only references with the provided URI

Fixes #375.

### DIFF
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -4,6 +4,7 @@
 -export([ clear_table/1
         , clear_tables/0
         , delete/2
+        , delete_object/1
         , install/2
         , lookup/2
         , match/1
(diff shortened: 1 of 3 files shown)


robertoaloi:
@jfacorro Tests seem to fail.

jfacorro:
@robertoaloi Fixed 😄

PR #379, opened 2020-01-10
fix not shutting down properly when using stdio transport

luos:
### Description

Handle method 'exit' received from the stdio transport.
Also change halt to init:stop to tear down gracefully.

I see in the meantime `els_db:stop(),` call was added, I think this can be removed as `init:stop` should tear down the applications in dependency order. 

Fixes erlang-ls/vscode#11 .

### DIFF
--- a/src/els_methods.erl
+++ b/src/els_methods.erl
@@ -203,7 +203,7 @@ exit(_Params, State) ->
                _        -> 1
              end,
   els_utils:halt(ExitCode),
-  {noresponse, State}.
+  {noresponse, #{}}.
 
 %%==============================================================================
 %% textDocument/didopen
--- a/src/els_methods.erl
+++ b/src/els_methods.erl
@@ -197,11 +197,13 @@ shutdown(_Params, State) ->
 
 -spec exit(params(), state()) -> no_return().
 exit(_Params, State) ->
+  lager:info("Language server stopping..."),
   ExitCode = case maps:get(status, State, undefined) of
                shutdown -> 0;
                _        -> 1
              end,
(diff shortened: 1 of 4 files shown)


alanz:
I presume the same logic could apply to the tcp server?

luos:
I am not familiar how the TCP LS works.

Is that also started by the Editor or is it started by the user? As It's possible to reconnect to a running TCP server, I am not sure it's a good idea to stop the TCP server if the connection is lost. In return it's pretty hard to reconnect to STDIN/STDOUT pair, so that's why this behavior for sure makes sense for VSCode.

alanz:
From the LSP client perspective I am pretty sure most of them start the server then connect to it, and if a connection is lost they restart the whole thing.

But perhaps it is best to confirm the solution with stdio first, then consider the tcp case.

luos:
Hi,

I managed to fix what was the issue, until now `{message, EmpyMap}` was returned from halt, and I changed it to `State`. I am not sure it's correct to just throw out all the state.

jfacorro:
> I managed to fix what was the issue.

That is awesome! 😄 

> I am not sure it's correct to just throw out all the state.

Until now the call to `els_utils:halt/1` would make the Erlang VM go down, so throwing out the state made sense, since it would be like having a new server from scratch. 

Now that we are using `init:stop/1` the Erlang node will stay up for a little while, but any request to it should be ignored (i.e. not dispatched). An ideal approach would be to transition the `status` into something that can be used in `els_methods:dispatch/2` to ignore everything. But I don't think this is a requirement for now.

robertoaloi:
Thanks!

PR #380, opened 2020-01-10
Use first config file found in a list of paths…

bolmar:
… allowing falling back to a global config file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,14 +264,14 @@ configuration file.
 
 Generally speaking, <app config path> should be:
 
-| Operating System | Database Dir                                        |
+| Operating System | Configuration Directory                             |
 |------------------|-----------------------------------------------------|
 | Linux            | /home/USER/.config/erlang\_ls                       |
 | OS X             | /Users/USER/Library/Application\ Support/erlang\_ls |
 | Windows          | c:/Users/USER/AppData/Local/erlang\_ls              |
 
 You can also run the following command on an Erlang shell to identify
-the Database Directory on your system:
+the Configuration Directory on your system:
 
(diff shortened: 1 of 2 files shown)


src/els_config.erl
```diff
@@ -21,7 +21,6 @@
 %%==============================================================================
 %% Macros
 %%==============================================================================
--define(DEFAULT_CONFIG_PATH, "erlang_ls.config").
```
jfacorro: After this change the code is still using this constant, so it would be nice to keep it.

README.md
```diff
@@ -252,9 +252,31 @@ Code completion is available for the following elements:
...
+* <project root>/erlang_ls.config
+* <app config path>/erlang_ls/erlang_ls.config
+
+Generally speaking, <app config path> should be:
+
+| Operating System | Configuration Directory                             |
+|------------------|-----------------------------------------------------|
+| Linux            | /home/USER/.config/erlang\_ls                       |
+| OS X             | /Users/USER/Library/Application\ Support/erlang\_ls |
+| Windows          | c:/Users/USER/AppData/Local/erlang\_ls              |
```
alanz: The `<app config path>` in the table already includes the `erlang_ls` directory, I presume it will not be repeated?

robertoaloi:
> Would it make sense to load the config files in order, from the default, to the user, to the project, with each replacing what is already there.
> 
> This allows "general" settings, which can be overridden at the project level as needed.

It's an interesting idea, but I guess it could be implemented separately from this PR. Maybe just open a ticket for this?

robertoaloi:
Agreed with @alanz to merge this.

PR #390, opened 2020-01-15
[#389] Remove invalid capabilities and/or values

### LINKED ISSUE #389: InelliJ LSP Client fails to initialize session
**Describe the bug**
The IntelliJ LSP client fails to initialize when using `erlang_ls`. After sending the `initialize` request it immediately sends a `shutdown` request.

**To Reproduce**
Just try to use the lsp-client with the latest version.

**Expected behavior**
The client should connect to the server.

**Actual behavior**
Shutdown and exit requests are sent right after initialize.

**Context**
 - `erlang_ls` version (tag/sha): `00b5296`
 - Editor used: `IntelliJ IDEA 2019.1`
 - LSP client used: `LSP Support v1.6.0`

jfacorro:
### Description

Remove invalid capabilities and/or values.

Fixes #389


src/els_methods.erl
```diff
@@ -151,16 +151,12 @@ initialize(Params, State) ->
               els_formatting_provider:is_enabled_document()
           , documentRangeFormattingProvider =>
               els_formatting_provider:is_enabled_range()
-          , documentOnTypeFormattingProvider =>
-              els_formatting_provider:is_enabled_on_type()
+          %%, documentOnTypeFormattingProvider =>
+          %%    els_formatting_provider:is_enabled_on_type()
```
alanz: Given the bug that is being fixed, this one should be ok, it just signals that it is not supported. It is in fact a valid capability.

robertoaloi: If I understood this correctly, IntelliJ implements version `3.14` of the protocol (understandable, since `3.15` just came out) and in that version this field cannot be a `boolean`, but it has to be a _struct_. This is addressed in `3.15`. I guess it all depends if we want to consider that as a protocol bug and/or support IntelliJ.
@alanz Is this the bug you are referring to, or is it something else?

jfacorro:
I have to update the tests.

src/els_methods.erl
alanz: I saw the bug report as invalid capabilities returned. I know the one is invalid (with my longstanding comment), was not sure of the other. If it is a problem it may as well stay commented out, it has the same practical interpretation.

PR #391, opened 2020-01-15
Add nop `workspace/didChangeconfiguration` processor.

### LINKED ISSUE #373: Implement `workspace/didChangeConfiguration`
As a placeholder to start with, it does get sent from vscode and any other clients if any config changes, even at the client level.

alanz:
Some clients such a vscode send this message on startup, it results in
a confusing pop-up message on the client.

Fixes #373


alanz:
Oops, should be workspace, not textDocument.

PR #397, opened 2020-01-17
[#406] Add and use `pa_dirs` configuration setting.

### LINKED ISSUE #406: Compiler diagnostics do not pick up behaviour defined in a project
**Describe the bug**

If a project defines a behaviour, which is used in an erlang source file, the compiler diagnostics will report it as an unknown reference.

**To Reproduce**

Create a simple project with a user defined behaviour, and a module using it. Open this using `erlang_ls`.

**Expected behavior**

No diagnostics are reported.

**Actual behavior**

An undefined variable error is reported.

**Context**
 - Editor used:
 - LSP client used:

alanz:
This allows compiler diagnostics to load existing behaviour beam files, avoiding false error reporting.

It does this by adding a new config file option, `pa_dirs`, corresponding to the `-pa` command line flag.

Fixes #406  .

### DIFF
--- a/README.md
+++ b/README.md
@@ -400,6 +400,8 @@ deps_dirs:
 include_dirs:
   - "include"
   - "_build/default/lib"
+pa_dirs:
+  - "_build/default/lib/mylib/ebin"
 macros:
   - name: DEFINED_WITH_VALUE
     value: 42
@@ -417,6 +419,7 @@ The following customizations are possible:
 | deps\_dirs         | List of directories containing dependencies. It supports wildcards.                                                                       |
 | apps\_dirs         | List of directories containing project applications. It supports wildcards.                                                               |
 | include\_dirs      | List of directories provided to the compiler as include dirs. It supports wildcards.                                                      |
+| pa\_dirs           | List of directories containing beam files for e.g. behaviours.                                                                            |
 | macros             | List of cusom macros to be passed to the compiler, expressed as a name/value pair. If the value is omitted or is invalid, 'true' is used. |
 | otp\_apps\_exclude | List of OTP applications that will not be indexed (default: megaco, diameter, snmp, wx)                                                   |
 | code\_reload       | Whether or not an rpc call should be made to a remote node to compile and reload a module                                                 |
--- a/priv/behaviour/LICENSE
+++ b/priv/behaviour/LICENSE
@@ -1,0 +1,191 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
+      Object form, made available under the License, as indicated by a
+      copyright notice that is included in or attached to the work
+      (an example is provided in the Appendix below).
+
+      "Derivative Works" shall mean any work, whether in Source or Object
+      form, that is based on (or derived from) the Work and for which the
+      editorial revisions, annotations, elaborations, or other modifications
+      represent, as a whole, an original work of authorship. For the purposes
+      of this License, Derivative Works shall not include works that remain
+      separable from, or merely link (or bind by name) to the interfaces of,
+      the Work and Derivative Works thereof.
+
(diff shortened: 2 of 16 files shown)


alanz:
This should probably have a default value too, to be 

    - "_build/default/lib/*/ebin"

src/els_compiler_diagnostics.erl
```diff
@@ -71,6 +71,8 @@ compile(Uri) ->
   Includes = [ {i, IncludePath}
                || IncludePath <- els_config:get(include_paths)
              ],
+  %% Extra paths to load beam files from, for e.g. behaviours
+  [code:add_patha(Dir) || Dir <- els_config:get(pa_dirs)],
```
robertoaloi: The approach proposed relies on loading binaries onto the server which have been produced by some external tool. This could be problematic and lead to a series of edge cases which could be confusing for the user (e.g. if different compiler options have been used).

Imagine the following scenario: the developer works on a behaviour and its implementation at the same time (two buffers open). When saving the behaviour, changes are not reflected on the server. The user should manually re-trigger compilation using an external tool (rebar3 in this case) for the server to pick them up.

I'd suggest an alternative approach. If the current module contains a behaviour definition (and, eventually, a parse transform), we look for the behaviour URI. We compile the behaviour and load the binary into memory (no need to persist the .beam to disk). The compilation of the current module would then succeed. We can revert the load if we want to avoid persisting the side effect.

This whole thing also leads to a reflection. As soon as we start operating on binary files (persisted or not), we need to/should ensure that the language server node is running the version of OTP specified by the `erlang_ls.config` or issues may arise. This is not the case at the moment, where the OTP version specified by the user is only used for indexing, not to select the runtime. Dynamically selecting the run-time could also problematic, since the server would have been compiled with one version of OTP, but use a different one at runtime. This was one of the major pain points with the EDTS project.

test/els_diagnostics_SUITE.erl
```diff
@@ -53,7 +54,23 @@ end_per_suite(Config) ->
 -spec init_per_testcase(atom(), config()) -> config().
 init_per_testcase(TestCase, Config) ->
   mock_notifications(),
-  els_test_utils:init_per_testcase(TestCase, Config).
+  Fun = fun () ->
```
robertoaloi: The project uses 2-spaces indentation, not 4, so we should try to be consistent.
@jfacorro is it possible to enforce this via Elvis?

test/els_test_utils.erl
```diff
@@ -71,23 +76,27 @@ init_per_suite(Config) ->
   DiagnosticsIncludePath = filename:join([ RootPath
                                          , <<"include">>
                                          , <<"diagnostics.hrl">>]),
+  DiagnosticsBehaviourPath = filename:join([ RootBehaviourPath
```
robertoaloi: Not related to this PR, so feel free to ignore it, but we should probably refactor this module and get rid of the duplications in this module.

README.md
```diff
@@ -356,8 +358,11 @@ The following customizations are possible:
 | deps\_dirs         | List of directories containing dependencies. It supports wildcards.                                  |
 | apps\_dirs         | List of directories containing project applications. It supports wildcards.                          |
 | include\_dirs      | List of directories provided to the compiler as include dirs. It supports wildcards.                 |
+| pa\_dirs         | List of directories containing beam files for e.g. behaviours. It supports wildcards`*`.                 |
```
robertoaloi: Not sure if we will go for this approach (see my comments below), but in case we do this should say 'It does not support wildcards at the moment', instead of the 'we support (not)' :)

priv/behaviour/LICENSE
```diff
@@ -0,0 +1,191 @@
+                                 Apache License
```
robertoaloi: Any reason why the `priv` directory was preferred to the `_SUITE_data` directory in this case?

test/els_diagnostics_SUITE.erl
robertoaloi: I never used editorconfig, but I'm all for it.

src/els_compiler_diagnostics.erl
robertoaloi: Something like:

```
{ok, my_module, B} = compile:file("/path/to/my_module.erl", [binary]).
code:load_binary(my_module, no_name, B).
```

michalmuskala: An alternative would be to filter out the `undefined_behaviour` warnings and re-implement the verification logic based on just parsing the files inside of erlang_ls without actually loading any modules

priv/behaviour/LICENSE
alanz: And the reason it is in priv is that it deliberately has warnings in it, to confirm we can get diagnostics when there is a behaviour. So it breaks the "compile clean" rule for the main stuff.  So it is the same reason as the other stuff currently in priv.

src/els_compiler_diagnostics.erl
alanz: I presume the approach should be to do the compile as now, then pick up "undefined_behaviour" warnings and use that to trigger an attempted load of the behaviour and recompile.  Being careful to avoid infinite loops.

alanz: > When saving the behaviour, changes are not reflected on the server.

Thinking about this, we currently invoke the compiler for the file to generate diagnostics. Are you proposing we *never* compile to beam, to avoid messing with the underlying build system. Which makes sense to me.  Or even do a cheap compile most times and avoid code gen.

alanz: Which I see already happens, via `basic_validation` flag.

alanz:
> I would prefer not relying on .beam files produced outside the language server, but I'm open to discuss this further.

I have just pushed a version that loads the binary into the running server, if it is asked to generate diagnostics for a module having `callback` attributes in it.

If we agree this is the approach to take, I will concoct a test for it. I already tried manually adding a callback to `mylib_behaviour`, saving it, and without recompiling via rebar checking that diagnostics on `mylib_impl` report the missing callback when I save it.

robertoaloi:
I wanted to give this I try myself. I created #433 which solves #406 while not adding a `pa` option. Essentially, instead of recompiling/parsing the file, I'm using the information which is already in the database (i.e. the list of pois for the current module and the path of the required behaviour). Before loading a behaviour, I'm getting a copy of the old code so that it can be restored afterwards. See what you think.

alanz:
I will give it a try and see if it works as advertised. It is definitely preferable not to need additional config.

alanz:
Closing in favour of #433

PR #409, opened 2020-01-23
[#405] handle error in els_compiler_diagnostic:parse/1

### LINKED ISSUE #405: els_compiler_diagnostics:parse does not handle the error case
**Describe the bug**

The function [`els_compiler_diagnostics:parse`](https://github.com/erlang-ls/erlang_ls/blob/fe765af40932d593367786bc41c65d5327e9757d/src/els_compiler_diagnostics.erl#L82-L91) does not deal with the `{error,_}` case.

```erlang
-spec parse(uri()) -> [diagnostic()].
parse(Uri) ->
  FileName = binary_to_list(els_uri:path(Uri)),
  {ok, Epp} = epp:open([ {name, FileName}
                       , {includes, els_config:get(include_paths)}
                       ]),
```

**Context**
 - Editor used:
 - LSP client used:

alanz:
If a `.hrl` file does not exist, or does not open for any reason, log
a warning and return an empty diagnostic list.

Fixes #405.

### DIFF
--- a/src/els_compiler_diagnostics.erl
+++ b/src/els_compiler_diagnostics.erl
@@ -85,13 +85,23 @@ compile(Uri) ->
 -spec parse(uri()) -> [diagnostic()].
 parse(Uri) ->
   FileName = binary_to_list(els_uri:path(Uri)),
-  {ok, Epp} = epp:open([ {name, FileName}
-                       , {includes, els_config:get(include_paths)}
-                       ]),
-  Res = [diagnostic(range(Line), Module, Desc, ?DIAGNOSTIC_ERROR)
-         || {error, {Line, Module, Desc}} <- epp:parse_file(Epp)],
-  epp:close(Epp),
-  Res.
+  case epp:open([ {name, FileName}
+                , {includes, els_config:get(include_paths)}
+                ]) of
+      {ok, Epp} ->
+          Res = [diagnostic(range(Line), Module, Desc, ?DIAGNOSTIC_ERROR)
+                 || {error, {Line, Module, Desc}} <- epp:parse_file(Epp)],
+          epp:close(Epp),
+          Res;
+      _ ->
+          Msg = io_lib:format("Compile: could not parse ~p", [FileName]),
+          lager:info(Msg),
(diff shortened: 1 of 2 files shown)


alanz:
I am not sure if it is better to perhaps wrap the whole thing in a try/catch.

Also, whether to synthesize a diagnostic in the failure case, indicating a parse failure.

src/els_compiler_diagnostics.erl
```diff
@@ -82,13 +82,18 @@ compile(Uri) ->
...
-  Res.
+  case epp:open([ {name, FileName}
+                , {includes, els_config:get(include_paths)}
+                ]) of
+      {ok, Epp} ->
+          Res = [diagnostic(range(Line), Module, Desc, ?DIAGNOSTIC_ERROR)
+                 || {error, {Line, Module, Desc}} <- epp:parse_file(Epp)],
+          epp:close(Epp),
+          Res;
+      _ ->
```
robertoaloi: I think we could (as you suggest) show an error to the user in this case. Something like:

    "Could not parse the included file ([ERROR HERE]).

alanz: We can do one (or both) of

1. Generate a diagnostic
2. Send a `showMessage`

We might not be able to attach the diagnostic to a file in the IDE, depending on what precisely the problem is.  But I think they should show up somewhere anyway, I am pretty sure they do in vscode.

robertoaloi: In case of issues within the hrl file, we already override the location of the original diagnostic message and display the message in the inclusion line:

https://github.com/erlang-ls/erlang_ls/blob/master/src/els_compiler_diagnostics.erl#L126

We could do the same with this. Thoughts?

alanz: All we have here is the Uri we are trying to parse, and we know that it failed, because the file does not exist, we are not searching in the right location, etc.

This should not actually happen, but if it does it indicates a misconfiguration of the project.  So I think a `window/ShowMessage` of type `ERROR` should be sent.

robertoaloi:
Closing as a non issue. See the discussion in #405 .

PR #420, opened 2020-01-31
Implement functionality for hot code swapping

onno-vos-dev:
### Description

One of the features that I miss most when coming from EDTS is the ability of hot code reloading my changes.
This PR introduces this ability and should help in getting users of EDTS switched over to erlang_ls.

There are a few things in this PR that I'm not a huge fan of and I'm happy to discuss:
- I have not found a nice way of being able to test this functionality in a test SUITE.
  I have exported the function that is called using `ifdef(TEST)` (which I personally generally dislike),
  after trying several ways this seemed like the least intrusive way of testing this.
  If you'd prefer another approach, please let me know.

- Ideally I'd like to provide the Mod and Function that should be used to recompile as part of a config as well.
  This to aid any users that have implemented some custom way of compiling their code.
  I'd like some input if you believe this is a decent idea.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,8 +431,12 @@ The following customizations are possible:
 | apps\_dirs             | List of directories containing project applications. It supports wildcards.                          |
 | include\_dirs          | List of directories provided to the compiler as include dirs. It supports wildcards.                 |
 | otp\_apps\_exclude     | List of OTP applications that will not be indexed (default: megaco, diameter, snmp, wx)              |
-| code\_reload\_enabled  | Whether or not an rpc call should be mode to a remote node to compile and reload a module            |
-| code\_reload\_node     | If code\_reload\_enabled is set to true, the node to be called should be provided. Example: "els@xx" |
+| code\_reload           | Whether or not an rpc call should be made to a remote node to compile and reload a module            |
+
+The `code_reload` takes a list of the following options:
+| Parameter              | Description                                                                                          |
+|------------------------|------------------------------------------------------------------------------------------------------|
+| node                   | The node to be called for code reloading. Example erlang_ls@hostname                                 |
 
 ## Troubleshooting
 
--- a/README.md
+++ b/README.md
@@ -423,14 +423,16 @@ The file format is `yaml`.
 
 The following customizations are possible:
 
-| Parameter          | Description                                                                                          |
-|--------------------|------------------------------------------------------------------------------------------------------|
-| otp\_path          | Path to the OTP installation                                                                         |
-| plt\_path          | Path to the dialyzer PLT file. When none is provided the dialyzer diagnostics will not be available. |
-| deps\_dirs         | List of directories containing dependencies. It supports wildcards.                                  |
-| apps\_dirs         | List of directories containing project applications. It supports wildcards.                          |
-| include\_dirs      | List of directories provided to the compiler as include dirs. It supports wildcards.                 |
-| otp\_apps\_exclude | List of OTP applications that will not be indexed (default: megaco, diameter, snmp, wx)              |
+| Parameter              | Description                                                                                          |
+|------------------------|------------------------------------------------------------------------------------------------------|
+| otp\_path              | Path to the OTP installation                                                                         |
+| plt\_path              | Path to the dialyzer PLT file. When none is provided the dialyzer diagnostics will not be available. |
+| deps\_dirs             | List of directories containing dependencies. It supports wildcards.                                  |
+| apps\_dirs             | List of directories containing project applications. It supports wildcards.                          |
+| include\_dirs          | List of directories provided to the compiler as include dirs. It supports wildcards.                 |
+| otp\_apps\_exclude     | List of OTP applications that will not be indexed (default: megaco, diameter, snmp, wx)              |
+| code\_reload\_enabled  | Whether or not an rpc call should be mode to a remote node to compile and reload a module            |
+| code\_reload\_node     | If code\_reload\_enabled is set to true, the node to be called should be provided. Example: "els@xx" |
 
 ## Troubleshooting
 
--- a/src/els_text_synchronization.erl
+++ b/src/els_text_synchronization.erl
@@ -8,6 +8,9 @@
         ]).
 
 -export([ generate_diagnostics/1 ]).
+-ifdef(TEST).
+-export([ maybe_compile_and_load/2 ]).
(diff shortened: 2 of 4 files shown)


michalmuskala:
Does it make sense to enable this feature without proving the name of the node? If it doesn't (and that would be my intuition), maybe it would be possible to get away with just one option?

onno-vos-dev:
> Does it make sense to enable this feature without proving the name of the node? If it doesn't (and that would be my intuition), would it be possible to get away with just one option?

Good point and no it does not make sense. The only thing is that it allows for a slightly cleaner toggle in `els_text_synchronization:maybe_compile_and_load/2` rather than casing over a `{'', _}`, `{Node, []}` and `_`. I'm open for changing, I don't have any strong opinions about it and totally see your point :)

README.md
```diff
@@ -423,14 +423,16 @@ The file format is `yaml`.
...
+| Parameter              | Description                                                                                          |
+|------------------------|------------------------------------------------------------------------------------------------------|
+| otp\_path              | Path to the OTP installation                                                                         |
+| plt\_path              | Path to the dialyzer PLT file. When none is provided the dialyzer diagnostics will not be available. |
+| deps\_dirs             | List of directories containing dependencies. It supports wildcards.                                  |
+| apps\_dirs             | List of directories containing project applications. It supports wildcards.                          |
+| include\_dirs          | List of directories provided to the compiler as include dirs. It supports wildcards.                 |
+| otp\_apps\_exclude     | List of OTP applications that will not be indexed (default: megaco, diameter, snmp, wx)              |
+| code\_reload\_enabled  | Whether or not an rpc call should be mode to a remote node to compile and reload a module            |
```
robertoaloi: Nit: mode -> made

README.md
```diff
@@ -423,14 +423,16 @@ The file format is `yaml`.
...
+| code\_reload\_node     | If code\_reload\_enabled is set to true, the node to be called should be provided. Example: "els@xx" |
```
robertoaloi: What about using a slightly different config format? Something like:

```
code_reload:
  node: nodename
```

This would allow us to enable/disable the feature with a single step and it would make it simpler to add extra options in the future (eg cookie, callback, etc)

src/els_text_synchronization.erl
```diff
@@ -42,4 +45,38 @@ generate_diagnostics(Uri) ->
   Params1  = #{ uri => Uri
               , diagnostics => CDiagnostics ++ DDiagnostics ++ EDiagnostics
               },
+  maybe_compile_and_load(Uri, CDiagnostics),
   els_server:send_notification(Method, Params1).
+
+-spec maybe_compile_and_load(uri(), [diagnostic()]) -> ok.
+maybe_compile_and_load(Uri, CDiagnostics) ->
```
robertoaloi: You can have a first clause where you match `CDiagnostics` with a `[]` to avoid the tuple in the case.

src/els_text_synchronization.erl
```diff
@@ -42,4 +45,38 @@ generate_diagnostics(Uri) ->
...
+      Module = els_uri:module(Uri),
+      Node = els_config:get(code_reload_node),
+      handle_rpc_result(rpc:call(Node, c, c, [Module]), Module);
+    _ ->
+      ok
+  end.
+
+-spec handle_rpc_result(term() | {badrpc, term()}, atom()) -> ok.
+handle_rpc_result({ok, Module}, _) ->
+  Msg = io_lib:format("Hot code swap success for: ~s", [Module]),
```
robertoaloi: Names are a bit inconsistent and could create confusion in the user:

* code_reload
* Hot code swap
* [HOT CODE RELOAD]

I would uniform them to `code_reload`.

src/els_text_synchronization.erl
```diff
@@ -42,4 +45,38 @@ generate_diagnostics(Uri) ->
...
+  els_server:send_notification(<<"window/showMessage">>,
+                               #{ type => ?MESSAGE_TYPE_INFO,
+                                  message => list_to_binary(Msg)
+                                });
+handle_rpc_result({badrpc, nodedown} = Err, Module) ->
+  Msg = io_lib:format("Hot code swap failed for: ~s with: ~w", [Module, Err]),
+  els_server:send_notification(<<"window/showMessage">>,
+                               #{ type => ?MESSAGE_TYPE_INFO,
```
robertoaloi: I would say we should use an error message type here.

src/els_text_synchronization.erl
```diff
@@ -42,4 +45,38 @@ generate_diagnostics(Uri) ->
...
+                                  message => list_to_binary(Msg)
+                                });
+handle_rpc_result(Err, Module) ->
```
robertoaloi: Why the double clause? I would just have a single one. In case of error, shows the message. No need to specially handle the nodedown.

src/els_text_synchronization.erl
```diff
@@ -8,6 +8,9 @@
         ]).
 
 -export([ generate_diagnostics/1 ]).
+-ifdef(TEST).
```
robertoaloi: I prefer not to use this kind of macros. There is no problem exporting this function.

robertoaloi:
Fantastic! Thank you!

robertoaloi:
As a note, this does not work for escripts (I was trying to setup for developing erlang_ls itself).
This is due to the fact most of the shell commands (including `c/1`) do not work for escripts. E.g.:

```
(erlang_ls_dbdbe4ef15aeafea62bdf724de90dbd270beee00@C02V80XBHTDD)2> c:c(els_protocol).
Recompiling /Users/robert.aloi/git/github/erlang-ls/erlang_ls/src/els_protocol.erl
{error,enotdir}
```

Or even:

```
(erlang_ls_dbdbe4ef15aeafea62bdf724de90dbd270beee00@C02V80XBHTDD)6> rr(els_protocol).
{error,beam_lib,
       {file_error,"/Users/robert.aloi/git/github/erlang-ls/erlang_ls/_build/debug/bin/erlang_ls/erlang_ls/ebin/els_protocol.beam",
                   enotdir}}
```

Not sure if these should be reported as bugs to OTP. Eventually, we may want to move away from shell commands and implement our own logic for compiling and reloading the module.

onno-vos-dev:
> Fantastic! Thank you!

Thank you for the fast feedback and merge!

PR #426, opened 2020-02-01
[#318] Introduce schema versioning

### LINKED ISSUE #318: Version the DB Schema
Currently, the server will skip the creation of the tables if a schema already exists. But the schema could change across server versions, so we should version it and remove the old cached entries if the current schema version is bigger than the one installed.

robertoaloi:
### Description
Fixes #318 .

### DIFF
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -20,6 +20,8 @@
                 , els_dt_signatures
                 ]).
 -define(TIMEOUT, infinity).
+-define(DB_SCHEMA_VSN, <<"1">>).
+-define(DB_SCHEMA_VSN_FILE, "DB_SCHEMA_VSN").
 
 %%==============================================================================
 %% Exported functions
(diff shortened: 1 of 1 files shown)


src/els_db.erl
```diff
@@ -134,3 +139,65 @@ epmd_path() ->
...
+-spec write_db_schema_vsn_file(string()) -> ok.
+write_db_schema_vsn_file(BaseDir) ->
+  ok = file:write_file(db_schema_vsn_file(BaseDir), ?DB_SCHEMA_VSN).
+
+-spec read_db_schema_vsn_file(string()) -> binary().
+read_db_schema_vsn_file(BaseDir) ->
+  case file:read_file(db_schema_vsn_file(BaseDir)) of
+    {ok, Vsn} ->
+      Vsn;
+    {error, Error} ->
```
alanz: For all currently existing databases this will fail.  Does that result in the db being deleted and recreated as if it was a schema mismatch?

robertoaloi: If you don't have the file (current situation), you will get a mismatch and the cache is deleted.

alanz:
Any reason this is not merged? @jfacorro ?

PR #432, opened 2020-02-10
[#408] Normalize the paths in the search directories

### LINKED ISSUE #408: Make file name references absolute when indexing
**Describe the bug**

If you have an `erlang_ls.config` with relative paths in it, e.g.

```yaml
include_dirs:
  - "include"
  - "../"
deps_dirs:
  - "../sibling_project"
```  
then references to functions from `../sibling_project` end up having a path with the `../` in it. Go to definition in vscode then works, but starting the language server in the sibling project does not work properly.

**To Reproduce**

As described above

**Expected behavior**

When files are opened, and the `erlang_ls` server starts up, all paths are absolute, so there is no confusion.

**Actual behavior**

As described above

**Context**
 - Editor used: vscode
 - LSP client used: erlang_ls

alanz:
Use absolute paths, normalised to remove duplicate path separators and embedded `.` or `..` parts.

This is particularly important on vscode when doing "go to definition", if the path has `..` in it, the file opens, but none of the indexing lookup works.

Fixes #408 .

### DIFF
--- a/src/els_utils.erl
+++ b/src/els_utils.erl
@@ -140,19 +140,20 @@ is_symlink(Path) ->
     {error, _} -> false
   end.
 
-%% Resolve paths recursively
+%% @doc Resolve paths recursively
 
 -spec resolve_path([string()], path(), boolean()) -> [string()].
 resolve_path(PathSpec, RootPath, Recursive) ->
   Path  = filename:join(PathSpec),
   Paths = filelib:wildcard(Path),
+
   case Recursive of
     true  ->
-      lists:append([ [P | subdirs(P)]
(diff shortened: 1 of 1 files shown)


src/els_utils.erl
```diff
@@ -194,3 +195,46 @@ contains_symlink(Path, RootPath) ->
       Parent = filename:join(ParentParts),
       is_symlink(Parent) orelse contains_symlink(Parent, RootPath)
   end.
+
+%%==============================================================================
+%% This section excerpted from the rebar3 sources, rebar_dir.els
```
robertoaloi: Nit: you probably meant `rebar_dir.erl`

PR #433, opened 2020-02-10
[#406] Add support for custom behaviours

### LINKED ISSUE #406: Compiler diagnostics do not pick up behaviour defined in a project

robertoaloi:
### Description

Fixes #406 .

Notice that the Dialyzer message about "callback info missing" needs to be addressed separately.

### DIFF
--- a/priv/code_navigation/src/diagnostics_behaviour.erl
+++ b/priv/code_navigation/src/diagnostics_behaviour.erl
@@ -1,0 +1,4 @@
+-module(diagnostics_behaviour).
+
+-callback one() -> ok.
+-callback two() -> ok.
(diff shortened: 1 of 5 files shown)


quantumpotato:
Hello, can I use a custom behavior just in the erlang shell? I am trying running `erl` in the same folder as the behaviour module + the module which uses it but it is not working.
Do I need to specify a path somehow to all behaviors used?

Issue #418, opened 2020-01-30
erlang_ls hangs when multiple editors want to use the same project


dszoboszlay:
**Describe the bug**
Each editor instance wants to run its own `erlang_ls` server for every project it opens. The project determines the path to the Mnesia DB and the node name of the `erlang_ls` server. This means that when multiple editor instances want to use the same project, they both attempt to start an `erlang_ls` server with the same DB path and node name.

`els_db:install/2` attempts to start the distribution protocol, but does not pattern match on the return value of `net_kernel:start/1`. Only the first editor instance's `erlang_ls` server will be able to take the desired node name, for other editor instances `net_kernel:start/1` would fail and the node will continue in nondistributed mode, using the `nonode@nohost` name.

Then `els_db:ensure_db/0` will verify the DB schema exists and starts Mnesia. This works, however Mnesia will believe that all tables exist only on a remote node (the one with the desired node name), not on the local node (`nonode@nohost`). Since it cannot connect to that remote node, it cannot load any of the tables either and `els_db:wait_for_tables/1` would block indefinitely.

**To Reproduce**
1. Open an arbitrary `.erl` file in Emacs.
2. Open the same `.erl` file (or a different one in the same application) in a new instance of Emacs.

**Expected behavior**
`erlang_ls` works in both Emacs instances.

**Actual behavior**
In the second Emacs instance `erlang_ls` hangs in _status:starting_.

**Context**
 - `erlang_ls` version (tag/sha): 5e754db
 - Editor used: Emacs
 - LSP client used: lsp-mode-20200115.950

robertoaloi:
Hi @dszoboszlay ! This is a known issue, but thanks for formalizing this in such a detailed manner. We'll get it fixed as soon as possible.

robertoaloi:
Reference: https://github.com/microsoft/language-server-protocol/issues/23

robertoaloi:
@dszoboszlay I can see two possibilities here:

1. Have a completely separate instance of the DB / server. This would require quite a bit of work and I'm not sure about the advantages.
2. Have both clients interacting with the same server. Not really how LSP is supposed to work at the moment (see referenced issue)

So, not sure about how to proceed with this one. Any reason why you cannot use `emacs-client` instead of `emacs`?

I am leaning towards just refusing the second client in a more obvious way, unless you have a better suggestion.

dszoboszlay:
[...]

PR #437, opened 2020-02-14
Add root uri basename to startup message.

alanz:
We can end up running multiple LSP servers for a given project. This makes it
clear where each one is running.


robertoaloi:
Running multiple servers for the same project is problematic. I think we should fix https://github.com/erlang-ls/erlang_ls/issues/418 which would make this change un-necessary.

alanz:
Bad terminology on my part. Running different servers for different project roots.  This is completely normal. e.g., right now I run vscode against a remote server with one erlang_ls instance running against the erlang_ls source, and another against the company source.

PR #441, opened 2020-02-17
Add include_dirs as additional option to dialyzer:run/1 in els_dialyzer_diagnostics

onno-vos-dev:
### Description

Today I tried to add `plt_path` to one of my projects but realized that dialyzer was not reporting any errors. The reason turned out to be that `dialyzer:run/1` was unable to operate on the src code since it was unable to find the includes that were required. 

Adding the `include_dirs` option solved the problem. 

Note:
I did not add a test for this since I was unable to find any dialyzer tests at all? Unless I'm mistaken, I suggest we add an issue for adding some tests for `els_dialyzer_diagnostics`


robertoaloi:
Thanks! \o/

PR #442, opened 2020-02-18
[#416] Add support for parse transforms

### LINKED ISSUE #416: Add support for parse transforms
In case parse transforms are used in a module, the compiler reports them as undefined.

robertoaloi:
### Description

Fixes #416.


robertoaloi:
I merged #444 and due to a mis-typed ticket number this got closed. Facepalm. Re-opening.

robertoaloi:
One potential problem with this is related to circular dependencies, which should properly be detected. But I suspect this issue is not so problematic in practice and can be tackled separately.

PR #445, opened 2020-02-19
[#440] global config

### LINKED ISSUE #440: Allow specifying alternative global config file path
At the moment the config path lookup code is

```erlang
-spec config_paths(path(), map()) -> [path()].
config_paths(RootPath, #{<<"erlang">> := #{<<"config_path">> := ConfigPath}}) ->
  [filename:join([RootPath, ConfigPath]) | default_config_paths(RootPath)];
config_paths(RootPath, _Config) ->
  default_config_paths(RootPath).
```

This means that if `erlang.config_path` is set, it is interpreted as being relative to the `RootPath`.

It would be useful to provide an alternative `erlang.global_config_path`, which specifies an absolute path for the config file, if it lives in a different location on the filesystem.

This would entail an additional option in the middle, as

```erlang
...
config_paths(RootPath, #{<<"erlang">> := #{<<"global_config_path">> := GlobalConfigPath}}) ->
  [ GlobalConfigPath | default_config_paths(RootPath)];
...
```

This is as an alternative to #395 

master as 769f1f4e044731c4213f0287fdbb80204cfaee40

alanz:
Provide the ability to set config file location via the Initialize request.

In javascript terms, this is expected to be

    erlang.config_path

The value can be absolute or relative, and can include the name of the file to
be loaded as well, otherwise is treated as a path.

Fixes #440 .

### DIFF
--- a/src/els_config.erl
+++ b/src/els_config.erl
@@ -149,8 +149,12 @@ handle_cast(_Msg, State) -> {noreply, State}.
 %%==============================================================================
 
 -spec config_paths(path(), map()) -> [path()].
-config_paths(RootPath, #{<<"erlang">> := #{<<"config_path">> := ConfigPath}}) ->
-  [filename:join([RootPath, ConfigPath]) | default_config_paths(RootPath)];
+config_paths( RootPath
+            , #{<<"erlang">> := #{<<"config_path">> := ConfigPath0}}) ->
+  ConfigPath = binary_to_list(ConfigPath0),
+  lists:append([ possible_config_paths(ConfigPath)
+               , possible_config_paths(filename:join([RootPath, ConfigPath]))
+               , default_config_paths(RootPath)]);
 config_paths(RootPath, _Config) ->
   default_config_paths(RootPath).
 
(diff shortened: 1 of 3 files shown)


src/els_config.erl
```diff
@@ -149,8 +149,12 @@ handle_cast(_Msg, State) -> {noreply, State}.
 %%==============================================================================
 
 -spec config_paths(path(), map()) -> [path()].
-config_paths(RootPath, #{<<"erlang">> := #{<<"config_path">> := ConfigPath}}) ->
-  [filename:join([RootPath, ConfigPath]) | default_config_paths(RootPath)];
+config_paths( RootPath
+            , #{<<"erlang">> := #{<<"config_path">> := ConfigPath0}}) ->
+  ConfigPath = binary_to_list(ConfigPath0),
+  lists:append([ possible_config_paths(ConfigPath)
+                , possible_config_paths(filename:join([RootPath, ConfigPath]))
```
jfacorro: Nit: alignment of commas with `[`. 😄

alanz: Need to get `erlfmt` integrated properly :)

PR #449, opened 2020-02-20
[#448] Ignore $/ notifications and requests

### LINKED ISSUE #448: Ignore $ notifications
According to the protocol:

```
Notification and requests ids starting with '$/' are messages which are protocol implementation
dependent and might not be implementable in all clients or servers. For example if the server 
implementation uses a single threaded synchronous programming language then there is little 
a server can do to react to a '$/cancelRequest'. If a server or client receives notifications or requests 
starting with '$/' it is free to ignore them if they are unknown.
```

Related: https://github.com/erlang-ls/vscode/issues/29

References: https://github.com/Microsoft/vscode-languageserver-node/issues/170

robertoaloi:
### Description

Fixes #448 .

### DIFF
--- a/elvis.config
+++ b/elvis.config
@@ -38,6 +38,7 @@
                                                                    , {right, "=/="}
                                                                    , {left , "=/="}
                                                                    ]}}
+                      , {elvis_style, function_naming_convention, #{ignore => [els_methods]}}
                       ]
          , ignore => [els_dodger]
          }
--- a/elvis.config
+++ b/elvis.config
@@ -38,7 +38,6 @@
                                                                    , {right, "=/="}
                                                                    , {left , "=/="}
                                                                    ]}}
-                      , {elvis_style, function_naming_convention, #{ignore => [els_methods]}}
                       ]
          , ignore => [els_dodger]
          }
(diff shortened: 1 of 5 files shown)


elvis.config
```diff
@@ -38,6 +38,7 @@
                                                                    , {right, "=/="}
                                                                    , {left , "=/="}
                                                                    ]}}
+                      , {elvis_style, function_naming_convention, #{ignore => [els_methods]}}
```
robertoaloi: @jfacorro Is there a way to allow a different regex only for a given module? In this case I would like to allow the `$` for the `els_methods` module...

onno-vos-dev: Not Juan, but I can answer this one. It is currently not possible. You can see the implementation here: https://github.com/inaka/elvis_core/blob/master/src/elvis_style.erl#L158

It's something that I've had on my personal projects list to fix 👍 I'll see if I can spend some time on the weekend to allow for a per module regex 👍

jfacorro: As @onno-vos-dev said, sadly there's not. 😞

jfacorro: The `els_method` module doesn't have any functions whose name starts iwth `$_` so I think this ignore could be removed, right?

robertoaloi: Yes, this line is a leftover from a previous commit where I had individual functions for the supported `$/` notifications.

PR #452, opened 2020-02-21
[#430] Expose logging level

### LINKED ISSUE #430: Allow debug logging via a command line flag
When trying to debug `erlang_ls` one has to recompile a new binary and reconfigure the LSP implementation to use this new binary to get debug log output.

A `--debug` flag that enables debug logging and debugging features in the normal binary.

None.

benrs:
### Description

Exposes lager's log level to the CLI.

Fixes #430.

### DIFF
--- a/src/erlang_ls.erl
+++ b/src/erlang_ls.erl
@@ -68,7 +68,8 @@ lager_config() ->
 
 -spec lager_handlers(string()) -> [any()].
 lager_handlers(LogRoot) ->
-  {ok, LoggingLevel} = application:get_env(?APP, log_level, ?DEFAULT_LOGGING_LEVEL),
+  {ok, LoggingLevel} =
+    application:get_env(?APP, log_level, ?DEFAULT_LOGGING_LEVEL),
   LogFile = filename:join([LogRoot, "debug.log"]),
   ok      = filelib:ensure_dir(LogFile),
   [ { lager_file_backend
--- a/src/erlang_ls.erl
+++ b/src/erlang_ls.erl
@@ -68,8 +68,7 @@ lager_config() ->
 
 -spec lager_handlers(string()) -> [any()].
 lager_handlers(LogRoot) ->
-  {ok, LoggingLevel} =
-    application:get_env(?APP, log_level, ?DEFAULT_LOGGING_LEVEL),
+  LoggingLevel = application:get_env(?APP, log_level, ?DEFAULT_LOGGING_LEVEL),
   LogFile = filename:join([LogRoot, "debug.log"]),
   ok      = filelib:ensure_dir(LogFile),
   [ { lager_file_backend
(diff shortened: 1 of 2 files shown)


src/erlang_ls.erl
```diff
@@ -64,11 +68,13 @@ lager_config() ->
 
 -spec lager_handlers(string()) -> [any()].
 lager_handlers(LogRoot) ->
+  {ok, LoggingLevel} =
```
robertoaloi: This will cause a crash. `application:get_env/3` returns the actual value, not a tuple. See:

http://erlang.org/doc/apps/kernel/application.html#get_env-3

benrs:
Thanks for catching that, I'll add unit tests in another PR.

src/erlang_ls.erl
```diff
@@ -64,11 +68,12 @@ lager_config() ->
 
 -spec lager_handlers(string()) -> [any()].
 lager_handlers(LogRoot) ->
+  LoggingLevel = application:get_env(?APP, log_level, ?DEFAULT_LOGGING_LEVEL),
   LogFile = filename:join([LogRoot, "debug.log"]),
```
robertoaloi: Shouldn't the filename also depend on the level now? Sorry I didn't catch this earlier.

benrs: Might make sense to keep it all in one place and just rename the file. Maybe just rename it to `server.log`?

jfacorro: IMO keeping the name of the logging level makes it less surprising since this is in general what `lager` does.

benrs: I think one problem with the logging level as a file name is that it may cause issues on Window machines, as far as I know we can't have `<` or `>` in file names which might occur with the way we're setting this up.

benrs: Made the suggested change. Made the following mappings in the filename to avoid the above:

```
">" -> "gt"
"<" -> "lt"
"=" -> "eq"
"!" -> "not"
```

Wasn't necessary to switch "=" and "!" but felt it made more sense to do so.

jfacorro: I didn't realize before the idea was to support the full lager [syslog style comparison flags](https://github.com/erlang-lager/lager#syslog-style-loglevel-comparison-flags). In that case I would definitely go for just using `server.log`. Sorry for generating the extra work 😞 🙏 .

robertoaloi: TIL. I didn't realize that either O_o

benrs: Haha, yeah that was the general idea since we're directly exposing lager's log level. Any reason not to?

robertoaloi: After learning about the syslog style comparison flags, it feels like @benrs original idea of having a simple `server.log` is actually the way to go. Also, we should update the README to reflect these changes. I will merge this PR as it is and PR a follow-up.

PR #469, opened 2020-02-24
468 async dialyzer and elvis checks

### LINKED ISSUE #468: Exclude directories from Dialyzer checking
Dialyzer as we all know is somewhat time consuming. Especially for large(r) files or files with lots of references, running dialyzer using source code only, can be somewhat time consuming. For certain files, I may however not be interested in seeing the dialyzer results in my editor.

I'd like to be able to pass a list of folders (with wildcards) in the `erlang_ls.config` that would allow skipping dialyzer for certain directories. 

N/A

(Happy to work on this feature but would appreciate a :+1: or :-1: )

onno-vos-dev:
### Description

After raising #468 I found that with a minor rewrite we could work around the slowness of Dialyzer by performing Elvis and Dialyzer checks asynchronously. The (opinionated) view here being that compiler diagnostics and the ability to use code_reload and run your tests quickly are more important than seeing Dialyzer and Elvis warnings in your editor. 

They still show up but with a slight delay. Basically the flow went from:
- compile -> dialyzer -> elvis -> code_reload -> notify editor

to:
- compile -> code_reload -> notify editor -> dialyzer -> elvis -> notify editor 

For the end-user very little changed other than that Dialyzer and Elvis warnings come with a slight delay after the user has been notified of compiler warnings / code_reload result. 

Fixes #468

### DIFF
--- a/src/els_text_synchronization.erl
+++ b/src/els_text_synchronization.erl
@@ -8,6 +8,7 @@
         ]).
 
 -export([ generate_diagnostics/1
+        , diagnostics/3
         , maybe_compile_and_load/2
         ]).
 
@@ -37,15 +38,26 @@ did_close(_Params) -> ok.
 
 -spec generate_diagnostics(uri()) -> ok.
 generate_diagnostics(Uri) ->
-  CDiagnostics = els_compiler_diagnostics:diagnostics(Uri),
-  DDiagnostics = els_dialyzer_diagnostics:diagnostics(Uri),
-  EDiagnostics = els_elvis_diagnostics:diagnostics(Uri),
+  Sync        = [els_compiler_diagnostics],
+  Diagnostics = diagnostics(Uri, Sync, []),
+  maybe_compile_and_load(Uri, Diagnostics),
+
+  Async = [ els_dialyzer_diagnostics
+          , els_elvis_diagnostics
+          ],
+  erlang:spawn(?MODULE, diagnostics, [Uri, Async, Diagnostics]),
+  ok.
+
+-spec diagnostics(uri(), [module()], [diagnostic()]) -> [diagnostic()].
+diagnostics(Uri, Modules, Previous) ->
+  Diagnostics = [apply(M, diagnostics, [Uri]) || M <- Modules],
+  AllDiagnostics = lists:append([Previous | Diagnostics]),
   Method = <<"textDocument/publishDiagnostics">>,
-  Params1  = #{ uri => Uri
-              , diagnostics => CDiagnostics ++ DDiagnostics ++ EDiagnostics
-              },
-  maybe_compile_and_load(Uri, CDiagnostics),
-  els_server:send_notification(Method, Params1).
+  Params  = #{ uri         => Uri
+             , diagnostics => AllDiagnostics
+             },
+  els_server:send_notification(Method, Params),
+  AllDiagnostics.
 
 -spec maybe_compile_and_load(uri(), [diagnostic()]) -> ok.
 maybe_compile_and_load(Uri, [] = _CDiagnostics) ->
--- a/test/els_diagnostics_SUITE.erl
+++ b/test/els_diagnostics_SUITE.erl
@@ -203,6 +203,15 @@ elvis(Config) ->
       file:set_cwd(RootPath),
       Uri = ?config(elvis_diagnostics_uri, Config),
       ok = els_client:did_save(Uri),
+      %% Only the compiler diagnostics are in this notification
+      {CMethod, CParams} = wait_for_notification(),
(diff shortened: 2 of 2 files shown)


src/els_text_synchronization.erl
```diff
@@ -37,15 +37,33 @@ did_close(_Params) -> ok.
 
 -spec generate_diagnostics(uri()) -> ok.
 generate_diagnostics(Uri) ->
+  CDiagnostics = sync_diagnostics(Uri),
+  async_diagnostics(Uri, CDiagnostics).
```
jfacorro: Why provide the compiler diagnostics here as well when they are already being sent as a notification in the synchronous diagnostics?

onno-vos-dev: We have to pass the compiler diagnostics down or else we loose them once the Dialyzer and Elvis diagnostics are reported. 

Take the following scenario as an example:
Given a module with a compiler error and an Elvis warning
When the diagnostics are run
Then the compiler error disappears once the Elvis warning comes in

By passing down the compiler error to the spawned process, the compiler errors are reported together with any Dialyzer and Elvis warnings and therefor the "compiler warnings are overwritten" when those notifications arrive. 

(I believe this also answers the question above in regards to generalizing this unless I misunderstood you?)

jfacorro: The question above is more about having a more generic approach to the diagnostics. 

Something like this:
```
-spec generate_diagnostics(uri()) -> ok.
generate_diagnostics(Uri) ->
  Sync        = [els_compiler_diagnostics],
  Diagnostics = diagnostics(Uri, Sync, []),
  maybe_compile_and_load(Uri, Diagnostics),

  Async = [ els_dialyzer_diagnostics
          , els_elvis_diagnostics
          ],
  erlang:spawn(?MODULE, diagnostics, [Uri, Async, Diagnostics]),
  ok.

-spec diagnostics(uri(), [module()], [diagnostic()]) -> [diagnostic()].
diagnostics(Uri, Modules, Previous) ->
  Diagnostics = [apply(M, diagnostics, [Uri]) || M <- Modules],
  AllDiagnostics = lists:append([Previous | Diagnostics]),
  Method = <<"textDocument/publishDiagnostics">>,
  Params  = #{ uri         => Uri
             , diagnostics => AllDiagnostics
             },
  els_server:send_notification(Method, Params),
  AllDiagnostics.
```

onno-vos-dev: Ok cool, I see what you mean 👍 I'll fix it up and follow it with a squash to tidy things up.

I've noticed a significant improvement in my work flow with code_reload which is a lot more responsive now. And with some more fixes that we'll have to do to get Dialyzer in a happy state (it's missing a lot of type spec errors right now) Dialyzer will definitely become the slowest one in the diagnostics flow. 

So we might eventually even want to consider sending a message for each step and do the Elvis step before Dialyzer (I'm gonna make an assumption here and say for many this would be more important to get quick feedback on or perhaps it's just my personal view) but I'm not very familiar with the overhead and impact of sending multiple messages to the IDE to handle. Any thoughts on this?

onno-vos-dev:
[commit 
8907f2f](https://github.com/erlang-ls/erlang_ls/pull/469/commits/8907f2fd7ab13cf6f28ef9fd9f830e61c9f217e8) is the one that applied the suggestion from Juan. Followed it up with a squash to tidy things up

src/els_text_synchronization.erl
jfacorro: > So we might eventually even want to consider sending a message for each step

This could definitely be the way to go in the future, although I quite like the fast feedback vs. the slow feedback we have now.

>  ... but I'm not very familiar with the overhead and impact of sending multiple messages to the IDE to handle.

I guess this depends on how each IDE handles these messages. My experience with Emacs in this regard is that if there are too many (over a 100) errors or warnings to be shown, then interactions become slow.

alanz: In the haskell language server, we cache the current diagnostics set for a file on the server, and update it based in `(Uri, diagnostics source)`. Whenever it changes we send to the client.

Also, we sort the diagnostics in order of occurrence in the file, and limit it to 100 max (a workspace config).

Further, if the compiler reports errors ( not warnings), we suppress diagnostics from other sources.

onno-vos-dev: @alanz Yeah we just discussed a similar approach but using a gen_server instead :) I'll start the work on this (probably on the weekend). This should create a much cleaner approach and especially for Dialyzer diagnostics, this should result in a smoother experience to the user :) (With some more changes that I'm working on, Dialyzer will likely take even longer...)

In that work I'll also include some tests for Dialyzer since all of this work isn't tested in a test suite other than "verifying that it works in a few different projects before creating a PR".

PR #473, opened 2020-02-25
[#471]: Ensure behaviour info can be found by Dialyzer by adding dependencies to files opt

### LINKED ISSUE #471: Callback info about the foo behaviour is not available
**Describe the bug**
`bar.erl:5: Callback info about the foo behaviour is not available`

**To Reproduce**
behaviour:
```erlang
-module(foo).

-export([ lorem/0 ]).

-define(CB_MOD, bar).

-callback lorem() -> atom().

lorem() -> ?CB_MOD:?FUNCTION_NAME().
```
implementation:
```erlang
-module(bar).

-export([ lorem/0 ]).

-behaviour(foo).

lorem() -> ipsum.
```

**Expected behavior**
behaviour info is available

**Actual behavior**
A clear and concise description of what happens instead.

**Context**
 - `erlang_ls` version (tag/sha): d7fbda7 (master as of creating this issue)
 - Editor used: GNU Emacs 26.3
 - LSP client used: `lsp-mode 20191219.1032`

onno-vos-dev:
Fixes #471

### DIFF
--- a/src/els_dialyzer_diagnostics.erl
+++ b/src/els_dialyzer_diagnostics.erl
@@ -29,7 +29,8 @@ diagnostics(Uri) ->
   case els_config:get(plt_path) of
     undefined -> [];
     DialyzerPltPath ->
-      WS = try dialyzer:run([ {files, [binary_to_list(Path)]}
+      Dependencies = deps_paths(els_diagnostics_utils:dependencies(Uri)),
+      WS = try dialyzer:run([ {files, [binary_to_list(Path) | Dependencies]}
                             , {from, src_code}
                             , {include_dirs, els_config:get(include_paths)}
                             , {plts, [DialyzerPltPath]}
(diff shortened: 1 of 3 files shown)


src/els_dialyzer_diagnostics.erl
```diff
@@ -29,7 +29,8 @@ diagnostics(Uri) ->
   case els_config:get(plt_path) of
     undefined -> [];
     DialyzerPltPath ->
-      WS = try dialyzer:run([ {files, [binary_to_list(Path)]}
+      Dependencies = deps_paths(els_diagnostics_utils:dependencies(Uri)),
+      WS = try dialyzer:run([ {files, [binary_to_list(Path) | Dependencies]}
```
jfacorro: Won't this end up analyzing all of the dependencies and also returning all the issues with all analyzed files? I don't think this is what we want, do we?

onno-vos-dev: It only looks for the dependencies of the particular module being analyzed. Surely this is more than only the one module but it does not appear to be much slower than it was before (disclaimer: no measurements just "feeling"). 

(Dialyzer runs done from Erlang LS do still misses some errors though but that's a different problem than the one I aim to fix here)

jfacorro: Right, dependencies in this case are only the behaviours and/or parse transforms used.

robertoaloi:
Thanks!

PR #467, opened 2020-02-24
[#465] Find record references

jfacorro:
### Description

Implements the possibility to find references for a record. 

Fixes #465

PR #474, opened 2020-02-25
[#467] Bump schema version, to ensure record references are indexed

robertoaloi:
### Description

Follow up for #467

### DIFF
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -20,7 +20,7 @@
                 , els_dt_signatures
                 ]).
 -define(TIMEOUT, infinity).
--define(DB_SCHEMA_VSN, <<"2">>).
+-define(DB_SCHEMA_VSN, <<"3">>).
 -define(DB_SCHEMA_VSN_FILE, "DB_SCHEMA_VSN").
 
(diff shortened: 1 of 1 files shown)


src/els_db.erl
```diff
@@ -20,7 +20,7 @@
                 , els_dt_signatures
                 ]).
 -define(TIMEOUT, infinity).
--define(DB_SCHEMA_VSN, <<"2">>).
+-define(DB_SCHEMA_VSN, <<"3">>).
```
jfacorro: Maybe we should use something in the line of semantic versioning for this. It makes more sense to me to bump a minor or a patch than bump the whole number. What do you think?

robertoaloi: The current check verifies exact equality and I don't see the advantage of having semver for the schema version. I am actually tempted by clearing the cache at every upgrade (so that we don't need to remember to bump this number), but that's not ideal either. My suggestion is to live with this for a while and see how things evolve.

alanz: Perhaps we should use a date for this, so it gives some idea of when it changed?

PR #494, opened 2020-02-28
[#421] Find references for macros

### LINKED ISSUE #421: Support find references on macros
It is already possible to go to the definition of a macro, but it would be also useful to find all references of a macro (when the point is over the name of the macro in the `-define(...).` attribute or any use of the macro later). This way one could easily check e.g. whether a macro is unused or which parts of the code base would a change in the macro definition affect.

When the point is over the name of the macro in the `-define(...).` attribute or any use of the macro later find references shall list all places where the macro is used, similar to how it works with functions.

Grep

**Additional context**

jfacorro:
### Description

Find references for macros.

Fixes #421.

### DIFF
--- a/src/els_indexer.erl
+++ b/src/els_indexer.erl
@@ -91,8 +91,9 @@ do_index(#{uri := Uri, id := Id, kind := Kind} = Document) ->
   %% References
   POIs  = els_dt_document:pois(Document, [ application
                                          , implicit_fun
-                                         , record_expr
+                                         , macro
                                          , record_access
+                                         , record_expr
                                          ]),
   ok = els_dt_references:delete_by_uri(Uri),
   [register_reference(Uri, POI) || POI <- POIs],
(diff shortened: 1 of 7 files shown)


src/els_indexer.erl
```diff
@@ -208,23 +209,21 @@ index_document(Uri, Text, sync) ->
...
-register_reference( Uri
-                  , #{kind := record_access, id := {Name, _}, range := Range}
-                  ) ->
-  els_dt_references:insert(#{ id    => {record, Name}
+register_reference(Uri, #{kind := macro, id := MacroName, range := Range}) ->
+  els_dt_references:insert(#{ id    => {macro, MacroName}
                             , uri   => Uri
                             , range => Range
                             });
 register_reference(Uri, #{id := {F, A}} = POI) ->
```
robertoaloi: Somewhat unrelated, but shouldn't we match on the `kind` here as well?

src/els_range.erl
```diff
@@ -17,99 +17,99 @@ compare( #{from := FromA, to := ToA}
...
   From = {Line, Column},
   To = From,
   #{ from => From, to => To };
-range({Line, _Column}, parse_transform, _Define) ->
+range({Line, _Column}, parse_transform, _Define, _Data) ->
   From = {Line, 1},
   To = From,
   #{ from => From, to => To };
-range(Pos, record_access, {Record, Field}) ->
+range(Pos, record_access, Record, Field) ->
```
robertoaloi: Why not to include the fieldname in the id, instead of passing the `Data` everywhere and ignoring it in all places except this one? Actually, this change does not sounds necessary for macros at all. Maybe revert?

src/els_indexer.erl
jfacorro: Yup, but we weren't doing it and I had already introduced other unrelated changes.

src/els_range.erl
jfacorro: The field name doesn't really belong in the `id` though. This makes the handling in the rest of the code much cleaner, and we are able to handle less special cases. I think we should do the same for POIs that have `id`s in the format `{F, A}` vs `{M, F, A}`, unifying them to `{M, F, A}`, which we end up doing anyway when adding them to the references table.

jfacorro: If you check the changes in this PR we were discarding the value for `Field` except in `els_range:range/3`.

robertoaloi: Fair enough.

PR #501, opened 2020-03-02
more config dirs

dgud:
Adds `src_dirs` config options for application/projects that have other source files in other directories than below `src`.

Also fixes bug with `include_dirs` which not was not included in search_dirs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,10 @@ all:
 	@ echo "Building escript..."
 	@ rebar3 escriptize
 
+
+debug:
+	rebar3 as debug escriptize
+
 ci:
 	rebar3 do compile, ct, proper --cover --constraint_tries 100, dialyzer, xref, cover, edoc
 
--- a/src/els_config.erl
+++ b/src/els_config.erl
@@ -80,6 +80,7 @@ do_initialize(RootUri, Capabilities, Config) ->
   OtpPath         = maps:get("otp_path", Config, code:root_dir()),
   DepsDirs        = maps:get("deps_dirs", Config, []),
   AppsDirs        = maps:get("apps_dirs", Config, ["."]),
+  SrcDirs         = maps:get("src_dirs", Config, []),
   IncludeDirs     = maps:get("include_dirs", Config, ["include"]),
   Macros          = maps:get("macros", Config, []),
   DialyzerPltPath = maps:get("plt_path", Config, undefined),
@@ -98,6 +99,7 @@ do_initialize(RootUri, Capabilities, Config) ->
   ok = set(otp_path       , OtpPath),
   ok = set(deps_dirs      , DepsDirs),
   ok = set(apps_dirs      , AppsDirs),
+  ok = set(src_dirs       , SrcDirs),
   ok = set(include_dirs   , IncludeDirs),
   ok = set(macros         , Macros),
   ok = set(plt_path       , DialyzerPltPath),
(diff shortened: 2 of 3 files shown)


robertoaloi:
Hi @dgud  and thanks for your contribution!

If I understand you correctly, you would like to add extra directories to the `src`, `include` and `test` ones which are currently used, which sounds reasonable to me.

Not sure we should do it at this level, though. Would it make more sense to modify the current  `project_paths` function to not hard-code the list of directories, but to allow the user to override that list via the config?

https://github.com/erlang-ls/erlang_ls/blob/026a9a701b5ef0a240c824af5c3de803beb6c0bf/src/els_config.erl#L196

Also, if we end up adding an option, we should ensure it's documented in the README and covered by tests.

Finally, can I ask you to decouple the `include_dirs` fix in a different PR? That sounds like a good fix which we can quickly introduce.

dgud:
A different option for 'project_paths' should work as well, your call.
This fix solves https://github.com/dgud/wings/ where `e3d`, `tools`, and especially `plugins_src` nested dir is an awkward dir structure that causes problems for me.
But it's a 20 years old projekt way before `rebar` forced dir structures :-)

Will split the PR later busy right now.

dgud:
Made a separate bugfix PR and removed that commit from this PR.

dgud:
I believe that the issue here is that not everything is searched below the rootpath.

For example I set the root dir for my otp project to be ~USER/git/otp I expect all otp apps
to be indexed when I'm working with otp stuff.

But that doesn't work since `erlang_ls` will only search after: ` ~USER/git/otp/[src|test|include]`
directories which doesn't exist and then will index the installed erlang release.
Then jump to function definition will jump to the wrong definition, e.g. in the installed released instead of the one in `~USER/git/lib/`

Also deps in a rebar project is not indexed, shouldn't `_build/lib/default/*` be included?
As far as I understand the deps work in `erlang_ls` itself by accident since it is in `erlang_ls`
script path.
But for another project it doesn't work, for example in the rebar3 project you can't see doc or jump
to definition of `getopt:tokenize/1`.

I'm starting to think that the default algorithm should be turned around, 
`erlang_ls` should by default search everything below `rootpath` and
mayby by default exclude some directories?

We need a way to exclude directories anyway for example in otp's repository all: 
`lib/*/test/*_data/` directories should not be indexed by default.
Which can not be done today.

Opinions?

robertoaloi:
@dgud Can you share the `erlang_ls.config` files you are using to configure `wings`, `otp` and `erlang_ls`? I suspect some of the paths are not correctly configured there.

Regarding rebar3 projects, right now we are not inferring information automatically from the `rebar.config`, but that will happen soon.

robertoaloi:
And yes, it would be good to be able to exclude directories from the indexing. Could you please raise an issue about that?

dgud:
For otp I don't have one, I just set the top of otp as my rootdir when emacs asked.

For wings I need the this patch to set the src_dirs:

erlang_ls.config:
```
src_dirs:
  - "e3d"
  - "plugins_src"

include_dirs:
  - "intl_tools"

deps_dirs:
  - "_deps/cl"

macros:
  - name: "wings_version"
    value: "\"lsp-compiled\""
```

But the issue here is the default behavior, as it is today a lot of projects need to create a
`erlang_ls.config` file to make it work, when it could have worked out of the box. 

So my question is why is `erlang_ls` only looking in `src test include` sub-dirs.
What does it gain by only looking in those dirs?

robertoaloi:
> But the issue here is the default behavior, as it is today a lot of projects need to create a
> `erlang_ls.config` file to make it work, when it could have worked out of the box.

True, and the plan is to address this as part of #6 . Actually, one of our driving principles about configuring directories should be that these arguments (e.g. `apps_dirs`, `src_dirs`, etc) should converge with the ones that can be specified in `rebar3`. That will make implementing #6 smooth.

> So my question is why is `erlang_ls` only looking in `src test include` sub-dirs.
> What does it gain by only looking in those dirs?

In the very old days we used this to limit the amount of files scanned, but nowadays indexing works in a very different way and this is probably not needed. So, happy to remove that limitation and that's what I was somehow suggesting when I proposed to make that list configurable. I guess we can also make `*` the default.

dgud:
I'm playing around a bit trying to understand how to solve the issues.
Ignore for now, want to keep the PR for discussion.

jfacorro:
Closing this PR since it's supeerseded by #569. Feel free to re-open if you disagree.

PR #502, opened 2020-03-02
[#413] Avoid negative columns

### LINKED ISSUE #413: Dialyzer and Elvis diagnostics are not visible in VS Code
Even if they seem to be correctly sent by the server,

jfacorro:
### Description

The diagnostics returned for dialyzer contained a negative column number which prevented any other diagnostics to be reported in VS Code.

Fixes #413 .


jfacorro:
I'm going to merge this PR since it is small and the change can be easily reverted. I've tested this in VS Code and in Emacs.

PR #504, opened 2020-03-02
Fix readme for code reload

onno-vos-dev:
### Description

I'm honestly not exactly sure how the `code_reload` should be structured from a `yaml`-purist perspective but plenty of yaml-examples I find use the following standard. 

Currently it's confusing since it's not well documented and most users would assume which is "wrong" according to how the code is written :cry: 
```yaml
code_reload:
  - node: foo@example
```

Since the test uses meck, the test didn't catch this which I'll see how to resolve in a PR that I'm working now that follows up on: https://github.com/erlang-ls/erlang_ls/pull/469 (cleaning up the code now and finalizing some Dialyzer tests)

### DIFF
--- a/README.md
+++ b/README.md
@@ -418,6 +418,8 @@ macros:
   - name: DEFINED_WITH_VALUE
     value: 42
   - name: DEFINED_WITHOUT_VALUE
+code_reload:
+  node: node@example
 ```
 
 The file format is `yaml`.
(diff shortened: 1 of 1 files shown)


seriyps:
There is this line below in README:

> The `code_reload` takes a list of the following options:

not sure if word "list" is appropriate

onno-vos-dev:
> There is this line below in README:
> 
> > The `code_reload` takes a list of the following options:
> 
> not sure if word "list" is appropriate

updated now :)

michalmuskala:
I would have a different question in that context - why YAML instead of Erlang terms like other config files in the context of using Erlang?

PR #505, opened 2020-03-02
[Request For Feedback] [#468] Move diagnostics to a gen_server based solution

onno-vos-dev:
### Description
This is an early review where I'd like to get some feedback on to hopefully polish things up. I'm opening this PR as I have a few questions I'd like to get input on. 

This PR addresses the discussion from: https://github.com/erlang-ls/erlang_ls/pull/469#pullrequestreview-365627511 

There are a few features that "sneaked" into this PR:
- If a user saves multiple times in a row, the gen_server is killed and restarted in order to stop currently running work and re-run the diagnostics with the latest state. This addresses some bugs I've seen with my colleagues where the `'textDocument/publishDiagnostics'` response sometimes seems to have been received out of order. By avoiding multiple publishDiagnostics messages as much as possible, we can hopefully reduce or completely get rid of this problem. It also saves significant time when running Erlang LS with Dialyzer enabled. 
- I added the first test for Dialyzer which blew up the diff a bit in `els_diagnostics_SUITE` and added some new files to accommodate the changes.

### Open questions:
* I don't like the fully qualified function call in `els_dialyzer_diagnostics` to an internal function. It feels like a hack but after banging my head trying to get `meck` to accept my mocked `Args`, I gave in. If anyone has any idea how to do this differently, let me know :+1: 
* Restarting the gen_server in case of a "quick save after a previous save", saves significant time when running with dialyzer in Erlang LS. It does however feel a bit hack-ish to just kill the gen_server. If anyone has any idea how to do this differently, let me know :+1: 
* I'd like to move away from `els_diagnostics_server:on_save/1` and  `els_diagnostics_server:on_save/1` and move these calls into `els_diagnostics.erl`. Looking by how that module is defined currently, it does feel like the wrong place as well. Any suggestions on where to place those functions? 
* Please go over the supervisor and child_spec :+1: I don't write them every day so an extra pair of eyes would be good. 

### Overview of work done:
- Move diagnostics to a gen_server based solution
- Keep state on diagnostics so that did_open/1, diagnostics can be presented,
  rather than re-run
- Async diagnostics for all diagnostics

Tests:
- wait_for_notification/0 now receives the Type in order to receive
  specific notifications
- Add test for els_dialyzer_diagnostics

Misc:
- Add exception dialyzer.plt in code_navigation directory
- Re-sort some code that appeared to have diverged in sorting

#### Edit:
The pipeline fails but a local `make ci` works for me. I'll try and figure out why, I have an idea but for the purpose of this initial review, it's not relevant.


src/els_diagnostics_server.erl
```diff
@@ -0,0 +1,168 @@
...
+%%==============================================================================
+%% Macros
+%%==============================================================================
+
+%%==============================================================================
+%% Record Definitions
+%%==============================================================================
+-record(state, { server_name :: atom()
+               , uri :: uri()
+               , diagnostics :: term()
```
alanz: It might make sense to keep the diagnostics in a map indexed by source (compiler, elvis, dialyzer) so that it is easy to add or update diagnostics for a new source.  This makes it easier to run the diagnostics sources async, and replace older diagnostics with new ones, but just for a given source.

src/els_diagnostics_server.erl
```diff
@@ -0,0 +1,168 @@
...
+    lists:foldl(
+      fun(Mod, AccumulatedDiagnostics0) ->
+          Diagnostics = diagnostic(Mod, Uri),
+          AccumulatedDiagnostics = AccumulatedDiagnostics0 ++ Diagnostics,
+          send_notification(Uri, AccumulatedDiagnostics),
+          AccumulatedDiagnostics
+      end, [], [ els_compiler_diagnostics
+               , els_elvis_diagnostics
+               , els_dialyzer_diagnostics
+               ]),
```
alanz: Eventually these should all run async, so we can report say compiler diagnostics as soon as they are ready, without having to wait for dialyzer to finish.

src/els_diagnostics_server.erl
```diff
@@ -0,0 +1,168 @@
...
+on_save(Uri) ->
+  els_diagnostics_sup:on_save(Uri),
+  Id = binary_to_atom(Uri, utf8),
+  ok = gen_server:cast(Id, run).
+
+-spec on_open(uri()) -> ok.
+on_open(Uri) ->
+  els_diagnostics_sup:on_open(Uri),
+  Id = binary_to_atom(Uri, utf8),
+  case gen_server:call(Id, get_diagnostics, 100) of
```
alanz: Are these diagnostics sorted by location?  The first 100 would make more sense than a random set.

src/els_diagnostics_server.erl
onno-vos-dev: I like that idea 👍 I'd have to update the API to allow for that while currently we always run all diagnostics (as long as they're enabled) so I'd like to add the map but avoid adding the individual apis since they'd not be used yet. Wdyt?

src/els_diagnostics_server.erl
onno-vos-dev: All of them actually run async due to the gen_server:cast and each time a diagnostic becomes available, all diagnostics are sent off in the Publish Diagnostics message. If we don't send all, the other diagnostics are lost to the IDE.

src/els_diagnostics_server.erl
onno-vos-dev: The 100 is the timeout in ms for this call. I found some cases where we got a timeout on this call and I wanted to shorten it as much as what I deemed reasonable.

For each uri we spin up a new gen_server so all gen_server are isolated from eachother and you can quickly jump between files, save or open and get the diagnostics async. (For example openening multiple files starts to fire up a bunch of gen_server and starts creating diagnostics for each and fire off a bunch of Publish Diagnostics messages.

jfacorro: I would remove the `get_diagnostics` feature, kill the diagnostics process for each `Uri` when the file is closed and just run the diagnostics again when it is opened. Therefore avoiding this logic completely.

src/els_diagnostics_server.erl
```diff
@@ -0,0 +1,168 @@
...
+    undefined -> gen_server:cast(Id, run);
+    Diagnostics -> Diagnostics
+  end.
+
+-spec diagnostic(atom(), uri()) -> [diagnostic()].
+diagnostic(els_compiler_diagnostics, Uri) ->
+  CDiagnostics = els_compiler_diagnostics:diagnostics(Uri),
+  maybe_compile_and_load(Uri, CDiagnostics),
```
jfacorro: Eventually I would really like to remove this special case for the `els_compiler_diagnostics`. Ideally there would be no coupling between these two, otherwise we could just move the whole `maybe_compile_and_load/2` function into the `els_compiler_diagnostics`.

src/els_app.erl
```diff
@@ -22,7 +22,8 @@
 -spec start(normal, any()) -> {ok, pid()}.
 start(_StartType, _StartArgs) ->
   ok = application:set_env(elvis, no_output, true),
-  els_sup:start_link().
+  els_sup:start_link(),
+  els_diagnostics_sup:start_link().
```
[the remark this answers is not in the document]
onno-vos-dev: No reason other than that during the development I had a few crashes here and there and ELS not starting made debugging that challanging :smile: Will fix

src/els_diagnostics_server.erl
onno-vos-dev: It feels kinda dirty to add that call to `els_client` which is the only place that handles the `did_close`. Re-running diagnostics on `didOpen` every time in particular when Dialyzer is enabled (or a large file is opened) can be quite expensive. Reporting what's already there can save a lot of overhead and time. 

Can you elaborate why you'd want to get rid of this?

src/els_dialyzer_diagnostics.erl
```diff
@@ -15,6 +15,20 @@
...
 
+%% For testing
+-export([ run_dialyzer/1
+        ]).
+
+%%==============================================================================
+%% Types
+%%==============================================================================
+-type dialyzer_arg() ::  {files, [file:filename()]}
+                       | {from, src_code}
```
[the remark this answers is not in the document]
onno-vos-dev: I'll move the `|` to the end of each tuple and align them :+1:

src/els_dialyzer_diagnostics.erl
```diff
@@ -31,11 +45,16 @@ diagnostics(Uri) ->
...
-                            , {include_dirs, els_config:get(include_paths)}
-                            , {plts, [DialyzerPltPath]}
-                            ])
+      WS = try
+             Args = [ {files, Files}
+                    , {from, src_code}
+                    , {include_dirs, els_config:get(include_paths)}
+                    , {plts, [DialyzerPltPath]}
+                    ],
+             %% Need the ?MODULE prefix here for mecking purposes in
```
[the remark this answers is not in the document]
onno-vos-dev: I tried this and for the life of me couldn't get it to accept my meck. (I have absolutely no idea why...) It seemingly just ignored my meck...

src/els_text_synchronization.erl
```diff
@@ -7,86 +7,26 @@
...
 did_save(Params) ->
   TextDocument = maps:get(<<"textDocument">>, Params),
   Uri          = maps:get(<<"uri">>         , TextDocument),
-  spawn (?MODULE, generate_diagnostics, [Uri]),
+  els_diagnostics_server:on_save(Uri),
   ok.
 
 -spec did_close(map()) -> ok.
-did_close(_Params) -> ok.
+did_close(_Params) ->
```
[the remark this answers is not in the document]
onno-vos-dev: See: https://github.com/erlang-ls/erlang_ls/pull/505#discussion_r386970496 for my reason why I chose not to.

src/els_diagnostics_server.erl
```diff
@@ -0,0 +1,168 @@
...
+-include("erlang_ls.hrl").
+
+%%==============================================================================
+%% Macros
+%%==============================================================================
+
+%%==============================================================================
+%% Record Definitions
+%%==============================================================================
+-record(state, { server_name :: atom()
```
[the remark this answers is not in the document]
onno-vos-dev: True, it used to be but due to changes, this got left behind. (Strangely Dialyzer/xref didn't pick this up :thinking: )

src/els_diagnostics_server.erl
onno-vos-dev: I agree, and seeing as I'd like to extend the "do-something-on-a-remote-node"functionality that I introduced previously, I think we should have a re-thinking of this. Are you fine with discussing this in an issue separate of this PR?

src/els_diagnostics_server.erl
jfacorro: Elvis is very cheap, Dialyzer is not be, but optimizing for the special case of people opening and closing the same file all the time is not the way to go IMO.

onno-vos-dev: Fair enough :+1:

src/els_diagnostics_server.erl
alanz: I am happy with small steps in the right direction.  I think we need to end up with a completely pluggable system, where we can load something with the appropriate behaviour, and register it as something contributing to the language server.

jfacorro:
@onno-vos-dev What is your plan regarding the changes and comments done in this PR?

onno-vos-dev:
> @onno-vos-dev What is your plan regarding the changes and comments done in this PR?

@jfacorro I have not forgotten about this branch :) I have branched off of it in order to avoid noise here and have been working on it. I got side-tracked trying to solve an issue I've seen (also on current master) when quick consecutive saves leave the old diagnostics hanging. I hope to finish this branch over the weekend but everything that I've marked as resolved, has been resolved in that branch (which I'll push to this branch and then PR it to master)

robertoaloi:
Closing this one as superseded by #617

PR #517, opened 2020-03-04
[#514] Register commands with a unique prefix

### LINKED ISSUE #514: Registering commands for multiple sessions fails
**Describe the bug**

When #314 landed it started registering commands with the client.  There is only one, called 'replace-lines'.  If you run more than one instance of a server against vscode, the second one fails because the command is already registered.

**To Reproduce**

Set up a workspace with multiple folders, e.g. erlang_ls source and your production source. Open a file from each.  The log in the second one (from the output pane) shows something like

```
[Error - 3:24:28 AM] failed to initialize - command 'replace-lines' already exists
[Trace - 3:24:28 AM] Sending request 'shutdown - (1)'.
```

**Expected behavior**

The server starts up and operates normally

**Actual behavior**

The server shuts down.

**Context**
 - `erlang_ls` version (tag/sha): 026a9a701b5ef0a240c824af5c3de803beb6c0bf
 - Editor used: vscode
 - LSP client used: company internal, based on OSS one

alanz:
The LSP executeCommand process requires the commands to be first registered with
the client. Vscode has a single registry of commands, for the entire client, and
shuts down a server if it tries to register a duplicate command.

This diff ensures that all commands used are unique, by prepending the
underlying operating system process id to the command when registered and used.

Fixes #514.

### DIFF
--- a/src/els_execute_command_provider.erl
+++ b/src/els_execute_command_provider.erl
@@ -5,6 +5,8 @@
 -export([ handle_request/2
         , is_enabled/0
         , options/0
+        , add_server_prefix/1
+        , strip_server_prefix/1
         ]).
 
 -include("erlang_ls.hrl").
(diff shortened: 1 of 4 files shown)


src/els_execute_command_provider.erl
```diff
@@ -48,3 +50,24 @@ execute_command(Command, Arguments) ->
...
+  [_, Command] = binary:split(PrefixedCommand, <<":">>),
+  Command.
+
+%% @doc Add a server-unique prefix to a command.
+-spec add_server_prefix(binary()) -> binary().
+add_server_prefix(Command) ->
+  Prefix = server_prefix(),
+  <<Prefix/binary, ":", Command/binary>>.
+
+%% @doc Generate a prefix unique to this running erlang_ls server.  This is
```
robertoaloi: Nits: typo `vscdoe`. Also, I would avoid mentioning specific clients in the code/comment and keep the comments editor-agnostic.

alanz: Ok. In this case it *is* driven by a particular design decision in a particular client

PR #521, opened 2020-03-05
[#518] Replace group leader when using stdio transport

### LINKED ISSUE #518: Restrict the processes that can write to standard output when using stdio transport
**Describe the bug**
There are some external dependencies that are sometimes writing to `stdout` to report warning or errors. If the server is communicating through `stdio` the intervention of these dependencies corrupt the messages and some clients start failing completely.

**To Reproduce**
Create a module `foo` that has a duplicated type spec like the following:

```erlang
-module(foo).
-export([baz/0]).
-type bar() :: atom().
-type bar() :: atom().

-spec baz() -> bar().
baz() -> ok.
```

Call `foo:baz()` from some other module and hover over the function call. 

In VS Code this will cause the client to stop being able to process any response from the server. Opening the deveoper tools (i.e. `Cmd+Shift+P -> Developer: Toogle Developer Tools`) will show one or more instances of the following error: `Error: Header must provide a Content-Length property.`.

In Emacs this isn't failing, I suspect it's because the client is more forgiving and is somehow able to recover from the cocrrupted messages.

**Expected behavior**
Nothing other that `els_stdio` should be able to write to standard output.

**Actual behavior**
Some clients (e.g. VS Code) stop working when something else writes to standard output.

**Context**
 - `erlang_ls` version (tag/sha): `0.1.0-49-g134838a`
 - Editor used: `VS Code 1.42.0`
 - LSP client used: `Erlang LS 0.0.9`

jfacorro:
### Description

Replace group leader when using stdio transport.

Fixes #518.

### DIFF
--- a/src/els_sup.erl
+++ b/src/els_sup.erl
@@ -40,6 +40,8 @@ init([]) ->
               , period    => 60
               },
   {ok, Transport} = application:get_env(erlang_ls, transport),
+  %% Restrict access to stdio when using that transport
+  (Transport =/= els_stdio) andalso restrict_stdio_access(),
   ChildSpecs = [ #{ id       => els_server
                   , start    => {els_server, start_link, [Transport]}
                   }
@@ -57,3 +59,44 @@ init([]) ->
                   }
                ],
   {ok, {SupFlags, ChildSpecs}}.
+
+%% @doc Restrict access to standard I/O
+%%
+%% Sets the `io_device' application variable to the current group
+%% leaders and replaces the group leader process of this supervisor,
+%% for a fake one. This fake group leader is propagated to all of this
+%% supervisor's children.
+%%
+%% This prevents any library that decides to write anything to
+%% standard output from corrupting the messages sent through JSONRPC.
+%% This problem is happening for example when calling `edoc:get_doc/2',
+%% which can print warnings to standard output.
+-spec restrict_stdio_access() -> ok.
+restrict_stdio_access() ->
+  lager:info("Use group leader as io_device"),
+  case application:get_env(erlang_ls, io_device, standard_io) of
+    standard_io ->
+      application:set_env(erlang_ls, io_device, erlang:group_leader());
+    _ -> ok
+  end,
+
+  lager:info("Replace group leader to avoid unwanted output to stdout"),
+  Pid = erlang:spawn(fun noop_group_leader/0),
+  erlang:group_leader(Pid, self()),
+
+  ok.
+
+%% @doc Simulate a group leader but do nothing
+-spec noop_group_leader() -> no_return().
+noop_group_leader() ->
+  receive
+    Message ->
+      lager:info("noop_group_leader got [message=~p]", [Message]),
+      case Message of
+        {io_request, From, ReplyAs, _} ->
+          From ! {io_reply, ReplyAs, ok};
+        _ ->
+          ok
+      end,
+      noop_group_leader()
+  end.


jfacorro:
> I saw in reading that the group leader is abused to harvest processes when shutting down. Will that be compromised by this?

I don't think it will. The processes that wouldn't be killed are the ones that are not under the supervision tree, as long as every process we spawn is linked to a process in the supervision tree it should be fine.

> Will it matter, given we shut down the O/S process as a whole when terminating the server?

I agree that it ultimately does not matter for the users. If by any chance some dangling process is still alive when the application is stopped, this could affect some of the tests, depending on what the process does. But this would still happen without this change in the PR, @robertoaloi was having some issues with the process that keeps the TCP connection in his indexing PR.

alanz:
Running this under emacs, I get

```
LSP :: Connected to [erlang-ls:14246 status:starting].
LSP :: erlang-ls:14246 initialized successfully
Error processing message (lsp-unknown-message-type #s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("error" #s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("code" -32001 "message" "Unexpected error while initialized")) "id" nil "jsonrpc" "2.0"))).
```

Bisecting blames 9050e6a4c8c6e6fbd2ecd13a13755cdffe27c308

robertoaloi:
If possible, I would appreciate @dszoboszlay 's input on this one.

alanz:
BTW, the error in emacs is complaining about this line getting hit

https://github.com/erlang-ls/erlang_ls/blob/9050e6a4c8c6e6fbd2ecd13a13755cdffe27c308/src/els_methods.erl#L73

alanz:
My setup is mac, catalina, emacs `GNU Emacs 26.3 (build 1, x86_64-apple-darwin18.2.0, NS appkit-1671.20 Version 10.14.3 (Build 18D109)) of 2019-09-02`

`Erlang/OTP 22 [erts-10.6.4] [source] [64-bit] [smp:16:16] [ds:16:16:10] [async-threads:1] [hipe] [dtrace]
`

jfacorro:
I found the problem! We are using `application:get_application()` which is relying on the `group_leader()`:
```
-spec get_application() -> 'undefined' | {'ok', Application} when
      Application :: atom().
get_application() -> 
    application_controller:get_application(group_leader()).
```

src/els_sup.erl
```diff
@@ -40,6 +40,8 @@ init([]) ->
               , period    => 60
               },
   {ok, Transport} = application:get_env(erlang_ls, transport),
+  %% Restrict access to stdio when using that transport
+  (Transport =/= els_stdio) andalso restrict_stdio_access(),
```
dszoboszlay: Isn't this the opposite of what you want? You call `restrict_stdio_access()` for every transport _except for_ `els_stdio`.

robertoaloi: It looks like @dszoboszlay is correct. I also suggest we do not rely on the `andalso` semantics for this, since it makes the code less readable. I would rather:

```
restrict_stdio_access(els_stdio) ->
  restrict_stdio_access();
restrict_stdio_access(_) ->
  ok.
```

dszoboszlay:
All in all, I don't think you should base your solution on changing the `group_leader` in one supervisor.

* It prevents `application:get_application()` from working properly.
* If you start a different application with a supervisor tree, this solution will not block that application's processes from writing to the standard output.

I'd recommend taking a completely different approach:

* Start Erlang with the `-nouser` command line argument. This will tell the `kernel` not to start a `user` process, basically blocking access to the `standard_io` I/O device.
* You can still access the standard output of the OS process by opening `{fd, 0, 1}` as a port:

```erlang
P = erlang:open_port({fd, 0, 1}, []),
erlang:port_command(P, <<"hello world!\n">>),
erlang:port_close(P).
```

It is very unlikely that any other library would accidentally write to this port! :smile:

src/els_sup.erl
jfacorro: 🤦

jfacorro: #532 532

PR #522, opened 2020-03-05
[#463] Speed up indexing

### LINKED ISSUE #463: Potential indexing improvements
* Request a table lock
* Consider using `mnesia:ets/2`
* Optimize delete/read sequences

robertoaloi:
### Description

This Pull Request completely revisits the indexing process. In the sample code base I'm experimenting with (~1.5 million lines of code) the "waiting" time on a restart (i.e. with an existing DB which needs to be loaded in memory) went from ~1 minute to ~17 seconds. I would appreciate if you could try it with other code bases too. This PR also gets rid of the worker pool for indexing, preparing for a follow-up PR where we can display indexing progress to the user.

Main changes:

* Remove worker pool for indexing
* Force a DB dump after indexing (so it's done in the background instead of on restart)
* Do not compress tables (with a little loss in terms of memory, we gain a lot in loading speed)
* Introduce `shallow` cloning for OTP libraries (i.e. do not store references for OTP libraries, not required for most use cases, will follow-up with a PR which makes these options configurable)

I plan to write a blog post about the findings, if I ever get the time to do so.

Fixes #463 .

### DIFF
--- a/src/els_indexer.erl
+++ b/src/els_indexer.erl
@@ -151,6 +151,11 @@ handle_call(_Request, _From, State) ->
 -spec handle_cast(any(), state()) -> {noreply, state()}.
 handle_cast({start, Dirs, Mode}, State) ->
   [index_dir(Dir, Mode) || Dir <- Dirs],
+  %% Indexing a directory can lead to a huge number of DB transactions
+  %% happening in a very short time window. After indexing, let's
+  %% manually trigger a DB dump. This ensures that the DB can be
+  %% loaded much faster on a restart.
+  els_db:dump_tables(),
   {noreply, State};
 handle_cast(_Msg, State) ->
   {noreply, State}.
--- a/src/els_indexer.erl
+++ b/src/els_indexer.erl
@@ -7,7 +7,7 @@
 -export([ find_and_index_file/1
         , index_file/1
         , index/3
-        , index_dir/2
+        , index_dirs/2
         , start_link/0
         , start/2
         ]).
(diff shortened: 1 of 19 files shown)


src/els_indexer.erl
```diff
@@ -229,3 +181,30 @@ register_reference(Uri, #{id := {M, F, A}, range := Range}) ->
                             , uri   => Uri
                             , range => Range
                             }).
+
+-spec index_dir(string(), mode()) -> {non_neg_integer(), non_neg_integer()}.
+index_dir(Dir, Mode) ->
+  lager:info("Indexing directory. [dir=~s] [mode=~s]", [Dir, Mode]),
```
jfacorro: Could we change this and the "Finished indexing..." log messages from `info` to `debug`? If they are at the `info` level it makes the logs very verbose by default.

alanz:
It seems this diff no longer cleanly shuts down mnesia.

```
2020-03-06 02:37:50.568 [info] <0.164.0>@els_indexer:index_dir:187 Indexing directory. [dir=/data/users/alanzimm/whatsapp/server/erl/wa/src] [mode=deep]
2020-03-06 02:37:50.858 [info] <0.161.0>@els_methods:initialized:183 initialized: [App=erlang_ls] [Version="0.1.0+build.819.refbedca34"]
2020-03-06 02:37:50.873 [info] <0.57.0> SIGTERM received - shutting down
2020-03-06 02:37:50.895 [info] <0.161.0>@els_methods:exit:208 Language server stopping...
2020-03-06 02:37:50.902 [error] <0.164.0>@els_indexer:try_index_file:157 Error indexing file [filename=/data/users/alanzimm/whatsapp/server/erl/wa/src/wa_sam_server.erl] exit:{aborted,{no_exists,[els_dt_document,<<"file:///data/users/alanzimm/whatsapp/server/erl/wa/src/wa_sam_server.erl">>]}}:[{mnesia,abort,1,[{file,"mnesia.erl"},{line,355}]},{els_db,lookup,2,[{file,"/home/alanzimm/erlang_ls/src/els_db.erl"},{line,89}]},{els_dt_document,lookup,1,[{file,"/home/alanzimm/erlang_ls/src/els_dt_document.erl"},{line,125}]},{els_indexer,index,3,[{file,"/home/alanzimm/erlang_ls/src/els_indexer.erl"},{line,64}]},{els_indexer,try_index_file,2,[{file,"/home/alanzimm/erlang_ls/src/els_indexer.erl"},{line,155}]},{els_indexer,'-index_dir/2-fun-2-',3,[{file,"/home/alanzimm/erlang_ls/src/els_indexer.erl"},{line,189}]},{els_utils,do_fold_files,5,[{file,"/home/alanzimm/erlang_ls/src/els_utils.erl"},{line,112}]},{timer,tc,3,[{file,"timer.erl"},{line,197}]}]
```

This log fragment is an extreme case, with vscode starting up an `erlang_ls` instance for the erlang_ls source directory because it is part of the configured workspace, but then almost immediately terminating it because there are no open files in the editor.

I presume this should be handled gracefully?

src/els_indexer.erl
```diff
@@ -88,61 +84,34 @@ do_index(#{uri := Uri, id := Id, kind := Kind} = Document) ->
...
+      POIs  = els_dt_document:pois(Document, [ application
+                                             , implicit_fun
+                                             , macro
+                                             , record_access
+                                             , record_expr
+                                             ]),
+      ok = els_dt_references:delete_by_uri(Uri),
+      [register_reference(Uri, POI) || POI <- POIs],
+      ok;
+    'shallow' ->
```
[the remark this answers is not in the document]
robertoaloi: The common part of the indexing is done before the `case`. This includes documents and signatures. What this PR is removing is storing OTP _references_. This means that code navigation to OTP modules works as before, and so does the docs on hover. If you include an OTP file (e.g. `assert.hrl`), you still get the definitions as before. If you open an OTP file, that file and its dependencies are indexed.

The only feature which disappears is the possibility of "find who calls lists:member within OTP", which I believe is fine. You can also open OTP as a `workspace/project` and get everything, if you want.

And the otp_indexing level will be configurable.

src/els_indexer.erl
robertoaloi: Yes, I can do that.

robertoaloi: Fixed.

robertoaloi:
> This log fragment is an extreme case, with vscode starting up an `erlang_ls` instance for the erlang_ls source directory because it is part of the configured workspace, but then almost immediately terminating it because there are no open files in the editor.

@alanz Could you tell me the steps I should follow to reproduce this locally?

robertoaloi:
@alanz We are shutting down Mnesia as a consequence of the shutdown, but the indexer process is still trying to index files in the background and it fails writing. So, it feels like the log message is harmless. We could explicitly stop any eventual indexing, but it feels un-necessary since we're halting the node, anyway.

robertoaloi:
Created https://github.com/erlang-ls/erlang_ls/issues/529 as a follow-up.

alanz:
My experience on a large code base

Before this diff
  - wait 8 mins for mnesia tables to load
  - database size of around 930MB
After this diff
  - wait 45 secs for mnesia tables to load
  - database size of around 840MB

code size:

```
 $ find ./ -name '*.[er( find ./ -name '*.erl' -print0 | xargs -0 cat ) | wc -l
1950428
$ ( find ./ -name '*.hrl' -print0 | xargs -0 cat ) | wc -l
67583
```

PR #528, opened 2020-03-06
[#516] Use syntax highlighting when using markdown to show information on hover 

### LINKED ISSUE #516: Erlang syntax highlighting for hover information
It would be nice to display Erlang code in the hover information formatted with Erlang syntax highlighting.

When the client supports markdown, it is possible to specify the language for a block of code by doing the following:

    ```erlang
    -spec foo() -> ok.
    ```

When the client doesn't support markdown it could fallback to the less fancy plain text formatting.

N/A.

**Additional context**
N/A.

jfacorro:
### Description

Use syntax highlighting when using markdown to show information on hover.

Fixes #516.

### DIFF
--- a/src/els_hover_provider.erl
+++ b/src/els_hover_provider.erl
@@ -61,8 +61,10 @@ get_docs(M, F, A) ->
     {<<>>, <<>>} ->
       <<>>;
     {Specs, Edoc} ->
-      #{ kind => content_kind()
-       , value => <<Specs/binary, "\n\n", Edoc/binary>>
+      ContentKind = content_kind(),
+      FormattedSpecs = format_code(ContentKind, Specs),
+      #{ kind  => ContentKind
+       , value => << FormattedSpecs/binary, "\n", Edoc/binary>>
        }
   end.
 
@@ -72,9 +74,18 @@ get_docs(M, F, A) ->
 -spec specs(atom(), atom(), non_neg_integer()) -> binary().
 specs(M, F, A) ->
   case els_dt_signatures:lookup({M, F, A}) of
-    {ok, [#{tree := Tree}]} -> list_to_binary(erl_prettypr:format(Tree));
-    {ok, []}                -> <<>>
+    {ok, [#{tree := Tree}]} ->
+      Specs = erl_prettypr:format(Tree),
+      unicode:characters_to_binary(Specs);
+    {ok, []} ->
+      <<>>
   end.
+
+-spec format_code(markup_kind(), binary()) -> binary().
+format_code(plaintext, Code) ->
+  Code;
+format_code(markdown, Code) ->
+  <<"```erlang\n", Code/binary, "\n```\n">>.
 
 -spec edoc(atom(), atom(), non_neg_integer()) -> binary().
 edoc(M, F, A) ->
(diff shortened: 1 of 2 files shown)


src/els_hover_provider.erl
```diff
@@ -103,7 +114,7 @@ format(Signature, Desc) when is_map(Desc) ->
   Lang         = <<"en">>,
   Doc          = maps:get(Lang, Desc, <<>>),
   FormattedDoc = list_to_binary(docsh_edoc:format_edoc(Doc, #{})),
-  <<"# ", Signature/binary, "\n", FormattedDoc/binary>>.
+  <<"### ", Signature/binary, "\n", FormattedDoc/binary>>.
```
alanz: yay

dgud:
Hmm now the spec look horrible for me.
Emacs version: 25.2.2
(I guess this branch caused it, I have not done a bisect)
Screenshot from 2020-03-09 15-40-00

alanz:
@dgud could this be because of https://github.com/emacs-lsp/lsp-ui/issues/285?

dgud:
Will try to debug tomorrow, I'm on linux but have on "old" emacs.

The problem seems to be the font-size in mark-down is way to large.
And the coloring is horrible when using a dark background.

jfacorro:
FWIW this is how it looks like in my current Emacs (26.1) setup:
Screenshot 2020-03-10 at 08 32 25

I'm using the [`monkai-theme`](https://github.com/oneKelvinSmith/monokai-emacs).

alanz:
It looks to me that @dgud is running in terminal mode (no overlay) and @jfacorro is using the overlay frame.  Which is what I use too.

dgud:
Hmm how do I enable overlay frame?

dgud:
emacs 26.3 works better, though the colors are still not great.
It's a pity that emacs-25 is that bad though that is the default for LTS-ubuntu 18.04.

@jfacorro I use deeper-blue theme a default-installed theme I believe.

jfacorro:
> Hmm how do I enable overlay frame?

I think the overlay is available because we are using GNU Emacs for Mac OS X. @robertoaloi uses Emacs in the terminal. How does it look for you @robertoaloi?

robertoaloi:
I do not use Emacs in the terminal. @dgud Could you please try how it looks by starting your Emacs 25 with the minimal `dotemacs` file we provide in this repo? That will ensure the weird formatting is not due to your specific theme or configuration:

https://github.com/erlang-ls/erlang_ls#troubleshooting-the-emacs-setup

Ensure you rename your `~/.emacs.d` folder when trying this.

dgud:
@robertoaloi 25 looks like above (minus the color scheme), and 26 looks good still a larger font but it fits the windows without breaking lines and moving code around when displayed.

I run `emacs -q -l erlang_ls/misc/..` 
Without removing the .emacs.d since then I can't talk to melpa, emacs-25's certs are outdated and I don't remember how I fixed that :-)

So it's I think it's an emacs-25 issue, but maybe an option for disabling the markdown formatting.

robertoaloi:
@dgud The `markdown` option is explicitly requested by the client (`lsp-mode` in this case). The server looks at the client capabilities and, if `markdown` is supported by the client, it favours that one. Otherwise, it falls back to use `plaintext`. In my view, this should be an option in `lsp-mode`:

https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L2953

PR #535, opened 2020-03-07
[#534] Fix off by one in folding ranges

### LINKED ISSUE #534: Folding ranges are off by one line
**Describe the bug**
All functions' folding range don't include the first head clause and include the line right after them.

**To Reproduce**
Use an IDE+LSP Client editor that supports and displays folding ranges (e.g. VSCode) then open any Erlang file that contains a function definition.

*Expanded*
Screenshot 2020-03-07 at 12 32 26
*Collapsed*
Screenshot 2020-03-07 at 12 32 19

**Expected behavior**
The folding range should include the head and the last non-empty line that is part of the function so that when collapsed only the first function head clause shows.

**Actual behavior**
When collapsed the first line of code right after the first head clause is shown and the line after the function is included in the collapsed lines.

**Context**
 - `erlang_ls` version (tag/sha): `0.1.0-80-g3453edf`
 - Editor used: `VS Code`
 - LSP client used: `0.0.9`

jfacorro:
### Description

Fix off by one in folding ranges.

Fixes #534.

### Emacs
Screenshot 2020-03-07 at 23 42 38

### VS Code
Screenshot 2020-03-07 at 23 44 30

### DIFF
--- a/include/erlang_ls.hrl
+++ b/include/erlang_ls.hrl
@@ -80,6 +80,14 @@
                      , character := number()
                      }.
 
+%% This is used for defining folding ranges. It is not possible to just use
+%% positions as this one `{Line + 1, 0}' in a folding range, because of
+%% differences on how clients behave in that case (see [#535]).
+%% According to the protocol "If the character value is greater than the line
+%% length it defaults back to the line length.". So we define a big enough
+%% value that represents the end of the line.
+-define(END_OF_LINE, 9999999).
+
 %%------------------------------------------------------------------------------
 %% Range
 %%------------------------------------------------------------------------------
@@ -93,6 +101,16 @@
 -type location() :: #{ uri   := uri()
                      , range := range()
                      }.
+
+%%------------------------------------------------------------------------------
+%% Folding Range
+%%------------------------------------------------------------------------------
+
+-type folding_range() :: #{ startLine      := pos_integer()
+                          , startCharacter := pos_integer()
+                          , endLine        := pos_integer()
+                          , endCharacter   := pos_integer()
+                          }.
 
 %%------------------------------------------------------------------------------
 %% Diagnostic
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -17,7 +17,7 @@
 
 -define(SERVER, ?MODULE).
 -define(TIMEOUT, infinity).
--define(DB_SCHEMA_VSN, <<"4">>).
+-define(DB_SCHEMA_VSN, <<"5">>).
 -define(DB_SCHEMA_VSN_FILE, "DB_SCHEMA_VSN").
 
 %%==============================================================================
--- a/src/els_folding_range_provider.erl
+++ b/src/els_folding_range_provider.erl
@@ -11,11 +11,7 @@
 %%==============================================================================
 %% Type Definitions
 %%==============================================================================
--type folding_range() :: #{ startLine      := pos_integer()
-                          , startCharacter := pos_integer()
-                          , endLine        := pos_integer()
-                          , endCharacter   := pos_integer()
-                          }.
+
 -type folding_range_result() :: [folding_range()] | null.
 
 %%==============================================================================
@@ -32,7 +28,7 @@ handle_request({document_foldingrange, Params}, State) ->
   #{ <<"textDocument">> := #{<<"uri">> := Uri} } = Params,
   {ok, [Document]} = els_dt_document:lookup(Uri),
   POIs = els_dt_document:pois(Document, [folding_range]),
-  Response = case [folding_range(POI) || POI <- POIs] of
+  Response = case [folding_range(Range) || #{range := Range} <- POIs] of
                []     -> null;
                Ranges -> Ranges
              end,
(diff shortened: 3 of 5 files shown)


include/erlang_ls.hrl
```diff
@@ -80,6 +80,11 @@
                      , character := number()
                      }.
 
+%% According to the protocol "If the character value is greater than
+%% the line length it defaults back to the line length.". So we define
+%% a big enough value that represents the end of the line.
+-define(END_OF_LINE, 9999999).
```
robertoaloi: Why do we need to do this? Can't we just extract the line length from the DB?

src/els_folding_range_provider.erl
```diff
@@ -42,10 +38,10 @@ handle_request({document_foldingrange, Params}, State) ->
...
 %%==============================================================================
 
--spec folding_range(poi()) -> folding_range().
-folding_range(#{range := #{from := {FromL, FromC}, to := {ToL, ToC}}}) ->
-  #{ startLine      => FromL
-   , startCharacter => FromC
-   , endLine        => ToL
-   , endCharacter   => ToC
+-spec folding_range(poi_range()) -> folding_range().
+folding_range(#{from := {FromLine, FromCol}, to := {ToLine, ToCol}}) ->
```
robertoaloi: This does not seem to work for me:

Screenshot 2020-03-08 at 15 47 45

include/erlang_ls.hrl
jfacorro: It's not the line we need but the last colum/character position of the line.

src/els_folding_range_provider.erl
jfacorro: Have you cleared the cache? Otherwise the POIs don't get updated unless you change the file.

include/erlang_ls.hrl
alanz: I worked around this in haskell-ide-engine (in some other context) by using offset 0 of the next line as the end of the previous one. It is less arbitrary.

jfacorro: I tried the `0` approach but it then folded the next line as well, at least this is what I remember when I tried that. I will check again.

jfacorro: The problem is in the difference in which the editors are processing the folding range information. 

For example Emacs hides the exact range you provide, if the range is `{StartLine, 0} -> {EndLine, 0}` then it looks like this:
Screenshot 2020-03-09 at 07 21 24
But in VS Code it looks like this, because apparently it assumes that when the start character is `0` then it should show the whole `StartLine`:
Screenshot 2020-03-09 at 07 13 38

(Notice that in both cases above the fold is also hiding the new line which shouldn't be part of the range)

When I tried to "solve" the issue by adding one to the start line `{StartLine + 1, 0} -> {EndLine, 0}` it looked like this in Emacs:
Screenshot 2020-03-09 at 07 23 32
But like this in VS Code:
Screenshot 2020-03-09 at 07 24 20

The alternative would be to not provide the start and end characters since [according to the protocol](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_foldingRange) they are both optional and `If not defined, defaults to the length of the end line.`.

jfacorro:
> This does not seem to work well with the latest Emacs client.

@robertoaloi What do you mean by that? I have tested it with `lsp-mode-20200306.626` and it works well with that one.

src/els_folding_range_provider.erl
robertoaloi: @jfacorro Yes, we should bump the DB version for this.

include/erlang_ls.hrl
robertoaloi: Can we add the explanation as a comment next to the `99999` value?

jfacorro: There is an explanation right above the value.

src/els_folding_range_provider.erl
robertoaloi: Tried after clearing the cache, it seems to work correctly in Emacs, even if the `...` is appended to the first line instead of being on a new line, but I'm sufficiently happy with that. :)

robertoaloi:
@jfa

> > This does not seem to work well with the latest Emacs client.
> 
> @robertoaloi What do you mean by that? I have tested it with `lsp-mode-20200306.626` and it works well with that one.

The issue was due to the lack of DB version bump (i.e. dirty cache).

include/erlang_ls.hrl
robertoaloi: I'm referring to the fact that we're not using a `Line + 1, 0` range because of differences on how clients behave in that case. The reason for having that is that, if we don't explicitly mention it, someone (e.g. an Emacs user) will "fix" that and potentially break things again for VSCode users.

jfacorro: Added more detailed comment.

src/els_db.erl
```diff
@@ -17,7 +17,7 @@
 
 -define(SERVER, ?MODULE).
 -define(TIMEOUT, infinity).
--define(DB_SCHEMA_VSN, <<"4">>).
+-define(DB_SCHEMA_VSN, <<"5">>).
```
jfacorro: Bumped DB schema version.

PR #536, opened 2020-03-09
Use OTP's get_doc if available

dgud:
EEP48 provides documentation for erlang application/modules.
OTP-23 implements EEP48 and provide the doc info if available.

If user uses an OTP version that provides the functionality and the
doc_chunks are available use that by default and if not fallback to
fetching the docs from src via 'erlang_ls' own implementation.

This is a foundation for further work where we should write an
doc_chunk to markdown renderer, when markdown available in the client.

### DIFF
--- a/src/els_hover_provider.erl
+++ b/src/els_hover_provider.erl
@@ -55,8 +55,44 @@ documentation(_M, _POI) ->
   <<>>.
 
 %% @doc get the docs
+%%
+%% Uses code:get_doc/1 if available and docs are available
+%% otherwise uses the source files to gather the documentation
+%% code:get_doc/1 is available from OTP-23
+%%
+%% Implemented with apply and hard coded record-tuple so the code can
+%% be compiled on older OTP's.
+%%
+%% Currently uses shell_docs:render/4 (OTP-23) which render plain text.
+%%
+%% Future work: shell_docs:render/4 should probably be copied and
+%% modified to render markdown, or make an PR to OTP with an option
+%% in erlang's shell_docs to render markdown instead of ANSI-codes.
 -spec get_docs(atom(), atom(), byte()) -> binary().
 get_docs(M, F, A) ->
+  try apply(code, get_doc, [M]) of
+    {ok, {docs_v1, _, _, <<"application/erlang+html">>, MDoc, _, _} = DocChunk}
+      when MDoc =/= none ->
+        case apply(shell_docs, render, [M, F, A, DocChunk]) of
+          {error, _R0} ->
+            docs_from_src(M, F, A);
+          FuncDoc ->
+            #{ kind => content_kind()
+             , value => unicode:characters_to_binary(FuncDoc)
+             }
+        end;
+    _R1 ->
+      docs_from_src(M, F, A)
+  catch error:undef ->
+      docs_from_src(M, F, A)
+  end.
+
+%%==============================================================================
+%% Internal functions
+%%==============================================================================
+
+-spec docs_from_src(atom(), atom(), byte()) -> binary().
+docs_from_src(M, F, A) ->
   case {specs(M, F, A), edoc(M, F, A)} of
     {<<>>, <<>>} ->
       <<>>;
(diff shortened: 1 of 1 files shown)


dgud:
Hmm I don't understand while the tests are failing

robertoaloi:
Having a look to the failing test cases...

robertoaloi:
@dgud It looks like the property-based tests are complaining about the `halt` function not being called correctly. I tried running them locally for your branch and they pass. I suspect there's some flakyness around the `shutdown` solution. @jfacorro Do you have any more input?

https://github.com/erlang-ls/erlang_ls/pull/536/checks?check_run_id=494725902

dgud:
Yeah I ran the tests locally before pushing the PR and that worked for me as well.

jfacorro:
I've seen the GitHub job fail from time to time for no obvious reason. My approach so far has been to re-trigger the job, but I just did that and it seems the action is not able to checkout the reference because it's not in the tree:
```
From https://github.com/erlang-ls/erlang_ls
 * [new branch]      285-indexing-progress-reports -> origin/285-indexing-progress-reports
 * [new branch]      285-show-progress-report-during-indexing -> origin/285-show-progress-report-during-indexing
 * [new branch]      458-auto-complete-exported-types -> origin/458-auto-complete-exported-types
 * [new branch]      478-elvis-diagnostics-fail -> origin/478-elvis-diagnostics-fail
 * [new branch]      features-list          -> origin/features-list
 * [new branch]      indexing-no-compressed -> origin/indexing-no-compressed
 * [new branch]      intellij               -> origin/intellij
 * [new branch]      master                 -> origin/master
 * [new branch]      notes                  -> origin/notes
 * [new tag]         0.1.0                  -> 0.1.0
git checkout --progress --force 200ce9fbf446a9f557fe655ab875461620d16fef
##[error]fatal: reference is not a tree: 200ce9fbf446a9f557fe655ab875461620d16fef
Removed matchers: 'checkout-git'
##[error]Git checkout failed with exit code: 128
##[error]Exit code 1 returned from process: file name '/home/runner/runners/2.165.2/bin/Runner.PluginHost', arguments 'action "GitHub.Runner.Plugins.Repository.v1_0.CheckoutTask, Runner.Plugins"'.
```
I'm not sure what's happening there 😞.

jfacorro:
@dgud Could you rebase on `master`? I made a change to the build [here](https://github.com/erlang-ls/erlang_ls/pull/543) so that it at least doesn't fail because of the `Checkout` action.

src/els_hover_provider.erl
```diff
@@ -57,6 +57,33 @@ documentation(_M, _POI) ->
...
+            docs_from_src(M, F, A);
+          FuncDoc ->
+            #{ kind => content_kind()
+             , value => unicode:characters_to_binary(FuncDoc)
+             }
+        end;
+    _R1 ->
+      docs_from_src(M, F, A)
+  catch error:undef ->
+      lager:info("err: ~p~n", [undef]),
```
robertoaloi: We should probably use a `warning` level here and a nicer error message. Also, no need for newline.

```suggestion
      lager:warning("Error fetching documentation. Using an old OTP version?", []),
```

src/els_hover_provider.erl
```diff
@@ -57,6 +57,33 @@ documentation(_M, _POI) ->
 %% @doc get the docs
 -spec get_docs(atom(), atom(), byte()) -> binary().
 get_docs(M, F, A) ->
+  %% Use apply and hard coded record so it can be compiled on older OTP's
```
robertoaloi: We should probably be more specific on what "older" means here. I suggest we update the `edoc` for this function with a nicer description, mentioning the minimum required OTP version for this to work.

src/els_hover_provider.erl
dgud: That is a debug printout used during development, so IMHO it should be removed.

robertoaloi: Fine with that, too.

robertoaloi:
Thanks @dgud ! 🚀

PR #537, opened 2020-03-10
[#458] Auto complete types

### LINKED ISSUE #458: Auto-completion for exported types

jfacorro:
### Description

Auto complete types when in a `spec` or `export_types` context.

Fixes #458.

### DIFF
--- a/src/els_completion_provider.erl
+++ b/src/els_completion_provider.erl
@@ -130,10 +130,10 @@ find_completion( Prefix
     %% Check for "[...] atom"
     [{atom, _, Name} | _] ->
       NameBinary = atom_to_binary(Name, utf8),
-      {ExportFormat, TypeOrFun} = function_context(Document, Line, Column),
+      {ExportFormat, POIKind} = function_context(Document, Line, Column),
       keywords()
       ++ modules(NameBinary)
-      ++ functions(Document, TypeOrFun, ExportFormat);
+      ++ functions(Document, POIKind, ExportFormat);
     _ ->
       []
   end;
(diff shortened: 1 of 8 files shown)


src/els_completion_provider.erl
```diff
@@ -140,44 +158,66 @@ item_kind_module(Module) ->
...
+
+  Exports = local_and_included_pois(Document, ExportKind),
+  POIs    = local_and_included_pois(Document, POIKind),
+  FAs     = [FA || #{id := FA} <- Exports],
+  Items   = resolve_functions(POIs, FAs, ExportedOnly, ExportFormat, ItemKind),
+  lists:usort(Items).
+
+-spec function_context(els_dt_document:item(), line(), column()) ->
+  {boolean(), poi_kind()}.
+function_context(Document, Line, Column) ->
```
jfacorro: I created this function instead of just passing the arguments `Document, Line, Column` to both `functions` and `exported_functions`, because the context that matters (as returned by `function_context`) is the one on the **current** document and not the one in the document we are fetching the exported functions from.

src/els_completion_provider.erl
```diff
@@ -232,50 +268,23 @@ record_fields(Document, RecordName) ->
...
-                       ]),
+  POIs = local_and_included_pois(Document, Type),
   Defs = [ #{ label => atom_to_binary(Name, utf8)
             , kind  => completion_item_kind(Type)
             }
            || #{id := Name} <- POIs
          ],
   lists:usort(Defs).
 
--spec local_definitions(els_dt_document:item(), def_type()) -> [map()].
```
jfacorro: These functions were generic enough that I could just use them without any modification to their logic so I moved them to the general `Helper Functions` sections.

src/els_range.erl
```diff
@@ -17,7 +17,21 @@ compare( #{from := FromA, to := ToA}
 compare(_, _) ->
   false.
 
--spec range(pos(), poi_kind(), any(), any()) -> poi_range().
+-spec range(pos() | {pos(), pos()}, poi_kind(), any(), any()) -> poi_range().
+range({{Line, Column}, {ToLine, ToColumn}}, Name, _, _Data)
```
jfacorro: There were cases (e.g. folding ranges) in which we wanted to provide both the beginning and the end of the range from the upstream logic so it made sense to have the creation here instead of sometimes calling `els_poi:new/[3,4]` from `els_parser` and sometime relying on the logic of `els_range:range/4`.

src/els_completion_provider.erl
```diff
@@ -140,44 +158,66 @@ item_kind_module(Module) ->
...
-  ExportsFA = [FA || #{id := FA} <- Exports],
-  List      = [ completion_item_function(POI, Arity)
-                || #{id := FA} = POI <- Functions, lists:member(FA, ExportsFA)
-              ],
-  lists:usort(List).
-
--spec completion_item_function(poi(), boolean()) -> map().
-completion_item_function(#{id := {F, A}, data := ArgsNames}, false) ->
+-spec functions(els_dt_document:item(), poi_kind(), boolean()) -> [map()].
+functions(Document, POIKind, ExportFormat) ->
```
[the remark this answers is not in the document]
alanz: maybe call it something like `definitions`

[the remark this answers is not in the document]
jfacorro: There is already another function called definitions in this same module 😢.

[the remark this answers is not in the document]
jfacorro: On second thought, maybe I could try to unify them.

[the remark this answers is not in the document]
jfacorro: I ended up unifying them in a single function.

src/els_completion_provider.erl
```diff
@@ -140,63 +158,59 @@ item_kind_module(Module) ->
...
+  definitions(Document, POIKind, _ExportFormat = false, _ExportedOnly = false).
+
+-spec definitions(els_dt_document:item(), poi_kind(), boolean()) -> [map()].
+definitions(Document, POIKind, ExportFormat) ->
+  definitions(Document, POIKind, ExportFormat, _ExportedOnly = false).
+
+-spec definitions(els_dt_document:item(), poi_kind(), boolean(), boolean()) ->
+  [map()].
+definitions(Document, POIKind, ExportFormat, ExportedOnly) ->
+  POIs     = local_and_included_pois(Document, POIKind),
```
robertoaloi: Nit: weird indentation. Also, I suggest we drop the convention of aligning matching operators in function bodies and to limit its use when column-aligning record definitions, types and such. WDYT?

jfacorro: I like the suggestion. 😄

PR #541, opened 2020-03-10
476 use getopt for cli arg parsing

### LINKED ISSUE #476: Use parser for server CLI options
The number of options for the server CLI is growing. Passing a wrong option today results in a verbose error.

```
$ erlang_ls --foo
escript: exception error: bad argument
  in function  list_to_integer/1
     called as list_to_integer("--foo")
...
```

Instead, we should use a parser for the CLI options. E.g.: https://hex.pm/packages/getopt

onno-vos-dev:
### Description
Using getopt allows us to get some nice features for free like an easy utility function for getting out the version from Erlang LS and getting a `help` to keep the arguments to Erlang LS and the documentation of them in one place.

Fixes #476 

Thanks [@sgillis](https://github.com/sgillis) and [@libanomarsweden](https://github.com/libanomarsweden) for helping out! :+1: 

Edit: Can someone restart the Github action for me? It passed on my Github account when I had https://github.com/erlang-ls/erlang_ls/pull/540 on top of this branch so it seems like a Github problem to me

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,7 +119,8 @@ Usage: Erlang LS [-v] [-t [<transport>]] [-p [<port>]] [-d [<log_dir>]]
   -t, --transport  Specifies the transport the server will use for the
                    connection with the client. [default: tcp]
   -p, --port       Used when the transport is tcp. [default: 10000]
-  -d, --log-dir    Directory where logs will be written. [default: <user_log>/erlang_ls/log]
+  -d, --log-dir    Directory where logs will be written.
+                   [default: filename:basedir(user_log, "erlang_ls")]
   -l, --log-level  The log level that should be used. [default: info]
 ```
 
--- a/README.md
+++ b/README.md
@@ -111,11 +111,17 @@ Compile the project:
 These are the command-line arguments that can be provided to the
 `erlang_ls` escript:
 
-| Argument                | Description                                                                                     |
-|-------------------------|-------------------------------------------------------------------------------------------------|
-| --transport tcp / stdio | Specifies the transport the server will use for the connection with the client (default: `tcp`) |
-| --port PORT             | Used when the transport is `tcp` (default: `10000`)                                             |
-| --log-dir DIR           | Directory where logs will be written. When not provided no logs are generated.                  |
+``` shell
+Usage: Erlang LS [-v] [-t [<transport>]] [-p [<port>]] [-d [<log_dir>]]
+                 [-l [<log_level>]] [<port_old>]
+
+  -v, --version    Print the current version of Erlang LS
+  -t, --transport  Specifies the transport the server will use for the
+                   connection with the client. [default: tcp]
+  -p, --port       Used when the transport is tcp. [default: 10000]
+  -d, --log-dir    Directory where logs will be written. [default: <user_log>/erlang_ls/log]
+  -l, --log-level  The log level that should be used. [default: info]
+```
 
 ### Emacs Setup
 
--- a/src/erlang_ls.erl
+++ b/src/erlang_ls.erl
@@ -14,6 +14,7 @@
 main(Args) ->
   %% Initialization
   application:load(lager),
+  application:load(getopt),
   application:load(?APP),
   ok = parse_args(Args),
   ok = lager_config(),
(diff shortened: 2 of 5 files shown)


src/erlang_ls.erl
```diff
@@ -22,38 +23,100 @@ main(Args) ->
   lager:info("Started erlang_ls server", []),
   receive _ -> ok end.
 
+-spec print_version() -> ok.
+print_version() ->
+  {ok, Vsn} = application:get_key(erlang_ls, vsn),
```
jfacorro: Maybe use `?APP` instead of `erlang_ls`.

jfacorro:
@onno-vos-dev Could you rebase on the current `master`? I just merged [this PR](https://github.com/erlang-ls/erlang_ls/pull/543) which supposedly will solve the `Checkout` failures we are seeing when re-running jobs.

src/erlang_ls.erl
onno-vos-dev: Squashed it into the main commit

README.md
```diff
@@ -111,11 +111,17 @@ Compile the project:
...
-| --log-dir DIR           | Directory where logs will be written. When not provided no logs are generated.                  |
+``` shell
+Usage: Erlang LS [-v] [-t [<transport>]] [-p [<port>]] [-d [<log_dir>]]
+                 [-l [<log_level>]] [<port_old>]
+
+  -v, --version    Print the current version of Erlang LS
+  -t, --transport  Specifies the transport the server will use for the
+                   connection with the client. [default: tcp]
+  -p, --port       Used when the transport is tcp. [default: 10000]
+  -d, --log-dir    Directory where logs will be written. [default: <user_log>/erlang_ls/log]
```
robertoaloi: Not sure about the default value. Is there really a `log` as the last section in the path?

onno-vos-dev: It is what getopt and erl are both telling me :) 

```erlang
Eshell V10.3.5.7  (abort with ^G)
1> filename:basedir(user_log, "erlang_ls").
"/home/onnovos/.cache/erlang_ls/log"
```

```shell
 [OTP: 21.3.8.11][iex: 1.9.0] src/github/erlang_ls git:(476-use-getopt-for-cli-arg-parsing)
erlang_ls -h
Usage: Erlang LS [-v] [-t [<transport>]] [-p [<port>]] [-d [<log_dir>]]
                 [-l [<log_level>]] [<port_old>]

  -v, --version    Print the current version of Erlang LS
  -t, --transport  Specifies the transport the server will use for the
                   connection with the client. [default: tcp]
  -p, --port       Used when the transport is tcp. [default: 10000]
  -d, --log-dir    Directory where logs will be written. [default:
                   /home/onnovos/.cache/erlang_ls/log]
  -l, --log-level  The log level that should be used. [default: info]
  <port_old>       Port provided as integer for backwards compatibility
                   reasons
```

onno-vos-dev: I've taken the output from `erlang_ls -h` and "masked" out my path with `<user_log>` which seemed to make most sense as a "variable". If you have another suggestion I'm happy to change it :)

jfacorro: The value is platform dependent. 

This is in Mac OS X for example:
```
Eshell V10.3.5.6  (abort with ^G)
1> filename:basedir(user_log, "erlang_ls").
"/Users/juan.facorro/Library/Logs/erlang_ls"
2>
```
We could show the actual call to `filename:basedir(user_log, "erlang_ls")` or mention that it's platform dependent and to check the documentation.
```
-d, --log-dir    Directory where logs will be written. [default:
                 filename:basedir(user_log, "erlang_ls")]
```
or
```
-d, --log-dir    Directory where logs will be written. [default:
                 platform dependent, please check the documentation]
```

onno-vos-dev: I'll change it to the first example you give. getopts automatically expands this so didn't realize that also the suffix is apparently different (missing log). 

Push coming up 👍

robertoaloi: Thanks @jfacorro for the clarification and @onno-vos-dev for the fix!

src/erlang_ls.erl
```diff
@@ -22,38 +23,100 @@ main(Args) ->
...
+parse_args(Args) ->
+  case getopt:parse(opt_spec_list(), Args) of
+    {ok, {[version | _], _BadArgs}} ->
+      print_version(),
+      halt(1);
+    {ok, {ParsedArgs, _BadArgs}} ->
+      %% Handle backwards compatibility of clients that do not provide args but
+      %% only supply an integer. Assume it to be the port and remove the default
+      %% that we pick for the port.
+      ValidArgs = case lists:keyfind(port_old, 1, ParsedArgs) of
```
robertoaloi: We can address this separately, but do we really need to keep backward compatibility for this right now? @jfacorro

jfacorro: This was to support previous versions of the `lsp-mode` client. If we want to force people to update to the latest, then we can remove this support.

PR #547, opened 2020-03-12
Do not make large node groups

dgud:
By default erlang nodes will try to make a fully connected net of all
erlang nodes discovered. We don't want that 'erlang_ls' nodes interfer
with other testing nodes on the same machine.

Also avoids confusing mnesia when several different schemas might
exist in the same group of nodes. Mnesia wants the same schema on all
nodes in the group.

An alternative solution is to make the nodes hidden, but that
might make connecting and inspecting the nodes harder.

If we later want to create a mesh of nodes this needs to be
revised, maybe use global_groups?

### Description

Fixes # .

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -29,6 +29,8 @@
 
 {minimum_otp_vsn, "21.0"}.
 
+{escript_emu_args, "%%! -connect_all false \n" }.
+
 {profiles, [ { debug
              , [ { deps
                    %% The upstream version of redbug does not support
@@ -41,7 +43,7 @@
                    ]
                  }
                , { escript_emu_args
-                 , "%%! -erlang_ls logging_enabled true \n"
+                 , "%%! -connect_all false -erlang_ls logging_enabled true \n"
                  }
                  %% For some reason we need to explicitly include redbug,
                  %% even if it already is a dependency.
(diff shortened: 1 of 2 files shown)


rebar.config
```diff
@@ -29,6 +29,8 @@
 
 {minimum_otp_vsn, "21.0"}.
 
+{escript_emu_args, "%%! -connect_all false \n" }.
```
robertoaloi: I don't have a problem with this, but how come this is an issue for you? Isn't a node "added" only at the first interaction (e.g. an explicit ping, or a remote spawn)? I imagine the language server node would not try to connect to other nodes. But maybe in your system the other nodes are trying to connect to it somehow?

dgud:
I have a problem with starting a new emacs with a file in a project, for example in otp, 
then 'erlang_ls' in my other emacs process exists/dies even though it doesn't have any files opened in the otp project.

I.e. I run one emacs for erlang_ls editing and one for otp editing and pretty often starting one
of them causes the other to loose it's connection to erlang_ls.

I come to think about this, but it didn't solve the problem but I still believe it's a good idea.
For example using observer to connect to one erlang_ls process will make global create a fully connected mesh.

PR #548, opened 2020-03-12
[#110] Add auto-completion for atoms

### LINKED ISSUE #110: Auto-complete for atom names

jfacorro:
### Description

This PR introduces an adaptation of the `erl_syntax_lib:fold/3` function, so that only some of the subtrees are processed for specific types of syntax tree nodes.

Fixes #110.

### DIFF
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -131,9 +131,7 @@ find_attribute_tokens(_) ->
 %% @doc Find points of interest in a spec attribute.
 -spec find_spec_points_of_interest(tree()) -> [poi()].
 find_spec_points_of_interest(Tree) ->
-  Fun = fun do_find_spec_points_of_interest/2,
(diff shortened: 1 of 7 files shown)


src/els_parser.erl
```diff
@@ -402,3 +409,59 @@ poi(Pos, Kind, Id) ->
...
+  S.
+
+-spec fold_2(fun((tree(), term()) -> term()), term(), [tree()]) ->
+  term().
+fold_2(F, S, [T | Ts]) ->
+  fold_2(F, fold(F, S, T), Ts);
+fold_2(_, S, []) ->
+  S.
+
+-spec subtrees(tree(), atom()) -> [[tree()]].
```
jfacorro: The most important change in this PR is in this function which filters out the nodes that might contains a syntax tree node of type `atom` which we don't want as a POI.

PR #552, opened 2020-03-13
[#485] Goto definition on an atom goes to that module if it exists.

### LINKED ISSUE #485: Goto module when requesting definition of atom
Isolated atoms are very often used to refer to a module (e.g. `erlang:apply/3`) so it would be a nice feature to be able to navigate to the module using the atom.

When placing the cursor over an atom and running the command to go to its definition, navigate to the module with the atom's name 

N/A.

**Additional context**
N/A.

bolmar:
### Description

Goto definition on an atom goes to that module if it exists.

Fixes #485 .


src/els_code_navigation.erl
```diff
@@ -36,9 +36,13 @@ goto_definition( Uri
...
                       Kind =:= export_entry ->
   find(Uri, function, {F, A});
-goto_definition(_Uri, #{ kind := behaviour, id := Behaviour }) ->
-  case els_utils:find_module(Behaviour) of
-    {ok, Uri}      -> find(Uri, module, Behaviour);
+goto_definition( _Uri
+               , #{ kind := Kind, id := Module }
+               ) when Kind =:= atom;
+                      Kind =:= behaviour;
+                      Kind =:= module ->
```
jfacorro: Does it make sense to add `module` here? Won't this be the case where the user is trying to get the definition of the module from the `-module(foo).` attribute?

PR #551, opened 2020-03-12
[#550] Add OTP 23 pipeline

robertoaloi:
### Description

Fixes #550 .

PR #556, opened 2020-03-13
[#106] Auto-complete built-in functions and types

### LINKED ISSUE #106: Code completion for BIF names

jfacorro:
### Description

Implements auto-completion for Erlang built-in functions and types. 

There's one caveat which is that the built-in functions are resolved dynamically based on the Erlang/OTP release in which **Erlang LS is running** and not the Erlang/OTP release specified in the **erlang_ls.config**. 

I considered trying to figure out what the were BIFs by parsing the clauses from source code in `erl_internal:bif/2`, but decided against it. It seemed overkill given that even though BIFs get added, they don't change that much.

A possible improvement would be to resolve them only once when the provider is initialized instead of resolving on every completion.

The list of built-in type is a hardcoded list taken from the information [here](https://erlang.org/doc/reference_manual/typespec.html).

Fixes #106.

### DIFF
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -259,10 +259,9 @@ attribute(Tree) ->
       end;
     {record, {Record, Fields}} ->
       [poi(Pos, record, Record, Fields)];
-    {type, {type, {Type, _, Args}}} ->
+    {Name, {Name, {Type, _, Args}}} when Name =:= type;
+                                         Name =:= opaque ->
       [poi(Pos, type_definition, {Type, length(Args)}, type_args(Args))];
-    {opaque, {opaque, {Type, _, Args}}} ->
-      [poi(Pos, type_definition, {Type, length(Args)})];
     _ ->
       []
   catch throw:syntax_error ->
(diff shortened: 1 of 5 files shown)


src/els_parser.erl
```diff
@@ -259,10 +259,9 @@ attribute(Tree) ->
       end;
     {record, {Record, Fields}} ->
       [poi(Pos, record, Record, Fields)];
-    {type, {type, {Type, _, Args}}} ->
+    {Name, {Name, {Type, _, Args}}} when Name =:= type;
```
jfacorro: This is a fix for a bug I found while getting type completions from the `dict` module.

test/els_fake_stdio.erl
```diff
@@ -87,6 +87,8 @@ handle_request({get_line, _Encoding, _Prompt}, State0) ->
     _ ->
       {noreply, State0}
   end;
+handle_request({get_chars, Encoding, _Prompt, Count}, State) ->
```
jfacorro: This is a fix for a bug found in #551.

robertoaloi:
I would not worry about the node inconsistency. We will introduce an extra node soon...

PR #566, opened 2020-03-14
Use open_port to spawn epmd

filmor:
This ensures that spaces in the path are treated correctly.

### Description

Exchange `os:cmd` for a call to `open_port({spawn_executable ...`, adjusted from https://stackoverflow.com/a/2249387/542190.

Fixes https://github.com/erlang-ls/vscode/issues/24.

### DIFF
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -136,7 +136,7 @@ transaction(F) ->
 
 -spec start_epmd() -> ok.
 start_epmd() ->
-  [] = os:cmd(epmd_path() ++ " -daemon"),
+  0 = els_utils:cmd(epmd_path(), ["-daemon"]),
   ok.
 
 -spec epmd_path() -> string().
--- a/src/els_utils.erl
+++ b/src/els_utils.erl
@@ -8,6 +8,7 @@
         , project_relative/1
         , resolve_paths/3
         , halt/1
+        , cmd/2
         ]).
 
 -include("erlang_ls.hrl").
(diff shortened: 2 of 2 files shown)


robertoaloi:
@richcarl Your answer is from 2010. Do you know if this is still the best to get this fix? I'd prefer not to have to fork the code from `os:cmd`.

richcarl:
> @richcarl Your answer is from 2010. Do you know if this is still the best to get this fix? I'd prefer not to have to fork the code from `os:cmd`.

spawn_executable is much safer than the general cmd, but as far as I know the only way of using it is by rolling your own. This piece of code is however way simpler than the actual os:cmd implementation, so not much of a worry.

filmor:
Fixed the spec. This is not a "fork" of the original `os:cmd`, I just simplified the version given in the SO answer as just to call `epmd -daemon` we need a lot less processing (in particular, the output is completely irrelevant as it is empty).

src/els_db.erl
```diff
@@ -12,7 +12,8 @@
         , stop/0
         , tables/0
         , transaction/1
-        , write/1
+        , write/1,
```
robertoaloi: The project currently uses a comma-first indentation. Things may change in the future, for now please follow the guidelines at https://github.com/erlang-ls/erlang_ls/blob/master/.github/CONTRIBUTING.md

robertoaloi: Also, why exporting it?

filmor: Ah, sorry, I did this just for testing, that's also why I didn't adhere to the guidelines. Will fix.

robertoaloi:
Thanks!

PR #569, opened 2020-03-15
dgud/config search apps dirs

dgud:
This PR replaces #501 or should be used as based.

Allows more projects to work out of the box, without using an erlang_ls.config file.
This PR is in my opinion a better default setup, but I don't know the thoughts you
have on how things should work, so I might have completely misunderstood things.

Recursively search for AppsDirs if not directly found, to find indexed applications,
this allows meta-projects where several applications should be included.

Adds all subdirs to search_path for lazy inclusion of other dirs/files in the project.

Removes tests dir which adds a lot to be indexed, when using otp src as project for instance.

### DIFF
--- a/src/els_config.erl
+++ b/src/els_config.erl
@@ -111,7 +111,8 @@ do_initialize(RootUri, Capabilities, {ConfigPath, Config}) ->
   ok = set(plt_path       , DialyzerPltPath),
   ok = set(code_reload    , CodeReload),
   %% Calculated from the above
-  ok = set(apps_paths     , project_paths(RootPath, AppsDirs, false)),
+  AppsPaths = apps_paths(RootPath, AppsDirs),
+  ok = set(apps_paths     , AppsPaths),
   ok = set(deps_paths     , project_paths(RootPath, DepsDirs, false)),
   ok = set(include_paths  , include_paths(RootPath, IncludeDirs, false)),
   ok = set(otp_paths      , otp_paths(OtpPath, false) -- ExcludePaths),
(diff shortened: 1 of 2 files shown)


robertoaloi:
I think we should split this PR into multiple ones and change one thing at the time, to speed-up the review process. In general, I'd prefer this module (that needs some refactoring) to do as little magic as possible. We want to introduce the concept of a project-type (e.g. rebar3) and the "adapter" modules implementing the various project-types should contain all the logic necessary to make things work out of the box.
Not sure about removing the "test" directory, it's something we rely upon.

dgud:
I can making the first and the last commits as separate PR's,
the other have no purpose without each other.
Or do you have other ideas?

Though removing test dir can not be done without adding it to search dirs.

Loading up OTP as project makes an erlang_ls process 2GB which I think is a bit excessive.
When directories are indexed, that is always done recursively so that all dirs below test are also
indexed which includes compiler and dialyzer test data, many erlang files that uses the same module names as other files.

Since everything not found is looked up via search path, I don't understand the problem that you rely on that, it will work, just indexed lazily and once it's in the database it will be used.
But that is me not understanding how things work or are supposed to work.

dgud:
This first commit is wrong and will be removed, it´s the other way around,  we should pick the first one
as was done before. But that is gives me the wrong module most often.
We need another way of solving this.

dgud:
I have split up the PR.

dgud:
Hmm test failing, I don't understand why and works fine locally.

alanz:
It could be https://github.com/erlang-ls/erlang_ls/issues/646, it is PropEr that is failing.

PR #571, opened 2020-03-15
[#560] Remove separation between debug and default mode

### LINKED ISSUE #560: Get rid of `debug` mode
Now that logging is configurable, there is no real need for a `debug` mode. I would get rid of the whole concept and always include redbug as a dependency. If we decide to do so, we should update all clients and docs that refer to a `debug` mode.

robertoaloi:
### Description

* Remove `debug` mode, keeping the alias
* Enable logging by default

Fixes #560 .

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -10,7 +10,9 @@
 {deps, [ {ranch,         "1.7.1"}
        , {jsx,           "2.9.0"}
        , {cowlib,        "2.3.0"}
-       , {redbug,        "1.2.1"}
+         %% The latest tagged version of redbug does not support
+         %% debugging escriptized code, so fetching it from git.
+       , {redbug,        {git, "https://github.com/massemanet/redbug.git", {ref, "60fb6e17c164f153a792985a88ac5c67a2364220"}}}
        , {lager,         "3.6.8"}
        , {yamerl,        "0.7.0"}
        , {docsh,         "0.7.2"}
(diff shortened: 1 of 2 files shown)


rebar.config
```diff
@@ -29,21 +31,14 @@
...
-                 }
-               , { escript_emu_args
-                 , "%%! -connect_all false -erlang_ls logging_enabled true \n"
-                 }
-                 %% For some reason we need to explicitly include redbug,
-                 %% even if it already is a dependency.
-               , {escript_incl_apps, [redbug]}
-               ]
-             }
+%% Keeping the debug profile as an alias, since many clients were
```
jfacorro: In what ways were clients relying on the `debug` profile? Because of the path to the escript?

src/erlang_ls.erl
```diff
@@ -111,7 +111,6 @@ set(transport, Name) ->
 set(port, Port) ->
   application:set_env(?APP, port, Port);
 set(log_dir, Dir) ->
-  application:set_env(?APP, logging_enabled, true),
   application:set_env(?APP, log_dir, Dir);
 set(log_level, Level) ->
   application:set_env(?APP, log_level, Level);
```
jfacorro: I just noticed that the rewrite of parsing the options using `getopt` removed the enabling of logging when a `log_level` is provided. I guess it's saner to just enable it in one place or have it enabled by default.

rebar.config
robertoaloi: Yes, exactly.

src/erlang_ls.erl
robertoaloi: Sure, feel free to raise an issue.

src/erlang_ls.erl
```diff
@@ -125,7 +124,7 @@ set(port_old, Port) ->
 
 -spec lager_config() -> ok.
 lager_config() ->
-  LoggingEnabled = application:get_env(?APP, logging_enabled, false),
+  LoggingEnabled = application:get_env(?APP, logging_enabled, true),
```
jfacorro: The value in `erlang_ls.app.src` is `false`, though. Do we want this to be `true` or `false` by default?

alanz: I would prefer it to be true, in general. It makes it easier to ask people for extra info when they report problems, to not have to first take them through a process of getting logging enabled.

robertoaloi: That was my intention, but missed the `.app.src` entry. Will follow-up.

jfacorro: I used to be of a different opinion, but I now think that logging enabled is a good default.

PR #572, opened 2020-03-15
Clarify path to default config file

Vagabond:
### Description

Adds some clarification to a place where I got tripped up.


robertoaloi:
Thanks!

PR #578, opened 2020-03-17
[#486] Find references for types

### LINKED ISSUE #368: Tighten up id field in poi()
At the moment `poi()` is defined as

```erlang
-type poi()       :: #{ kind  := poi_kind()
                      , id    := any()
                      , data  := any()
                      , range := poi_range()
                      }.
```

In practice we only ever use `{F,A}` or `{M,F,A}` for the `id` field.

Define a type for `id` capturing those options only.

### LINKED ISSUE #486: Find references for type
It's sometimes useful to find all places where a type is being used.

Show a list of references of the type when placing the cursor in the usage or definition of the type.

N/A.

**Additional context**
N/A.

jfacorro:
### Description

Implements finding references for types. Tightens the `id` field type definition for `poi()`s.

Fixes #486. Fixes #368.

### DIFF
--- a/src/els_document_highlight_provider.erl
+++ b/src/els_document_highlight_provider.erl
@@ -45,20 +45,17 @@ find_highlights(Uri, #{ kind := Kind
                       }) when Kind =:= application;
                               Kind =:= implicit_fun;
                               Kind =:= function;
-                              Kind =:= variable;
                               Kind =:= export_entry ->
   Key = case Id of
           {F, A}    -> {els_uri:module(Uri), F, A};
-          {M, F, A} -> {M, F, A};
-          V         -> {els_uri:module(Uri), V, 0}
+          {M, F, A} -> {M, F, A}
         end,
-  case els_dt_references:find_by_id(Key) of
+  case els_dt_references:find_by_id(Kind, Key) of
     {ok, []} ->
       null;
     {ok, Refs} ->
       R = [ document_highlight(R) ||
-            #{uri := U, range := R} <- ordsets:to_list(Refs)
-              , Uri == U
+            #{uri := U, range := R} <- ordsets:to_list(Refs), Uri =:= U
           ],
       R
   end;
--- a/src/els_dt_references.erl
+++ b/src/els_dt_references.erl
@@ -20,8 +20,8 @@
 -export([ delete_by_uri/1
         , find_all/0
         , find_by/1
-        , find_by_id/1
-        , insert/1
+        , find_by_id/2
+        , insert/2
         ]).
 
 %%==============================================================================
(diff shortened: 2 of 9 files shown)


src/els_document_highlight_provider.erl
```diff
@@ -45,20 +45,17 @@ find_highlights(Uri, #{ kind := Kind
                       }) when Kind =:= application;
                               Kind =:= implicit_fun;
                               Kind =:= function;
-                              Kind =:= variable;
```
jfacorro: `variable` was removed since they are not indexed and included in the references table.

src/els_dt_references.erl
```diff
@@ -96,12 +97,31 @@ find_all() ->
...
+  InternalId = {kind_to_category(Kind), Id},
+  Pattern = #els_dt_references{id = InternalId, _ = '_'},
   find_by(Pattern).
 
 -spec find_by(tuple()) -> {ok, [item()]}.
 find_by(Pattern) ->
   {ok, Items} = els_db:match(Pattern),
   {ok, [to_item(Item) || Item <- Items]}.
+
+-spec kind_to_category(poi_kind()) -> function | type | macro | record.
```
jfacorro: The idea of this function is to include in the guards all the supported POIs for which it is possible to find a reference.

src/els_document_highlight_provider.erl
alanz: At some point I would like to see highlights of local variables, according to the language scoping rules. But I guess that is not an indexing feature, it can come from the compiler for the specific module.

src/els_document_highlight_provider.erl
```diff
@@ -45,20 +45,17 @@ find_highlights(Uri, #{ kind := Kind
...
                               Kind =:= export_entry ->
   Key = case Id of
           {F, A}    -> {els_uri:module(Uri), F, A};
-          {M, F, A} -> {M, F, A};
-          V         -> {els_uri:module(Uri), V, 0}
+          {M, F, A} -> {M, F, A}
```
alanz: what about matching `string () | atom ()`?  Or is that excluded based on the guard clauses?

jfacorro: Exactly, those cases are excluded because of the filtering done in the guard.

src/els_document_highlight_provider.erl
jfacorro: I think it is possible to make this work through reference, which we could use also to go to the definition of a variable, but this would require some more complex analysis during parsing.

alanz: I think it will add a lot of bulk to the index, and it might be simpler to just compute it as needed. It will always be in a local scope, and erlang functions tend to be quite small.  And no global variables.

jfacorro: That makes a lot more sense 😄.

Issue #400, opened 2020-01-18
Correctly handle RootUri during initialization


robertoaloi:
**Describe the bug**

According to the LSP protocol, if no folders are open, the `rootUri` parameter during initialization should be set to `null`. Whilst most editors always set the proper `rootUri` (e.g. Emacs does it via _projectile_),  Sublime Text 3 (aka ST3) editor does not. If `null` is sent to the server, the server does not respond to the initialization message.

**To Reproduce**
Open a `.erl` file in a newly configured Sublime editor.

**Context**
 - `erlang_ls` version (tag/sha): latest
 - Editor used: ST3
 - LSP client used: LSP

**Notes**
We should also log the initialization message from the server side.

alanz:
What should the behaviour be?

I would propose  separating initialisation into two parts, one which just acks the message, and another which kicks off indexing, set the various paths, loading config files and the rest, which we trigger as soon as we get a URI. 

The problem with that is that the server is then required to have project root determination logic, based on the URI of the first file it is asked to open.

alanz:
And [this](https://github.com/microsoft/language-server-protocol/issues/176#issuecomment-278296141) comment may be relevant

> The root Uri basically determines the workspace you have open. In 99% of the cases documents will be inside that workspace but there is no enforcement of this and a language server should handle documents outside the root gracefully. All language servers developed by the VS Code team do so. If we receive a document (for example a json file) from outside the workspace we still provide basic language features for them but can't respect any workspace specific setting since we can't load any.

> At the end it is up to the implementer of the LS to handle this correctly.

jfacorro:
> I would propose separating initialisation into two parts, one which just acks the message, and another which kicks off indexing, set the various paths, loading config files and the rest, which we trigger as soon as we get a URI.
> 
> The problem with that is that the server is then required to have project root determination logic, based on the URI of the first file it is asked to open.

I like this approach but I fear that things will get too complex, because:
1.  As you mention, we would need to implement the logic to determine what is the root URI based on the first file URI we get. The result might be surprising to the user.
2. Until we have a rootURI we don't create the DB and without the DB we fail to provide any functionality.

The alternative approach I was thinking about to support a `null` value for `rootUri` on initializationa are the following:

1.  Use the current working directory (i.e. `file:get_cwd/0`) from where the `erlang_ls` escript is started.
2. Use a temporary directory (e.g. under `/tmp` in *nix).
[...]

PR #584, opened 2020-03-17
[#400] Use the current working directory as rootUri when missing

jfacorro:
### Description

Use the current working directory as `rootUri` when missing.

After trying different things and giving it some thought I decided to go for the simplest approach. Please check the comments in the related issue #400 and share your thoughts.

### DIFF
--- a/src/els_methods.erl
+++ b/src/els_methods.erl
@@ -120,11 +120,17 @@ method_to_function_name(Method) ->
 
 -spec initialize(params(), state()) -> result().
 initialize(Params, State) ->
-  #{ <<"rootUri">> := RootUri
+  #{ <<"rootUri">> := RootUri0
      %% TODO: Use ClientCapabilities in completion_provider
      %%       to verify when Context is present
    , <<"capabilities">> := Capabilities
    } = Params,
+  RootUri = case RootUri0 of
+              null ->
+                {ok, Cwd} = file:get_cwd(),
+                els_uri:uri(unicode:characters_to_binary(Cwd));
+              _ -> RootUri0
+            end,
   InitOptions = maps:get(<<"initializationOptions">>, Params, #{}),
   ok = els_config:initialize(RootUri, Capabilities, InitOptions),
   DbDir = application:get_env(erlang_ls, db_dir, default_db_dir()),


alanz:
This behaviour is only a problem with specific clients.

Can we validate with them that this approach does in fact do something that does not cause surprise?

jfacorro:
@distortedsignal I think you are Sublime Text user. Could you share your view on what you would expect  when opening an isolated Erlang file in Sublime Text? 

If you have time, please try out this branch and provide feedback on the behaviour you get.

There's some discussion in the related issue (#400) that could be relevant to understand the context.

src/els_methods.erl
```diff
@@ -120,11 +120,17 @@ method_to_function_name(Method) ->
 
 -spec initialize(params(), state()) -> result().
 initialize(Params, State) ->
-  #{ <<"rootUri">> := RootUri
+  #{ <<"rootUri">> := RootUri0
      %% TODO: Use ClientCapabilities in completion_provider
      %%       to verify when Context is present
    , <<"capabilities">> := Capabilities
    } = Params,
+  RootUri = case RootUri0 of
```
robertoaloi: Nit: this logic could sit in a separate function (e.g. `default_root_uri/0`)

robertoaloi:
I think we can proceed with this (some feedback from Sublime users would be appreciated).
In general, I believe we should start introducing the concept of a `workspace` in the server, something defined by the protocol that we have been sort of ignoring.
The `root_uri`, the way I understand it, represents a `workspace`, and a workspace can be made of multiple folders. Not sure yet how to combine this with the current way of configuring _projects_.

distortedsignal:
From an ST perspective, I would expect this to either be the file-local path (as in this PR) or the user's root directory (for me, on MacOS, `/Users/tom`, on Windows, `C:\Users\tom`, etc.). I don't think either of these is wrong, and I think this is a fine first-step in the right direction for the server.

The kind-of-tricky thing about Sublime Text is that there are two different "modes" that it can operate in - "I'm in a project", where you have a folder you're anchored to (which I think the server already handles?), and "I'm not in a project", where files can come from basically anywhere on disk (or be raw text buffers). I almost never work outside of the "I'm in a project" mode, so I might not be the right person to ask about this.

jfacorro:
@distortedsignal Thank you for your input! 😄 

> The kind-of-tricky thing about Sublime Text is that there are two different "modes" that it can operate in [...]

This is something I suspected but wasn't completely sure about.

> [...] - "I'm in a project", where you have a folder you're anchored to (which I think the server already handles?), [...]

Yes, `erlang_ls` handles this mode already. 👍 

> [...] and "I'm not in a project", where files can come from basically anywhere on disk (or be raw text buffers).

This is the case we are trying to address in this PR (as you already know). I think the PR's approach is good enough in the sense that if the users' mental model is "I'm not in a project" I guess the expectation is to not have the context of that project (i.e. being able to go to the definition of a file that has not been opened).

PR #585, opened 2020-03-17
Add a code lens to show the basic erlang_ls config for a file.

### LINKED ISSUE #574: els_execute_command_provider:strip_server_prefix should not fail
**Describe the bug**

The `els_execute_command_provider:strip_server_prefix` function is currently

```erlang
strip_server_prefix(PrefixedCommand) ->
  [_, Command] = binary:split(PrefixedCommand, <<":">>),
  Command.
```
If the `PrefixedCommand` does not in fact have a `:` in it, the function crashes.

**Expected behavior**

`strip_server_prefix(<<"foo">>)` returns `<<"foo">>`

**Context**
 - `erlang_ls` version (tag/sha): aa34eb1bedd91171280883c1e718efcca0cbdf8c
 - Editor used:
 - LSP client used:

alanz:
### Description

Provide a code lens that identifies the root directory used for a given file, which is clickable, and if clicked gives the erlang ls version, the root directory, and the full path of the config file used.

Fixes #574, in passing


alanz:
This is a screenshot of the lens displayed at the top, and the result of clicking on it at the bottom

Screenshot 2020-03-17 at 16 54 42

test/els_execute_command_SUITE.erl
```diff
@@ -0,0 +1,100 @@
...
+-spec erlang_ls_info(config()) -> ok.
+erlang_ls_info(Config) ->
+  Uri = ?config(code_navigation_uri, Config),
+  PrefixedCommand
+    = els_execute_command_provider:add_server_prefix(<<"info">>),
+  #{result := Result}
+    = els_client:workspace_executecommand(PrefixedCommand, [#{uri => Uri}]),
+  Expected = [],
+  ?assertEqual(Expected, Result),
+  timer:sleep(10), %% otherwise the stdio client does not get the Notification
```
jfacorro: I have done this myself sometimes, but I think we should use a similar approach to the one in `els_test_utils:wait_for/2` but instead of waiting for a message we can provide a function. This is mainly to avoid flakiness in the tests, which can happen if we run the builds in slower machines.

The approach I'm reffering to would be definind an `els_test_utils:wait_for_fun/3` like so:

```erlang
-spec wait_for_fun(fun(), non_neg_integer(), non_neg_integer()) -> {ok, any()} | timeout.
wait_for_fun(_CheckFun, _WaitTime, 0) ->
  timeout;
wait_for_fun(CheckFun, WaitTime, Retries) ->
  case CheckFun() of
    {true, Value} ->
      {ok, Value};
    false ->
      timer:sleep(WaitTime),
      wait_for_fun(CheckFun, WaitTime, Retries - 1)
  end.
```

src/els_execute_command_provider.erl
```diff
@@ -46,6 +47,23 @@ execute_command(<<"replace-lines">>
             },
   els_server:send_request(Method, Params),
   [];
+execute_command(<<"info">>
+               , [#{ <<"uri">>   := _Uri }] = _Arguments) ->
+  {ok, Version} = application:get_key(?APP, vsn),
+  BinVersion = list_to_binary(Version),
+  Root = filename:basename(els_uri:path(els_config:get(root_uri))),
+  ConfigPath = list_to_binary(els_config:get(config_path)),
```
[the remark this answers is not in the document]
alanz: The method in els_config calls 
```erlang
  Value = maps:get(Key, State, undefined),
```
So it returns `undefined`.  Which can blow up when constructing the message, I guess.

Will add a case on it.

src/els_execute_command_provider.erl
```diff
@@ -46,6 +47,23 @@ execute_command(<<"replace-lines">>
...
+  lager:info("execute_command info: [ConfigPath=~p]", [ConfigPath]),
```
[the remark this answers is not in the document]
alanz: removing

src/els_execute_command_provider.erl
```diff
@@ -46,6 +47,25 @@ execute_command(<<"replace-lines">>
...
   els_server:send_request(Method, Params),
   [];
+execute_command(<<"info">>
+               , [#{ <<"uri">>   := _Uri }] = _Arguments) ->
+  {ok, Version} = application:get_key(?APP, vsn),
+  BinVersion = list_to_binary(Version),
+  Root = filename:basename(els_uri:path(els_config:get(root_uri))),
+  ConfigPath = case els_config:get(config_path) of
+                 undefined -> <<"undefined">>;
+                 Path -> list_to_binary(Path)
```
jfacorro: 😄 There's now an `els_utils:to_binary/1` function.

alanz: Not in current master, that I can see.

https://github.com/erlang-ls/erlang_ls/blob/99e624c843a0fe50868a6c9614dd1dae08ce6bb3/src/els_utils.erl#L3-L12

src/els_client.erl
```diff
@@ -45,6 +46,9 @@
         , start_link/2
         , stop/0
         , workspace_symbol/1
+        , workspace_executecommand/2
+
+        , get_notifications/0
```
robertoaloi: This function seems only to be used for testing purposes. In that case, I would prefer to not export it and to use this instead:

https://erlang.org/doc/man/sys.html#get_state-1

test/els_execute_command_SUITE.erl
```diff
@@ -0,0 +1,100 @@
...
+  #{result := Result}
+    = els_client:workspace_executecommand(PrefixedCommand, [#{uri => Uri}]),
+  Expected = [],
+  ?assertEqual(Expected, Result),
+  CheckFun = fun() -> case els_client:get_notifications() of
+                        [] -> false;
+                        Notifications -> {true, Notifications}
+                      end
+             end,
+  {ok, [Notification]} = els_test_utils:wait_for_fun(CheckFun, 10, 3),
```
robertoaloi: It looks like notifications are only introduced for testing purposes, to understand when something was called. I think this could be achieved differently, by mocking, passing-through and message passing. I can show you an example, if you need.

src/els_config.erl
```diff
@@ -74,8 +74,8 @@ initialize(RootUri, Capabilities, InitOptions) ->
   Config = consult_config(config_paths(RootPath, InitOptions)),
   do_initialize(RootUri, Capabilities, Config).
 
--spec do_initialize(uri(), map(), map()) -> ok.
-do_initialize(RootUri, Capabilities, Config) ->
+-spec do_initialize(uri(), map(), {undefined|path(), map()}) -> ok.
```
robertoaloi: Things look a bit weird now. The `consult_config/1` function now returns a tuple, but on line 74 we match it as a `Config`. Why do we need this change?

test/els_execute_command_SUITE.erl
alanz: Well, the core functionality of the feature being added is to send that notification.  And the client is intended to be an interface to the server. So there *should* be a way to actually receive the notifications, by something using the client.

If you feel strongly and can show an alternative based on mocking I will change it, but not convinced it is the right approach.

robertoaloi: I will not push for this, I think it can eventually be cleaned up later. From my perspective, the only blocker for this PR is to have a configurable list of code lenses for the project via the config file (empty by default for now, until we have a bunch). But I am also ok for merging this and addressing that in a follow-up before next release. @jfacorro opinions?

src/els_config.erl
alanz: The tuple returns the path of the file it was loaded from, so we can return it in the lens result.  `do_initialize` expects the tuple, and stores the path.  So perhaps the variable name needs changing, but it *is* the config, just includes where it came from.

src/els_client.erl
alanz: > This function seems only to be used for testing purposes.

What is the purpose of the client?  Either it is for testing, in which case this function makes sense, or it is to actually interact with a server, in which case receiving notifications seems to be an obvious requirement.

What am I missing?

test/els_execute_command_SUITE.erl
alanz: Ok. And given it is a feature for debugging config files, I would suggest we make it a configuration parameter from the client rather, since we may not be adding the feature to the config file we expect to be used.

src/els_client.erl
robertoaloi: > They are provided for convenience, allowing developers to avoid having to create their own state extraction functions

That's the part I was referring to.

robertoaloi:
@alanz I am going to merge this PR as-is and add a mechanism for configuring code lenses in a separate PR.

src/els_client.erl
alanz: ok, but the `get_notifications` function returns all notifications received, and empties the list in the state, which `sys:get_state` does not do. This allows a multi-step test to assert on notifications for each step.  This gets more important when you are processing more complex scenarios as lenses and code actions get more capable.

PR #591, opened 2020-03-19
[#306] Use unicode module

### LINKED ISSUE #306: Make sure code handles unicode properly
Make sure we are using `unicode:characters_to_binary` instead of `list_to_binary` in all the relevant places.

See https://github.com/erlang-ls/erlang_ls/pull/298/files#r355965608

jfacorro:
### Description

Use unicode module instead of `[io]list_to_binary/1` and `binary_to_list/1`

Fixes #306.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest]
-        otp-version: [21.3, 22.2, 23]
+        otp-version: [21.3, 22.2]
     runs-on: ${{ matrix.platform }}
     container:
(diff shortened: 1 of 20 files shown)


jfacorro:
> Is it possible to set up elvis or similar to warn about using `binary_to_list` for future code changes?

Good idea 😄. Yes, there's the [No Call](https://github.com/inaka/elvis_core/wiki/Rules#no-call) rule which allows to warn about calls to a specific `{M, F, A}` or `{F, A}`.

robertoaloi:
A good reading about this is: https://adoptingerlang.org/docs/development/hard_to_get_right/#handling-strings-in-erlang

PR #592, opened 2020-03-19
[#306] Handle non-unicode as well

### LINKED ISSUE #306: Make sure code handles unicode properly

jfacorro:
### Description

The previous PR assumed all string were always unicode, this is not the case. This PR also remove the `debug` target from the `Makefile` and adds an `install` target which can be evolved as needed (e.g. to support Windows).

Fixes #306.

### DIFF
--- a/src/els_code_action_provider.erl
+++ b/src/els_code_action_provider.erl
@@ -85,10 +85,10 @@ make_unused_variable_action(Uri, Range, UnusedVariable) ->
    } = Range,
   %% processing messages like "variable 'Foo' is unused"
   {ok, #{text := Bin}} = els_utils:lookup_document(Uri),
-  Line = unicode:characters_to_list(els_text:line(Bin, StartLine)),
+  Line = els_utils:to_list(els_text:line(Bin, StartLine)),
 
   {ok, Tokens, _} = erl_scan:string(Line, 1, [return, text]),
-  UnusedString = unicode:characters_to_list(UnusedVariable),
+  UnusedString = els_utils:to_list(UnusedVariable),
   Replace =
         fun(Tok) ->
             case Tok of
@@ -102,7 +102,7 @@ make_unused_variable_action(Uri, Range, UnusedVariable) ->
     [ replace_lines_action( Uri
                       , <<"Add '_' to '", UnusedVariable/binary, "'">>
                       , ?CODE_ACTION_KIND_QUICKFIX
-                      , unicode:characters_to_binary(UpdatedLine)
+                      , els_utils:to_binary(UpdatedLine)
                       , Range)].
 
 %%------------------------------------------------------------------------------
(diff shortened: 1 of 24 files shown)


src/els_code_action_provider.erl
```diff
@@ -85,10 +85,10 @@ make_unused_variable_action(Uri, Range, UnusedVariable) ->
    } = Range,
   %% processing messages like "variable 'Foo' is unused"
   {ok, #{text := Bin}} = els_utils:lookup_document(Uri),
-  Line = unicode:characters_to_list(els_text:line(Bin, StartLine)),
+  Line = els_utils:to_list(els_text:line(Bin, StartLine)),
```
alanz: Why would this not be unicode? It should be a line of source code we end up with.

src/els_code_action_provider.erl
```diff
@@ -85,10 +85,10 @@ make_unused_variable_action(Uri, Range, UnusedVariable) ->
...
 
   {ok, Tokens, _} = erl_scan:string(Line, 1, [return, text]),
-  UnusedString = unicode:characters_to_list(UnusedVariable),
+  UnusedString = els_utils:to_list(UnusedVariable),
```
alanz: ditto

src/els_code_action_provider.erl
```diff
@@ -102,7 +102,7 @@ make_unused_variable_action(Uri, Range, UnusedVariable) ->
     [ replace_lines_action( Uri
                       , <<"Add '_' to '", UnusedVariable/binary, "'">>
                       , ?CODE_ACTION_KIND_QUICKFIX
-                      , unicode:characters_to_binary(UpdatedLine)
+                      , els_utils:to_binary(UpdatedLine)
```
alanz: ditto

src/els_code_action_provider.erl
jfacorro: The line from the source file could be using some other enconding, right?

In general I replaced every usage of `unicode:characters_to_[list|binary]/1` for this function regardless of the possibility of getting non-unicode in each replacement. It's simpler IMO to have the general rule of not using list_to_binary or binary_to_list and use instead the functions in `els_utils`.

robertoaloi:
Can we please, as part of this same PR:

* Add the Elvis check for list_to_binary and binary_to_list
* Add a point to the GUIDELINES doc where we ask to be aware of Unicode when handling strings and reference this document? https://adoptingerlang.org/docs/development/hard_to_get_right/#handling-strings-in-erlang

src/els_code_action_provider.erl
alanz: ok, just wanted to understand that before diving into this in too much detail. Makes sense.  I guess I should have looked more closely to see that it is calling the new routine now, rather than some standard one.

PR #593, opened 2020-03-19
Implement file URI handling for Windows

filmor:
Windows `file://` URIs are quite a bit different than Unix `file://` in that they can contain a host (for UNC paths like `\\some_server\some_share` and that the `/` prefix has to be stripped in `file:///c:/some_path`. Also, the colon usually gets URI-encoded which needs to be undone.

### DIFF
--- a/src/els_uri.erl
+++ b/src/els_uri.erl
@@ -30,9 +30,35 @@ module(Uri) ->
   binary_to_atom(filename:basename(path(Uri), <<".erl">>), utf8).
 
 -spec path(uri()) -> path().
-path(<<"file://", Path/binary>>) ->
-  Path.
+path(Uri) ->
+  Uri1 = http_uri:decode(Uri),
+  #{ host := Host
+   , path := Path
+   , scheme := <<"file">>
+  } = uri_string:parse(Uri1),
+
+  case {is_windows(), is_local(Host)} of
+    {_, true} ->
+      string:slice(Path, 1);
+    {true, false} ->
+      <<"//", Host/binary, Path/binary>>;
+    {false, false} ->
+      error(badarg)
+  end.
 
 -spec uri(path()) -> uri().
 uri(Path) ->
-  <<"file://", Path/binary>>.
+  Path1 = case is_windows() of
+           true -> els_utils:to_binary(string:replace(Path, "\\", "/"));
+           false -> Path
+         end,
+  <<"file:///", Path1/binary>>.
+
+-spec is_windows() -> boolean().
+is_windows() ->
+  {OS, _} = os:type(),
+  OS =:= win32.
+
+-spec is_local(binary()) -> boolean().
+is_local(Host) ->
+  Host =:= <<>> orelse Host =:= <<"localhost">>.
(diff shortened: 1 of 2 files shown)


src/els_uri.erl
```diff
@@ -30,9 +30,47 @@ module(Uri) ->
...
 uri(Path) ->
-  <<"file://", Path/binary>>.
+  Path1 =
+  case is_windows() of
+    true ->
+      list_to_binary(string:replace(Path, <<"\\">>, <<"/">>));
+    _ ->
+      Path
+  end,
+  <<"file:///", Path1/binary>>.
```
jfacorro: Optional: The expression in this function could be a little more packed.
```suggestion
  Path1 = case is_windows() of
           true -> els_utils:to_binary(string:replace(Path, "\\", "/"));
           false -> Path
         end,
  <<"file:///", Path1/binary>>.
```
Notice the usage of `els_utils:to_binary/1` instead of `list_to_binary/1`.

src/els_uri.erl
```diff
@@ -30,9 +30,47 @@ module(Uri) ->
...
+
+-spec is_windows() -> boolean().
+is_windows() ->
+  case os:type() of
+    {win32, _} ->
+      true;
+    _ ->
+      false
+  end.
```
jfacorro: Optional: A more concise alternative.
```suggestion
-spec is_windows() -> boolean().
is_windows() ->
  {OS, _} = os:type(),
  OS =:= win32.
```

src/els_uri.erl
```diff
@@ -30,9 +30,47 @@ module(Uri) ->
...
+
+-spec is_local(binary()) -> boolean().
+is_local(<<>>) ->
+  true;
+is_local(<<"localhost">>) ->
+  true;
+is_local(_) ->
+  false.
```
jfacorro: Optional: A more concise alternative.
```suggestion
-spec is_local(binary()) -> boolean().
is_local(Host) ->
  Host =:= <<>> orelse Host =:= <<"localhost">>.
```

src/els_hover_provider.erl
```diff
@@ -133,6 +133,7 @@ edoc(M, F, A) ->
   try
     {ok, Uri} = els_utils:find_module(M),
     Path      = els_uri:path(Uri),
+    lager:debug("Getting edoc for ~p", [Path]),
```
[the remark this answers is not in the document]
filmor: Sure, if you don't deem it necessary I'll remove it. I found it useful.

src/els_uri.erl
filmor: I'll apply this one as well and squash the commits.

robertoaloi:
Thanks for this! @filmor ! One thing we are lacking in this project (which is the root cause of this kind of bugs) is a CI pipeline for Windows. Do you think it's something you could help with?

jfacorro:
I think these might have introduced some issues when trying to find files. Clients are sending URI for local files as `file:///Users/juan.facorro/dev/erlang-ls/erlang_ls/src/els_compiler_diagnostics.erl` and the changes here are not compatible with that.

`uri_string:parse/1` has the following behaviour:
```erlang
1> uri_string:parse(<<"file://host/foo">>).
#{host => <<"host">>,path => <<"/foo">>,scheme => <<"file">>}
2> uri_string:parse(<<"file:///foo">>). 
#{host => <<>>,path => <<"/foo">>,scheme => <<"file">>}
```
Which means the extra `/` added in this PR is not really correct.

PR #594, opened 2020-03-19
[#285] Introduce background jobs

### LINKED ISSUE #285: Show indexing progress to clients
On project initialization, the indexing process can consume a lot of CPU and memory, which many users have been reporting. Displaying a clear indication that the indexing is happening in the background would improve the user experience.

robertoaloi:
### Description

This PR introduce a framework to run supervised _background jobs_ within the language server. Background jobs have the ability of report progress, using the new `$/progress` functionality introduced with the version `3.15` of the LSP protocol, if the client supports it. Progress reports are logged as `DEBUG` information.

The indexing process has been converted to use background jobs as part of the same PR. This is how it now looks in VS Code:

Screenshot 2020-03-19 at 22 32 12

The Emacs client advertises itself in the `initialize` as supporting the feature. In reality, the implementation is just a stub, so no UI is produced (yet).

Finally, indexing has now been disabled while running the test. While implementing this functionality, I realized that we were triggering the indexing as part of each `init_per_testcase`. Given that we were not killing the language server between test cases and the indexing was unsupervised, multiple indexing processes ended up accumulating data in the same DB, eventually causing the `initialize` call to timeout (due to the big `wait-for-tables`). This was source of flakyness in the tests.

Fixes #285 .


src/els_background_job.erl
```diff
@@ -0,0 +1,190 @@
...
+%%==============================================================================
+%% Internal functions
+%%==============================================================================
+-spec step(pos_integer()) -> pos_integer().
+step(0) -> 0;
+step(N) -> 100 / N.
+
+-spec progress_msg(non_neg_integer(), pos_integer()) -> binary().
+progress_msg(Current, Total) ->
+  list_to_binary(io_lib:format("~p / ~p", [Current, Total])).
```
jfacorro: I think we need to add the "No Call" rule like @alanz suggested to avoid the usage of this functions.

src/els_indexing.erl
```diff
@@ -215,3 +187,15 @@ index_dir(Dir, Mode) ->
              "[succeeded=~p] "
              "[failed=~p]", [Dir, Mode, Time/1000/1000, Succeeded, Failed]),
   {Succeeded, Failed}.
+
+-spec entries_apps() -> [{string(), 'deep' | 'shallow'}].
```
jfacorro: Not requiring a change but why not make a single function out of these three that takes the name of the configuration entry as an argument?

src/els_work_done_progress.erl
```diff
@@ -0,0 +1,93 @@
...
+send_create_request() ->
+  Token = els_progress:token(),
+  Method = <<"window/workDoneProgress/create">>,
+  Params = #{token => Token},
+  ok = els_server:send_request(Method, Params),
+  Token.
+
+-spec value_begin(binary(), binary(), percentage()) -> value_begin().
+value_begin(Title, Message, Percentage) ->
+  #{ kind        => 'begin'
```
jfacorro: Shouldn't these `kind` values be added to `erlang_ls.hrl` instead?

test/els_progress_SUITE.erl
```diff
@@ -0,0 +1,136 @@
...
+  ?assertEqual(1, meck:num_calls(sample_job, task, '_')),
+  ?assertEqual(0, meck:num_calls(sample_job, on_complete, '_')),
+  ?assertEqual(1, meck:num_calls(sample_job, on_error, '_')),
+  ok.
+
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+-spec wait_for_completion(pid()) -> ok.
+wait_for_completion(Pid) ->
```
jfacorro: Could be useful to move to `els_test_utils`.

test/els_progress_SUITE.erl
```diff
@@ -0,0 +1,136 @@
...
+
+%%==============================================================================
+%% Testcases
+%%==============================================================================
+
+-spec sample_job(config()) -> ok.
+sample_job(_Config) ->
+  {ok, Pid} = new_background_job(),
+  wait_for_completion(Pid),
+  ?assertEqual(length(entries()), meck:num_calls(sample_job, task, '_')),
```
jfacorro: I think this module is not included in this PR, is it? How can the tests not fail? 😮

src/els_background_job.erl
robertoaloi: Not sure if that's too restrictive, but feel free to add it and then we can decide if keeping or not. Fixing this occurrence in the meantime.

robertoaloi: I think we should also add this to the guidelines.

src/els_work_done_progress.erl
robertoaloi: Some of the _types_ have very similar names (e.g. `kind`) and to avoid collisions, we end up with very long prefixed type names (e.g. `work_done_progress_begin`). My suggestion is that we slowly move away from a centralized header file with all type definitions and specify _local_ types, exporting them when necessary. For example, look at how the _progress token_ is exported in this PR. WDYT?

test/els_progress_SUITE.erl
robertoaloi: Probably. Let's do it once we have a second use case, though.

test/els_progress_SUITE.erl
robertoaloi: The `sample_job` is mocked using the `non_strict` option. This allows to define mocks for non-existing modules and functions. Something that I discovered recently with `meck` and makes our life so nice.

src/els_background_job.erl
robertoaloi: In retrospection, since this seems to be the only change I would make, I would prefer to merge it as-is and fix once we introduce the elvis restriction (so we can try it).

src/els_work_done_progress.erl
jfacorro: I agree that we don't want long macro names, but now we will have a mixture of both.

test/els_progress_SUITE.erl
jfacorro: I doubt I will remember this function exists if/when I need it 😞.

test/els_progress_SUITE.erl
jfacorro: Magic 🌟

src/els_work_done_progress.erl
robertoaloi: I know. I just didn't want to make a big switch. Something else we should decide on and add to the guidelines.

PR #599, opened 2020-03-20
[#597] Refactor code lenses, making them customizable

### LINKED ISSUE #597: Customizable code lenses
PR #585 introduces support for code lenses. Let's make the list of lenses configurable via the `erlang_ls.config` file.

robertoaloi:
### Description

* Rename the "info" code lens into "server-info" for clarity
* Do not activate the "server-info" code lens by default
* Allow code lenses to be configurable via the `erlang_ls.config` file
* Add behaviour definition for code lenses
* Convert the `server-info` code lens to use the new behaviour
* Refactor `command` functions into a separate module
* Make the prefixing for commands transparent

Fixes #597 .


alanz:
Before reviewing the contents of the PR, I have a philosophical problem with one part of it.

> - Do not activate the "server-info" code lens by default
> - Allow code lenses to be configurable via the erlang_ls.config file

Right now I am using this feature to try to track down why a particular sub-project is not loading a config file.  I would not be able to do this if I first need a properly configured config file to enable the feature in the first place. At which point I no longer need it.

robertoaloi:
@alanz I understand why you want this lens to debug things on your scenario, but I don't think that reason suffice to push this on all Erlang LS users. Being present on each module is quite invasive, if you don't care about it. With the refactored version, you can simply change one line in the `els_code_lens_server_info` module to enable it by default:

```
-spec is_default() -> boolean().
is_default() ->
  false.
```

Just set it to `true` and you will get it by default. In my mind it's an extremely easy patch to maintain for local development/troubleshooting. You can even stick a fork the module with the `true` into a `patches` directory that you can prepend to your path (e.g. using the `.erlang` file or `rebar3` functionalities. That way, you will have it on by default.

Also, if you would like to troubleshoot your specific sub-project, we're here to help.

PR #600, opened 2020-03-20
[#593] Fix URI handling bug

jfacorro:
### Description

Fix URI handling bug.

Fixes #593.

### DIFF
--- a/src/els_uri.erl
+++ b/src/els_uri.erl
@@ -39,7 +39,7 @@ path(Uri) ->
 
   case {is_windows(), is_local(Host)} of
     {_, true} ->
-      string:slice(Path, 1);
+      Path;
     {true, false} ->
       <<"//", Host/binary, Path/binary>>;
     {false, false} ->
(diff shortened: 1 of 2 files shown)


robertoaloi:
Merging this since the current version breaks on Mac OS. @filmor could you check it works on Windows, too? We can refine it after-merge, if it does not.

filmor:
It will not work on Windows. The slicing is required because the path component will look like `/c:/some_thing` with is not a valid path. I'll look into it again.

src/els_uri.erl
```diff
@@ -52,7 +52,7 @@ uri(Path) ->
            true -> els_utils:to_binary(string:replace(Path, "\\", "/"));
            false -> Path
          end,
-  <<"file:///", Path1/binary>>.
+  <<"file://", Path1/binary>>.
```
filmor: This is definitely wrong if you don't guarantee an absolute path.

jfacorro: I may be wrong, but as far as I could see, the clients always send absolute paths to the server. Isn't this the case?

filmor: It is, but this is the other way round (from Erlang to client). I've found the "official" logic and will try to implement it here: https://github.com/microsoft/vscode-uri/blob/master/src/index.ts#L1005-L1026

PR #602, opened 2020-03-21
Implement URI logic that should work on all systems

filmor:


filmor:
I'll test this tomorrow again on Windows and will try to work on a proper platform-independent GitHub action for Erlang.

filmor:
Running this manually on Windows works fine, but building is pretty broken right now. I've rolled back those two commits and will create a new PR for them, the change as such works.

jfacorro:
> Running this manually on Windows works fine, but building is pretty broken right now. I've rolled back those two commits and will create a new PR for them, the change as such works.

@filmor Should we close this PR then?

filmor:
No, we should merge this PR ;)

The problem is more that I can't really run the test-suite on Windows right now, but I'll be working on that in a separate PR.

jfacorro:
@robertoaloi Tried this branch. It works on my machine.

PR #605, opened 2020-03-24
[#604] Add run-ct-test code lens, refactor code lens behaviour

### LINKED ISSUE #604: Implement Code Lens to run CT test

robertoaloi:
### Description

This PR introduces a new _Code Lens_ and a _Command_, named "ct-run-test", through which is now possible to run Common Test testcases by clicking on a button from a test suite.

The code lens is activated automatically whenever the `ct.hrl` header is included in a module. The lens will be enabled for all functions of arity `1` in the test suite and a _blacklist_ avoids functions such as `init_per_suite/1` or `end_per_suite/` to be included.

The CT test is executed as a _background job_ and a CT hook is used to report diagnostics in case of failed or skipped tests.

I used this opportunity to refine the _code lens_ behaviour, introducing some extra callbacks that should hopefully simplify the creation of new code lenses. I will document the behaviour properly once the API is settled.

More things need to be implemented or fixed, but I believe this PR can give us a starting point.

Finally, I moved some more type definitions away from the `erlang_ls.hrl`. Having them as part of individual modules give us a nicer way to interact with types (e.g. via type completion) and avoids collisions between names. I plan to go in this direction for the whole project.

Fixes #604 .

### DIFF
--- a/src/els_code_lens_ct_run_test.erl
+++ b/src/els_code_lens_ct_run_test.erl
@@ -1,0 +1,61 @@
+%%==============================================================================
+%% Code Lens: ct_run_test
+%%==============================================================================
+
+-module(els_code_lens_ct_run_test).
+
+-behaviour(els_code_lens).
+-export([ command/1
+        , command_args/2
+        , is_default/0
+        , pois/1
+        , precondition/1
+        , title/1
+        ]).
+
+-include("erlang_ls.hrl").
+
+-spec command(poi()) -> els_command:command_id().
+command(_POI) ->
+  <<"ct-run-test">>.
+
+-spec command_args(els_dt_document:item(), poi()) -> [any()].
+command_args( #{uri := Uri} = _Document
+            , #{id := {F, A}, range := #{from := {Line, _}}} = _POI) ->
+  [#{ module => els_uri:module(Uri)
+    , function => F
+    , arity => A
+    , uri => Uri
+    , line => Line
+    }].
+
+-spec is_default() -> boolean().
+is_default() ->
+  true.
+
+-spec pois(els_dt_document:item()) -> [poi()].
+pois(Document) ->
+  Functions = els_dt_document:pois(Document, [function]),
+  [POI || #{id := {F, 1}} = POI <- Functions, not is_blacklisted(F)].
+
+-spec precondition(els_dt_document:item()) -> boolean().
+precondition(Document) ->
+  Includes = els_dt_document:pois(Document, [include_lib]),
+  case [POI || #{id := "common_test/include/ct.hrl"} = POI <- Includes] of
+    [] ->
+      false;
+    _ ->
+      true
+  end.
+
+-spec title(poi()) -> binary().
+title(_POI) ->
+  <<"Run test">>.
+
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+
+-spec is_blacklisted(atom()) -> boolean().
+is_blacklisted(Function) ->
+  lists:member(Function, [init_per_suite, end_per_suite]).
--- a/src/els_code_lens_ct_run_test.erl
+++ b/src/els_code_lens_ct_run_test.erl
@@ -58,4 +58,4 @@ title(_POI) ->
 
 -spec is_blacklisted(atom()) -> boolean().
 is_blacklisted(Function) ->
-  lists:member(Function, [init_per_suite, end_per_suite]).
+  lists:member(Function, [init_per_suite, end_per_suite, group]).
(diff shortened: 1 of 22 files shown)


alanz:
I notice it currently drops `.beam` files into the test directory next to the source files.

src/els_cth.erl
```diff
@@ -0,0 +1,37 @@
+-module(els_cth).
```
alanz: This is quite a cryptic name, should there perhaps be a document comment saying what this module is for?

robertoaloi:
> I notice it currently drops `.beam` files into the test directory next to the source files.

This is probably due to the `auto_compile` option. It's one of the things I need to follow-up. Should we create a ticket for that and address it separately?

robertoaloi:
> But I presume this is a work in progress.

It is indeed. There are many scenarios where things do not work correctly. My plan is to start having the feature (which per se should be quite stable, even if the tests could fail for various reasons) so that people can play around with it and report specific issues, that we can address individually. If we see that the lens creates too much confusion and we haven't address all the major concerns, we can disable it by default before next release.

@alanz Can I ask you to report the specific issues you noticed?

src/els_cth.erl
robertoaloi: 100% agree. Do you have a suggestion?

src/els_command_ct_run_test.erl
```diff
@@ -0,0 +1,59 @@
...
+
+-spec execute(map()) -> ok.
+execute(#{ <<"module">>   := M
+         , <<"function">> := F
+         , <<"arity">>    := A
+         , <<"line">>     := Line
+         , <<"uri">>      := Uri
+         }) ->
+  Msg = io_lib:format("Running Common Test [mfa=~s:~s/~p]", [M, F, A]),
+  lager:info(Msg, []),
```
jfacorro: Why this combination of `io_lib:format/2` + `lager:info/2` and not only use `lager:info/2`?
```suggestion
  lager:info("Running Common Test [mfa=~s:~s/~p]", [M, F, A]),
```

src/els_cth.erl
```diff
@@ -0,0 +1,37 @@
...
+  State.
+
+-spec on_tc_skip(suite(), testcase(), skip_reason(), state()) -> state().
+on_tc_skip(_Suite, _TestCase, {_ReasonType, Reason}, State) ->
+  publish_result(Reason, State),
+  State.
+
+-spec publish_result(reason(), state()) -> ok.
+publish_result(Reason, #{options := #{uri := Uri, line := Line}}) ->
+  Message = unicode:characters_to_binary(io_lib:format("~p", [Reason])),
```
jfacorro:
```suggestion
  Message = els_utils:to_binary(io_lib:format("~p", [Reason])),
```

src/els_code_lens_ct_run_test.erl
```diff
@@ -0,0 +1,61 @@
...
+
+-spec pois(els_dt_document:item()) -> [poi()].
+pois(Document) ->
+  Functions = els_dt_document:pois(Document, [function]),
+  [POI || #{id := {F, 1}} = POI <- Functions, not is_blacklisted(F)].
+
+-spec precondition(els_dt_document:item()) -> boolean().
+precondition(Document) ->
+  Includes = els_dt_document:pois(Document, [include_lib]),
+  case [POI || #{id := "common_test/include/ct.hrl"} = POI <- Includes] of
```
jfacorro: Does this mean that we are detecting if a module is a Common Test suite based on it including this header file?

src/els_code_lens_ct_run_test.erl
```diff
@@ -0,0 +1,61 @@
...
+title(_POI) ->
+  <<"Run test">>.
+
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+
+-spec is_blacklisted(atom()) -> boolean().
+is_blacklisted(Function) ->
+  lists:member(Function, [init_per_suite, end_per_suite]).
```
jfacorro: Why not include all the other callbacks from Common Test?

src/els_command_ct_run_test.erl
robertoaloi: Because the `Message` is re-used in the `Title`.

src/els_code_lens_ct_run_test.erl
robertoaloi: Exactly. We can refine this strategy in the future if we want to (e.g. if the name ends up with a specific suffix, etc).

src/els_code_lens_ct_run_test.erl
robertoaloi: Because the other ones do not have arity `1`. But I missed `group`, which I'm adding now.

robertoaloi:
> I notice it currently drops `.beam` files into the test directory next to the source files.

Created https://github.com/erlang-ls/erlang_ls/issues/610 as a follow-up

PR #607, opened 2020-03-25
[#606] Precalculate specs

### LINKED ISSUE #606: Do not store spec tree in DB
Currently, we store the spec "tree" in raw format in the DB (both in the _Document_ table and in the _Signature_ table). We can pre-calculate the pretty-printed version of the spec and store that in the DB. That should allow us to save some space. This is especially relevant for big code bases (since the DB is in memory). Keep any eye on the indexing times to ensure they don't increase significantly.

robertoaloi:
### Description

In this PR we pre-calculate the pretty-printed version of the specs during indexing. This avoids to store (sometimes huge) syntax trees in the DB and un-necessary recalculation of those specs. I realized we store the specs twice, both in the _document_ table and the _signatures_ table. For the time being I'm leaving things as they are (after all the _signatures_ table is just an index), but we should consider storing them once.

I tried this change both in the `erlang_ls` codebase and in a bigger codebase (1.5 millions lines of code). Here are some number.

#### Erlang LS

|               | Before | After  |
|---------------|--------|--------|
| Document DB   | 195 MB | 165 MB |
| Signatures DB | 31 MB  | 1MB    |

#### Bigger Code Base

|               | Before | After  |
|---------------|--------|--------|
| Document DB   | 577 MB | 505 MB |
| Signatures DB | 97 MB  | 4 MB   |

Not entirely sure why, for the larger code base, the Signatures DB shrinked more than the document. Number of table entries is the same before/after.
There were no significant difference in the indexing time.
 
Fixes #606 .

### DIFF
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -17,7 +17,7 @@
 
 -define(SERVER, ?MODULE).
 -define(TIMEOUT, infinity).
--define(DB_SCHEMA_VSN, <<"6">>).
+-define(DB_SCHEMA_VSN, <<"7">>).
 -define(DB_SCHEMA_VSN_FILE, "DB_SCHEMA_VSN").
 
 %%==============================================================================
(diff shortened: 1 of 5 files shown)


alanz:
I did not measure the individual DB, but `du` reports one codebase index going from 1.1GB to 840MB, and another from 540MB to 146MB.

PR #612, opened 2020-03-26
[#476] list possible transport options

### LINKED ISSUE #476: Use parser for server CLI options

liamjjmcnamara:
### Description

List possible transport options in README

Fixes #476.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,8 @@ Usage: Erlang LS [-v] [-t [<transport>]] [-p [<port>]] [-d [<log_dir>]]
 
   -v, --version    Print the current version of Erlang LS
   -t, --transport  Specifies the transport the server will use for the
-                   connection with the client. [default: tcp]
+                   connection with the client, either "tcp" or "stdio".
+                   [default: tcp]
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -29,7 +29,8 @@ Usage: Erlang LS [-v] [-t [<transport>]] [-p [<port>]] [-d [<log_dir>]]
 
   -v, --version    Print the current version of Erlang LS
   -t, --transport  Specifies the transport the server will use for the
-                   connection with the client. [default: tcp]
+                   connection with the client, either "tcp" or "stdio".
```
jfacorro: Please add this to the options returned by `--help` [here](https://github.com/erlang-ls/erlang_ls/blob/master/src/erlang_ls.erl#L68).

PR #617, opened 2020-03-27
[#616] Introduce Diagnostics and General Provider

### LINKED ISSUE #616: Introduce diagnostics provider

robertoaloi:
### Description

This PR is a replacement for #505. The PR introduces a _provider_ dedicated to diagnostics and a _general_ provider to handle what in the LSP protocol specifications are referred as _general_ messages.
This allows the server to have a more linear and modular structure, where all processes are part of the same supervision tree.

* Introduce _general_ provider
* Introduce _diagnostics_ provider
* Start _providers_ when application is started and not on _initialize_
* Run diagnostics as _background jobs_
* Move diagnostics logic away from the _text synchronization_ module
* Make diagnostics configurable via the `erlang_ls.config` file
* Allow indexing disabling as an _InitOption_

Sorry about the size of this PR, but diagnostics really needed some love.

Fixes #616 .


src/els_diagnostics.erl
```diff
@@ -54,10 +77,35 @@ make_diagnostic(Range, Message, Severity, Source) ->
...
+                end
             },
-  els_server:send_notification(Method, Params).
+  {ok, Pid} = els_background_job:new(Config),
+  Pid.
+
+%% @doc Return the callback module for a given Diagnostic Identifier
+-spec cb_module(els_diagnostics:diagnostic_id()) -> module().
+cb_module(Id) ->
+  binary_to_atom(<<"els_", Id/binary, "_diagnostics">>, utf8).
```
jfacorro:
```suggestion
  binary_to_existing_atom(<<"els_", Id/binary, "_diagnostics">>, utf8).
```

src/els_general_provider.erl
```diff
@@ -0,0 +1,125 @@
+-module(els_general_provider).
```
jfacorro: Shouldn't this be called `els_initialize_provider`? The word general doesn't say much about what it does and IMO gives the idea that this module contains what is in the `els_provider`.

src/els_diagnostics_provider.erl
```diff
@@ -0,0 +1,111 @@
...
+                              , uri => Uri
+                              }|Rest]}
+  end;
+handle_info(_Request, State) ->
+  State.
+
+-spec handle_request(any(), state()) -> {any(), state()}.
+handle_request({run_diagnostics, Params}, State) ->
+  #{in_progress := InProgress} = State,
+  #{ <<"textDocument">> := #{ <<"uri">> := Uri}} = Params,
```
jfacorro: Nit: space right after `#{`.
```suggestion
  #{<<"textDocument">> := #{<<"uri">> := Uri}} = Params,
```

src/els_diagnostics_provider.erl
```diff
@@ -0,0 +1,111 @@
...
+  lager:debug("Starting diagnostics jobs [uri=~p]", [Uri]),
+  Jobs = els_diagnostics:run_diagnostics(Uri),
+  Entry = #{uri => Uri, pending => Jobs, diagnostics => []},
+  {noresponse, State#{in_progress => [Entry|InProgress]}}.
```
jfacorro: If I'm understanding the code right, if there is an ongoing group of diagnostics running for `Uri1` and a new group of diagnostics is triggered for the same `Uri1` we could end up with a weird situation where, one group and the other gets published alternatively, right?

What I'm trying to ask is why not make the `in_progress` value a map keyed on `uri`? Checking first if there is an existing entry for `Uri` and if there is:
1.  Send an empty diagnostics list.
2. Kill all running diagnostics jobs for that `Uri`.
3. Start the new group of diagnostics jobs.

src/els_general_provider.erl
alanz: The spec lists lifecyle messages under the heading "general". `initialize`, `initialized`, `shutdown`, `exit`.

So this matches the spec, but I think the spec is using a bad naming.

src/els_general_provider.erl
```diff
@@ -0,0 +1,125 @@
...
+        }
+   }.
+
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+-spec node_name(uri(), binary()) -> atom().
+node_name(RootUri, OtpPath) ->
+  <<SHA:160/integer>> = crypto:hash(sha, <<RootUri/binary, OtpPath/binary>>),
+  list_to_atom(lists:flatten(io_lib:format("erlang_ls_~40.16.0b", [SHA]))).
```
alanz: unrelated observation.  It would be useful for debugging if this was made up with an additional component being the last part of the root directory, as reported in the info command.

src/els_general_provider.erl
robertoaloi: What Alan said. I also think is not a great name, but that's what those messages are referred to in the specs.

src/els_diagnostics_provider.erl
robertoaloi: True. I would prefer to tackle this as a follow-up, though. Created #619 for this purpose.
We also need to add support for "cancellation" for background jobs, which could be used for this.
Also, `3.15` of the protocol adds a `version` field which should help, too.

src/els_general_provider.erl
robertoaloi: Yes, I'm also not happy about the current naming convention. Feel free to open an issue about this and we tackle this.

PR #618, opened 2020-03-28
[#156] Xref diagnostics

### LINKED ISSUE #156: Publish xref-like errors as diagnostics for undefined functions
If a fully qualified function does not exist, an error should be presented to the user.

robertoaloi:
### Description

_Xref_ diagnostics will report an error for all those cases where a function _application_, an _implicit fun_, an _import_ or _export_ entry does not have a corresponding definition in the server, resembling what `xref` would do. The advantage, compared to `xref`, is that these diagnostics operate on the POI/DB level and don't require _binaries_ or `debug_info`.

Based on top of  #616 , which should be merged first (and this one rebased on top of the latest master).

Fixes #156.

### DIFF
--- a/priv/code_navigation/src/diagnostics_xref.erl
+++ b/priv/code_navigation/src/diagnostics_xref.erl
@@ -3,4 +3,8 @@
 -export([ main/0 ]).
 
 main() ->
-  lists:map(1, 2, 3).
+  lists:map(1, 2, 3),
+  non_existing() ++ existing().
+
+existing() ->
+  lists:seq(1, 3).
--- a/priv/code_navigation/src/diagnostics_xref.erl
+++ b/priv/code_navigation/src/diagnostics_xref.erl
@@ -1,0 +1,6 @@
+-module(diagnostics_xref).
+
+-export([ main/0 ]).
+
+main() ->
+  lists:map(1, 2, 3).
--- a/src/els_xref_diagnostics.erl
+++ b/src/els_xref_diagnostics.erl
@@ -47,8 +47,14 @@ source() ->
 %% Internal Functions
 %%==============================================================================
 -spec make_diagnostic(poi()) -> els_diagnostics:diagnostic().
-make_diagnostic(#{range := Range}) ->
-  Message = <<"Cannot find definition for this function">>,
+make_diagnostic(#{range := Range, id := Id}) ->
+  Function = case Id of
+               {F, A} -> lists:flatten(io_lib:format("~p/~p", [F, A]));
+               {M, F, A} -> lists:flatten(io_lib:format("~p:~p/~p", [M, F, A]))
+             end,
+  Message = els_utils:to_binary(
+              io_lib:format( "Cannot find definition for function ~s"
+                           , [Function])),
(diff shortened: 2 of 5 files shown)


robertoaloi:
@jfacorro @alanz This is now ready for review.

src/els_xref_diagnostics.erl
```diff
@@ -0,0 +1,65 @@
...
+  Message = <<"Cannot find definition for this function">>,
+  Severity = ?DIAGNOSTIC_ERROR,
+  els_diagnostics:make_diagnostic( els_protocol:range(Range)
+                                 , Message
+                                 , Severity
+                                 , source()).
+
+-spec has_definition(poi(), els_dt_document:item()) -> boolean().
+has_definition(POI, #{uri := Uri}) ->
+  case els_code_navigation:goto_definition(Uri, POI) of
```
jfacorro: Isn't this is a somewhat expensive operation? Have you tried running this in a big module where there are a lot of those POIs? This is a background job so I assume it won't impact responsiveness though.

priv/code_navigation/src/diagnostics_xref.erl
```diff
@@ -0,0 +1,6 @@
+-module(diagnostics_xref).
+
+-export([ main/0 ]).
+
+main() ->
+  lists:map(1, 2, 3).
```
jfacorro: Should there be a function call that is valid as well in this example module?

src/els_xref_diagnostics.erl
robertoaloi: That is a fair point. No, I haven't tried since, as you say, this is a background job. Worth checking.

priv/code_navigation/src/diagnostics_xref.erl
robertoaloi: Good point. Added calls to both local and non-local, existing and non-existing functions. One is also picked by the compiler.

robertoaloi:
One question I have is how do we want to handle functions which are the result of parse transforms. For example, after this PR the server will complain that functions such as `lager:info/2` do not exist.

jfacorro:
> One question I have is how do we want to handle functions which are the result of parse transforms. For example, after this PR the server will complain that functions such as `lager:info/2` do not exist.

I don't think we can handle correctly unless we work with the compiled AST in `debug_info`. A non-trivial alternative that comes to mind is using a code lens or code action (not sure which), that enables ignoring these warnings. This is a whole new feature in itself, since we should enable ignoring other warnings as well. 

Adding a configuration entry in `erlang_ls.config` to ignore them is not something I would like to have. If we add these kind of things, then it's very likely that the file will end up extremely bloated. I think the idea is that `erlang_ls.config` should have as little customizations as possible, only changing defaults when the project deviates from the standard.

robertoaloi:
@jfacorro I would avoid going the ignore way, too. The `xref` tool exists already and operates on the debug_info. I'd say we can merge this is a stub providing a minimum set of functionality and then move to a proper xref solution. This also means we should finalize the "runtime node" discussion.

robertoaloi:
Created  #621 as a follow-up.

PR #634, opened 2020-03-30
[#622] Use general provider consistently

### LINKED ISSUE #622: Move all "general" messages to dedicated provider
Now that a provider exists for "general" message, we should move the following messages there as well:

* initialized
* shutdown
* exit

robertoaloi:
### Description

The major advantage of this PR (apart from having a more consistent codebase in the `els_methods` module) is to serialize _general_ requests via the dedicated provider. After moving only the `initialize` request, I noticed some flakyness in the proper tests which was due to these messages being handled in the wrong order.

Fixes #622 .

### DIFF
--- a/src/els_methods.erl
+++ b/src/els_methods.erl
@@ -5,12 +5,11 @@
 -export([ dispatch/4
         ]).
 
--export([ initialized/2
+-export([ exit/2
+        , initialize/2
+        , initialized/2
         , shutdown/2
-        , exit/2
-        ]).
-
--export([ textdocument_completion/2
+        , textdocument_completion/2
         , textdocument_didopen/2
         , textdocument_didchange/2
         , textdocument_didsave/2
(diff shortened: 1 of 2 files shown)


src/els_methods.erl
```diff
@@ -117,12 +113,26 @@ method_to_function_name(Method) ->
...
 %% Initialized
 %%==============================================================================
 
 -spec initialized(params(), state()) -> result().
-initialized(_Params, State) ->
+initialized(Params, State) ->
+  Provider = els_general_provider,
+  Request  = {initialized, Params},
+  _Response = els_provider:handle_request(Provider, Request),
   %% Report to the user the server version
```
jfacorro: Why not move all of this to the `els_general_provider` module?

robertoaloi: Because I wanted to refactor this separately, for two main reasons:

1. In `3.15`, the initiate method allows replying with `ServerInfo`, which include exactly name and version. We should not send them back explicitly as a "showMessage" because this is not part of the protocol.
2. By moving the implementation to the provider, the `Methods` and `Params` should be returned in a single value, so it can be used transparently by this module.

I will create an issue with this.

Issue #614, opened 2020-03-26
No diagnostics reported if a dependency does not compile


alanz:
**Describe the bug**

When a file is compiled having a parse transform or behaviour, the module providing that is located and compiled so it is available to the compiler.  But if compilation fails for that module, no warning or error is reported, only a warning is written to the log

https://github.com/erlang-ls/erlang_ls/blob/8ff322fddcf162ed0af494704fd9fd31b3ed3268/src/els_compiler_diagnostics.erl#L247-L262

**To Reproduce**

Open a file with a behaviour where the behaviour module does not compile.

**Expected behavior**

A diagnostic is provided with the reason the behaviour could not be used.

**Actual behavior**

A warning in the file using the behaviour that the behaviour could not be found.

**Context**
 - `erlang_ls` version (tag/sha): 8ff322fddcf162ed0af494704fd9fd31b3ed3268
 - Editor used: emacs
 - LSP client used: lsp-mode

PR #635, opened 2020-03-30
[#614] Report diagnostic if behaviour dependency does not compile

alanz:
When a file is compiled having a parse transform or behaviour, the module
providing that is located and compiled so it is available to the compiler.

If that compilation fails, generate a diagnostic in the including file to report
it.

Addresses #614, still need to process parse transform.


alanz:
ping @robertoaloi

src/els_compiler_diagnostics.erl
```diff
@@ -170,19 +172,32 @@ range(none) ->
...
   POIs       = els_dt_document:pois(Document, [include]),
   IncludeId  = include_id(IncludePath),
   [Range || #{id := Id, range := Range} <- POIs, Id =:= IncludeId];
 inclusion_range(IncludePath, Document, include_lib) ->
   POIs       = els_dt_document:pois(Document, [include_lib]),
   IncludeId  = include_lib_id(IncludePath),
+  [Range || #{id := Id, range := Range} <- POIs, Id =:= IncludeId];
+inclusion_range(IncludePath, Document, behaviour) ->
+  POIs       = els_dt_document:pois(Document, [behaviour]),
+  %% IncludeId  = include_id(IncludePath),
```
robertoaloi, line 194: Any reason for the commented code? Also, these should be `BehaviourId` and `ParseTransformId`, since they are not related to _includes_ (probably a leftover from copy/paste).

src/els_compiler_diagnostics.erl
```diff
@@ -231,42 +256,59 @@ string_to_term(Value) ->
...
+        Path = els_utils:to_list(els_uri:path(Uri)),
+        Opts = lists:append([macro_options(), include_options(), [binary]]),
+        case compile:file(Path, Opts) of
+          {ok, Module, Binary} ->
+            code:load_binary(Module, atom_to_list(Module), Binary),
+            [];
+          Error ->
+            lager:warning("Error compiling dependency [error=~w]", [Error]),
+            %% Compile again to generate diagnostics
+            get_dependency_diagnostics(Uri, IncludingPath)
```
robertoaloi, line 290: Why compiling it again? We just did that. Shouldn't we add the options to the `compile:file` above so that we actually return warnings and errors?

robertoaloi:
@alanz Left some minor comment. I am happy to include this one, but bear in mind things will probably change once we move compilation to the runtime node.

src/els_compiler_diagnostics.erl
alanz: It seems I misread the docs, thought you did not get the warnings when specifying `binary` as an option.

src/els_diagnostics_utils.erl
```diff
@@ -24,10 +24,16 @@ dependencies(Uri) ->
...
-  IncludedUris = included_uris(Document),
-  dependencies(Uris ++ IncludedUris, Acc ++ [Id || #{id := Id} <- Deps]).
+  case els_dt_document:lookup(Uri) of
+    {ok, [Document]} ->
+      els_dt_document:lookup(Uri),
+      Deps = els_dt_document:pois(Document, [behaviour, parse_transform]),
+      IncludedUris = included_uris(Document),
+      dependencies(Uris ++ IncludedUris, Acc ++ [Id || #{id := Id} <- Deps]);
+    Error ->
+      lager:info("els_diagnostics_utils:dependencies [Error=~p]", [Error]), %%AZ
```
robertoaloi, line 34: Nit: lager already takes care of including pid, module, function and line number in the log messages.

robertoaloi, line 34:
```suggestion
      lager:info("Lookup failed [Error=~p]", [Error]),
```

robertoaloi: Also, let's try not use initials in comments or todo. They are hard to understand for other contributors.

src/els_compiler_diagnostics.erl
```diff
@@ -279,37 +294,26 @@ load_dependency(Module, IncludingPath) ->
     case els_utils:find_module(Module) of
       {ok, Uri} ->
         Path = els_utils:to_list(els_uri:path(Uri)),
-        Opts = lists:append([macro_options(), include_options(), [binary]]),
-        case compile:file(Path, Opts) of
+        case compile:file(Path, diagnostics_options_load_code()) of
+          {ok, [], []} ->
+            lager:warning("AZ:WTF finding dependency, {ok, [], []}"),
```
robertoaloi, line 299: Can we please update the comment?

src/els_compiler_diagnostics.erl
```diff
@@ -230,11 +234,22 @@ include_options() ->
 
 -spec diagnostics_options() -> [any()].
 diagnostics_options() ->
+  lists:append([ diagnostics_options_bare()
+               , [basic_validation]
+               ]).
```
robertoaloi, line 239:
```suggestion
  [basic_validation|diagnostics_options_bare()].
```

src/els_compiler_diagnostics.erl
```diff
@@ -230,11 +234,22 @@ include_options() ->
...
+
+-spec diagnostics_options_load_code() -> [any()].
+diagnostics_options_load_code() ->
+  lists:append([ diagnostics_options_bare()
+               , [binary]
+               ]).
```
robertoaloi, line 245:
```suggestion
  [binary|diagnostics_options_bare()].
```

src/els_compiler_diagnostics.erl
```diff
@@ -230,11 +234,22 @@ include_options() ->
...
+
+-spec diagnostics_options_bare() -> [any()].
+diagnostics_options_bare() ->
   lists:append([macro_options()
```
robertoaloi, line 249:
```suggestion
  lists:append([ macro_options()
```

src/els_compiler_diagnostics.erl
alanz: oops, leftover debug

src/els_diagnostics_utils.erl
alanz: My approach is to put initials in for my debug comments, I neglected to remove it before pushing. Agree they should not be there in the repo

Issue #633, opened 2020-03-30
Do not consider all erlang functions as bifs


robertoaloi:
From a comment by @michalmuskala in https://github.com/erlang-ls/erlang_ls/pull/598:

> This is not 100% accurate. Not all functions in the erlang module are automatically imported. We probably should use erl_internal:bif/2 here.

Issue #621, opened 2020-03-29
Use xref for xref diagnostics


robertoaloi:
https://github.com/erlang-ls/erlang_ls/pull/618 introduces a stub for xref-like diagnostics, which uses the DB to figure out calls to undefined functions. This approach can only work partially, since it will provide false positives in case of parse transforms. The `xref` tool provided by OTP should be used instead.

alanz:
In terms of #644, given it is such a common usage, and the error reports are really annoying, perhaps we should do something like #645 for them.

alanz:
Given the lack of progress on build servers, I propose to focus on the existing xref implemention, and tweak it slightly.

Perhaps we should call it something else?

PR #645, opened 2020-04-01
[#641] do not report xref errors for pseudo functions

### LINKED ISSUE #641: xref cannot find record_info
**Describe the bug**

If a module uses `record_info`, the XRef diagnostics reports an error that the function `record_info/2` cannot be found.

According to the [docs](http://erlang.org/doc/reference_manual/records.html)

> To each module using records, a pseudo function is added during compilation to obtain information about records:
>
>`    record_info(fields, Record) -> [Field]`
>`     record_info(size, Record) -> Size`
>
> Size is the size of the tuple representation, that is, one more than the number of fields.

**To Reproduce**

Open a module using `record_info`

**Context**
 - `erlang_ls` version (tag/sha): 7ad7adb1a82b255f9cd947af2083b871ab5cac30
 - Editor used:
 - LSP client used:

### LINKED ISSUE #644: xref cannot find lager:xxx
**Describe the bug**

The xref diagnostics report on functions that do not have a definition.

`lager` uses a parse_transform to convert calls such as `lager:info("foo")` into
 something appropriate.  But the xref processing does not know about this, and
 so reports an error because it cannot resolve the reference.

**To Reproduce**

Open a file with a lager logging call in it.

**Expected behavior**

No xref warning for the lager call

**Actual behavior**

The lager call generates an xref error saying it s not defined.

**Context**
 - `erlang_ls` version (tag/sha): 6d0d808ba9804c2e85d5793d620717175821f9f4
 - Editor used:
 - LSP client used:

alanz:
The functions

    record_info/2
    module_info/0
    module_info/1

are pseudo functions generated and processed by the erlang compiler. Do not
report xref errors if they are encountered, as they are always valid.

Closes #641.
Closes #644.

### DIFF
--- a/src/els_xref_diagnostics.erl
+++ b/src/els_xref_diagnostics.erl
@@ -67,13 +67,16 @@ make_diagnostic(#{range := Range, id := Id}) ->
 
 -spec has_definition(poi(), els_dt_document:item()) -> boolean().
 has_definition(#{ kind := application
-                , id := {erlang, module_info, 0} }, _) -> true;
+                , id := {module_info, 0} }, _) -> true;
 has_definition(#{ kind := application
-                , id := {erlang, module_info, 1} }, _) -> true;
+                , id := {module_info, 1} }, _) -> true;
 has_definition(#{ kind := application
                 , id := {record_info, 2} }, _) -> true;
 has_definition(#{ kind := application
                 , id := {behaviour_info, 1} }, _) -> true;
+has_definition(#{ kind := application
+                , id := {lager, Level, Arity} }, _) ->
+  lager_definition(Level, Arity);
 has_definition(POI, #{uri := Uri}) ->
   case els_code_navigation:goto_definition(Uri, POI) of
     {ok, _Uri, _POI} ->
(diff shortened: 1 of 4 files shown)


src/els_xref_diagnostics.erl
```diff
@@ -62,6 +62,12 @@ make_diagnostic(#{range := Range, id := Id}) ->
                                  , source()).
 
 -spec has_definition(poi(), els_dt_document:item()) -> boolean().
+has_definition(#{ kind := application
+                , id := {erlang, module_info, 0} }, _) -> true;
+has_definition(#{ kind := application
+                , id := {erlang, module_info, 1} }, _) -> true;
```
michalmuskala, line 68: This seems wrong. Those functions are not defined on the erlang module. Why would they show up like this in here?

alanz: Empirical testing, I logged what I got when I opened files with the issue.

And it *does* work with this match.

michalmuskala: I wonder if it's related to this https://github.com/erlang-ls/erlang_ls/issues/633 - since those functions also exist on the `erlang` module

robertoaloi: The `module_info/0,1` functions are special and inserted by the compiler to **all** modules, not just the `erlang` one:

https://erlang.org/doc/reference_manual/modules.html#module_info-0-and-module_info-1-functions

I suggest we hold on for this one and instead focus on #621 , which will make this not needed.

robertoaloi:
@alanz should we close this one for now and focus on the build server / real xref? We can eventually re-open it later, keep the branch.

alanz:
Re-opening, to be able to use this feature on code with `record_info` in it.

michalmuskala:
Given https://github.com/erlang-ls/erlang_ls/issues/633 is solved, we probably need `{module_info, _}` rather than `{erlang, module_info, _}` included

alanz:
> Given #633 is solved, we probably need {module_info, _} rather than {erlang, module_info, _} included

I just confirmed that by running it manually. Will update the tests to catch it too.

robertoaloi (approved):

PR #647, opened 2020-04-01
Add all subdirs to search path

dgud:
All erlang files in a subdir of RootUri should be found,
not only files below 'src', 'test' or 'include'.

### DIFF
--- a/src/els_config.erl
+++ b/src/els_config.erl
@@ -118,10 +118,10 @@ do_initialize(RootUri, Capabilities, {ConfigPath, Config}) ->
   ok = set(code_lenses    , CodeLenses),
   ok = set(diagnostics    , Diagnostics),
   %% All (including subdirs) paths used to search files with file:path_open/3
+  ProjectDirs = els_utils:resolve_paths( [[ RootPath ]], RootPath, true),
+
   ok = set( search_paths
-          , lists:append([ project_paths(RootPath, AppsDirs, true)
-                         , project_paths(RootPath, DepsDirs, true)
-                         , include_paths(RootPath, IncludeDirs, false)
+          , lists:append([ ProjectDirs
                          , otp_paths(OtpPath, true)
                          ])
           ),
--- a/src/els_utils.erl
+++ b/src/els_utils.erl
@@ -187,16 +187,24 @@ is_symlink(Path) ->
 
 -spec resolve_path([path()], path(), boolean()) -> [path()].
 resolve_path(PathSpec, RootPath, Recursive) ->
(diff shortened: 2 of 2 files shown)


robertoaloi:
Hi @dgud and sorry for keeping your PRs in a limbo-state for such a long time. We had a few internal discussions with @alanz and decided that we will give a try to the Build Server Protocol (or maybe a _soft_ version of it, using Erlang RPC as the transport). Effectively, what is today the "runtime node" will become an actual "build server". See the first draft of the idea here:

https://github.com/erlang-ls/erlang_ls/blob/master/specs/runtime_node.md

This means that, for example, the language server will not be responsible for mangling paths, but that will be delegated to the build server, allowing us to simplify the language server significantly.

We are implementing a `rebar3` build server as a plugin (so that rebar3 projects can work out of the box) and we are discussing possible solutions for custom build systems (like the ones used in our custom proprietary code basis). We will probably provide a Build Server Library, so that you can simply implement a behaviour for custom build tools. It would be valuable for us to understand what would be the requirements from your side and discuss how this could work for the OTP code base. Maybe we could organize a quick call?

dgud:
No problem.

Sure, email me  dgud 'via' erlang.org  I'm available most of the times, locked in at home :-)

src/els_config.erl
```diff
@@ -118,10 +118,10 @@ do_initialize(RootUri, Capabilities, {ConfigPath, Config}) ->
   ok = set(code_lenses    , CodeLenses),
   ok = set(diagnostics    , Diagnostics),
   %% All (including subdirs) paths used to search files with file:path_open/3
+  ProjectDirs = els_utils:resolve_paths( [[ RootPath ]], RootPath, true),
+
   ok = set( search_paths
-          , lists:append([ project_paths(RootPath, AppsDirs, true)
-                         , project_paths(RootPath, DepsDirs, true)
```
robertoaloi, line 123: @dgud @alanz Are dependencies guaranteed to be within the root path? If they are not, we may have a problem here, since we are not searching any longer among dependencies.

dgud:
That might be a problem, didn't think of that.
But I don't have time to fix that now.

src/els_config.erl
robertoaloi: Confirmed as a major issue which breaks the latest master. This should be fixed in #732.

robertoaloi:
@dgud I have to revert this one. My fix doesn't work and the current master is broken. I am opening a separate issue to address your initial concern (do not limit the search to the /src, /test and /include paths).

PR #649, opened 2020-04-01
Don't index test directories

dgud:
They should be included in the search path anyway,
so will be indexed eventually.


robertoaloi:
Won't this change prevent references from test apps to be correctly identified?

robertoaloi:
As discussed, this will require some modifications. We cannot just exclude the whole `test` directory, but we could introduce a configuration value to blacklist/ignore directories matching a given pattern. Considering the attempt at utilizing a build server, this may not be worth at this stage, but that's up to you.

PR #598, opened 2020-03-20
[#527] Add 'imported' flag to auto-imported built-in functions

jfacorro:
### Description

Add 'imported' flag to auto-imported built-in functions.

Fixes #527.

PR #654, opened 2020-04-05
[#633] Follow up on #598: Do not consider all exports from erlang module as bifs

### LINKED ISSUE #633: Do not consider all erlang functions as bifs

onno-vos-dev:
Fixes #633 

**TIL**, when taking all functions that are listed as bifs in `erl_internal:bif/2` and checking if they're part of the `erlang`-module, I found that `is_bitstr/1` and `monitor/3` are not. I cannot find a trace of `is_bitstr/1` other than it was introduced in R11 by the looks of it nor `monitor/3`. My guess is that they were removed a long time ago but simple not removed from the `erlang_internal:bif/2` list.

### DIFF
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -184,7 +184,7 @@ application(Tree) ->
     undefined -> [];
     {F, A} ->
       Pos = erl_syntax:get_pos(Tree),
-      case is_erlang_bif(F, A) of
+      case erl_internal:bif(F, A) of
         %% Call to a function from the `erlang` module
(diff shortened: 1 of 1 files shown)


robertoaloi:
Thanks!

michalmuskala:
Reported the old bifs upstream https://github.com/erlang/otp/pull/2591

Issue #646, opened 2020-04-01
transient proper test failures in CI


alanz:
**Describe the bug**
It looks like we are getting transient proper failures for the xref tests, trying to look up for a document before it is indexed. See https://github.com/erlang-ls/erlang_ls/pull/645/checks?check_run_id=551528841

**Context**
 - `erlang_ls` version (tag/sha): https://github.com/alanz/erlang_ls/commit/6ca0bd9326bde1f864a229a81959b58149979927
 - Editor used:
 - LSP client used:

robertoaloi:
Stacktrace (in case logs from the GitHub actions get cleaned up):

```
[warning] Background job aborted. [reason={{badmatch,{ok,[]}},[{els_xref_diagnostics,run,1,[{file,"/__w/erlang_ls/erlang_ls/src/els_xref_diagnostics.erl"},{line,34}]},{els_background_job,handle_info,2,[{file,"/__w/erlang_ls/erlang_ls/src/els_background_job.erl"},{line,137}]},{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,637}]},{gen_server,handle_msg,6,[{file,"gen_server.erl"},{line,711}]},{proc_lib,init_p_do_apply,3,[{file,"proc_lib.erl"},{line,249}]}]}] [pid=<0.1918.0>]
```

robertoaloi:
It turns out it is not an indexing issue (indexing is disabled in the proper tests). The URIs used by the proper tests are not corresponding to _real_ files, so it is expected that those URIs do not have a corresponding entry in the DB. Most of the times this is not an issue because the diagnostics don't have the time to run during the proper tests, but sometimes that crash happens. Notice that this behaviour (which is easy to fix) is not the root cause of the failing build.

The real issue here is that both the client and the server use integers as the `request id`. Sometimes they end up using the same `id`, so a server-initiated request is occasionally (i.e. depending on timing) considered a response to a client-initiated request with the same id). I am going to fix that.

alanz:
Good. There should be a separate incrementing request id for server-initiated vs client-initiated requests.

alanz:
And in an earlier diff I changed the indexing to always use absolute paths for the files when indexing them, perhaps we should do that in the tests too? We specifiy them in the test as relative, but make them absolute at run time and run the test against the absolute ones.

We do that in the haskell-ide-engine tests too.

robertoaloi:
[...]

PR #659, opened 2020-04-07
Fix leading slash stripping for Windows

filmor:
### Description

I must have messed up something in the Regex, it didn't match for some reason in `re:run`, it does in `re:replace`...


filmor:
The build errors are completely unrelated...

robertoaloi:
@filmor Yes, the build error is due to some files not being correctly indexed before the tests. That will be addressed in #646

robertoaloi:
Giving a blind approval. We really need a Windows pipeline.

PR #660, opened 2020-04-07
Fix build

filmor:
Fixes a few issues with the rebar.config and build process.


robertoaloi:
Hi @filmor , this PR seems to touch multiple issues at once.  I think it would be better to tackle them separately.

filmor:
Give me a hint on what kind of split you want. The first commit is already in a different PR, I could remove it from this one (it's in there accidentally as I kind of needed it for trying to run tests). The rest is all related to `elvis_core` changes not correctly propagated and the current linting step creating severe issues when trying to run on Windows.

robertoaloi:
@filmor I would leave the non-elvis parts (redbug, lock file) out. I would also prefer an issue explaining what the PR is trying to address, so we can track the discussion.

@jfacorro did you know about the `project-fifo` plugin? Any opinion on using that one? Weird that it's not just called "elvis".

filmor:
The lock file in its current state is broken (refers to different package hashes than are actually installed). If you really want me to, I move the `redbug` stuff to a separate PR, but the only thing that changes there is that you forgot to put it in the `app.src`, that's why it wasn't included in the `escript`.

robertoaloi:
Facepalm. Cool then, you can leave everything here, then.

filmor:
There will probably be a few follow ups to make all build steps fully work on Windows, but this is a step forward.

filmor:
@robertoaloi, care to review?

robertoaloi:
Looks good, sorry it took so long to review. Thanks!

PR #661, opened 2020-04-07
[#657] Be able to configure lenses via config file

### LINKED ISSUE #657: code lens configuration via erlang_ls.config does not work
**Describe the bug**
According to #599 we should be able to enable non-default code lenses in the config file.

This does not work, as the config is expressed as binary strings, and these cannot be specified via the config file.

This needs a similar treatment as that for specifying enabled diagnostics providers, as landed in #652.

**To Reproduce**
Add

```yaml
code_lenses:
  - server_info
```

to the config file

**Expected behavior**

When the server starts up, a lens is displayed on the top line of the file.

**Actual behavior**
Nothing happens

**Context**
 - `erlang_ls` version (tag/sha): 4b2a79be3286b6553c2f36a3b84d9f9b5db1dc49
 - Editor used:
 - LSP client used:

robertoaloi:
### Description

Fixes #657 .

Please notice how the configuration option has been renamed to `lenses` for simplicity.

### DIFF
--- a/src/els_code_lens.erl
+++ b/src/els_code_lens.erl
@@ -61,7 +61,11 @@ default_lenses() ->
 
 -spec enabled_lenses() -> [lens_id()].
 enabled_lenses() ->
-  els_config:get(code_lenses).
+  Config = els_config:get(lenses),
+  Default = default_lenses(),
+  Enabled = maps:get("enabled", Config, []),
+  Disabled = maps:get("disabled", Config, []),
+  lists:usort((Default ++ valid(Enabled)) -- valid(Disabled)).
 
 -spec lenses(lens_id(), els_dt_document:item()) -> [lens()].
 lenses(Id, Document) ->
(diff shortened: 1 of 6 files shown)


src/els_code_lens.erl
```diff
@@ -61,7 +61,11 @@ default_lenses() ->
 
 -spec enabled_lenses() -> [lens_id()].
 enabled_lenses() ->
-  els_config:get(code_lenses).
+  Config = els_config:get(lenses),
+  Default = default_lenses(),
+  Enabled = maps:get("enabled", Config, []),
+  Disabled = maps:get("disabled", Config, []),
+  lists:usort((Default ++ valid(Enabled)) -- valid(Disabled)).
```
alanz, line 68: I wonder if we shouldn't extract this as a standard function? We are using it for the diagnostics logic too.

robertoaloi: I thought about it, but maybe that would make just things harder to read. As long as Elvis does not complain about duplicated code I guess we are fine.

robertoaloi:
I plan to remove the "code_" from the remaining bits, but that can wait.

PR #662, opened 2020-04-07
Change default transport to stdio

alanz:
This is the most common case, and in general if specifying tcp transport a port
needs to be specified too, as clients will launch multiple instances under some
circumstances.


alanz:
https://github.com/emacs-lsp/lsp-mode/pull/1567

robertoaloi:
@alanz The README also needs to be updated.

PR #668, opened 2020-04-08
[#386] go to definition from module import

### LINKED ISSUE #386: go to definition does not work for explicitly imported functions
**Describe the bug**

If I have a module with

```erlang
-import(some_module, [
    a_function/2
]).
```

then go to definition works for `some_module:a_function(A,B)`, but not for `a_function(A,B)`.

**Context**
 - `erlang_ls` version (tag/sha): 00b52965de2934fea96adc91930aea47f2e960a4
 - Editor used: VSCode
 - LSP client used: erlang_ls

alanz:
A function imported as

    -import(some_module, [fn/3]).

is treated as local, and does not need the `some_module` prefix when used.

Ensure that go to definition works at the usage sites as for fully qualified
calls.

Fixes #386.


alanz:
> That would allow us to keep the find function generic and simple. Unless there is a reason why that is not possible?

I agree that `goto_definition` and `find` should be separate, as `find` is a building-block for other functionality too.  But I am pretty sure that `find` should be aware of module imports in order to function properly for the `{F, A}` case.

And when I worked on this I was feeling my way, aiming at functionality, was pretty sure I would get feedback on code organisation.

src/els_code_navigation.erl
```diff
@@ -76,19 +76,29 @@ find([], _Kind, _Data) ->
...
+    [] ->
+      case maybe_imported(Uri, Document, Kind, Data) of
+        {ok, U, P} -> {ok, U, P};
+        {error, not_found} ->
+          find(lists:usort(include_uris(Document) ++ Uris0), Kind, Data)
+      end;
+    Definitions ->
+      {ok, Uri, hd(els_poi:sort(Definitions))}
+  end.
+
```
alanz, line 101: Note: this was only split out because elvis complained about too much indentation otherwise.

src/els_code_navigation.erl
```diff
@@ -104,3 +114,18 @@ add_include_uri(#{ id := Id }, Acc) ->
...
+-spec maybe_imported(uri(), els_dt_document:item(), poi_kind(), any()) ->
+        {ok, uri(), poi()} | {error, not_found}.
+maybe_imported(Uri, Document, function, {F, A}) ->
+  POIs = els_dt_document:pois(Document, [import_entry]),
+  case [{M, F, A} || #{id := {M, FP, AP}} <- POIs, FP =:= F, AP =:= A] of
+    [] -> {error, not_found};
+    [Id|_] ->
+      goto_definition(Uri, #{ kind => application, id => Id
+                            , data => undefined
+                            , range => #{ from => {0, 0}, to => {0, 0}}})
```
alanz, line 128: So this should rather be a recursive call to find, skipping the call to goto_definition. I agree, it felt awkward when I did it.

PR #670, opened 2020-04-16
[#456] Runtime node

### LINKED ISSUE #456: Have a separate Erlang node for each project
The node would represent the "runtime" for the project. It co-operates with the server node and uses the OTP version specified in the `erlang_ls.config`.

robertoaloi:
### Description

This PR introduces an extra Erlang node into the mix. The node represents the _runtime_ for the project. 
The `run_ct_test` command has been converted to use the new runtime as a proof of concept.

There are several advantages in having this decoupling:

* The runtime can now respect the OTP version specified by the user in the `erlang_ls.config` file. As long as the server node and the project node can communicate via RPC, we should be fine.
* It is possible to use a custom Erlang node as the _runtime_. How to start the node and how to interact with it can be customized via the `erlang_ls.config` file. This allows us, among other things, to move the current _code reload_ solution to a more generic solution.
* For `rebar3` projects, we can now re-use code from existing plugins/providers, without re-inventing the wheel (see https://github.com/erlang-ls/rebar3_erlang_ls for more information on the `rebar3` integration). For example, I could drop all the CT code from the codebase.
* It will be possible for the user to customize things such a _formatter_ without having an Erlang LS bloated with dependencies, since a format request will be a RPC call to the agent that can decide on the formatter to use

The configuration code for the `runtime` and `run_ct_test` blocks has been moved from the  `els_config` module (which was growing out of control) to separate ones. The intention is to create a config behaviour later on for the separate blocks.

Fixes #456 .


axelson:
@robertoaloi is there an overview of erlang_ls's architecture/runtime nodes documented anywhere? For [ElixirLS](https://github.com/elixir-lsp/elixir-ls/) we're probably going to move to multi-node (discussion in https://github.com/elixir-lsp/elixir-ls/issues/115) so I want to try to learn from erlang_ls's implementation as much as possible.

robertoaloi:
@axelson Considering this is the first PR in that direction, not yet. But I'm planning to add a section about the Erlang LS architecture in http://erlang-ls.github.io/ once things stabilize. Happy to discuss how Elixir LS is doing and see if we can have something in common (or at least a similar architecture).

axelson:
Okay, sounds good thanks.

rebar.config
```diff
@@ -29,6 +29,7 @@
 {plugins, [ rebar3_proper
           , coveralls
           , {rebar3_lint, {git, "https://github.com/project-fifo/rebar3_lint.git", {tag, "0.1.11"}}}
+          , {rebar3_erlang_ls, {git, "git@github.com:erlang-ls/rebar3_erlang_ls.git", {branch, "master"}}}
```
jfacorro, line 32: Why is this plugin necessary? What benefits does it bring?

src/els_distribution.erl
```diff
@@ -0,0 +1,46 @@
...
+  case net_kernel:connect_node(Name) of
+    true ->
+      lager:info("Connected to existing node [name=~p]", [Name]);
+    false ->
+      Cmd = els_config_runtime:get_start_cmd(),
+      Args = els_config_runtime:get_start_args(),
+      Path = els_config_runtime:get_otp_path(),
+      lager:info("Using Erlang from ~s", [Path]),
+      %% TODO: Do not just spawn
+      spawn(fun() -> els_utils:cmd(Cmd, Args, Path) end)
```
jfacorro, line 27: Shouldn't this clause also try to connect to the node after starting it?

src/els_distribution.erl
```diff
@@ -0,0 +1,46 @@
...
+
+%% @doc Turns a non-distributed node into a distributed one
+-spec start(atom()) -> ok.
+start(Name) ->
+  ok = ensure_epmd(),
+  lager:info("Starting new node [name=~p]", [Name]),
+  net_kernel:start([Name, shortnames]),
+  lager:info("New node started [name=~p]", [Name]).
+
+%%==============================================================================
```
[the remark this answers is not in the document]
robertoaloi: It was a leftover from a previous version. Removed.

src/els_config.erl
```diff
@@ -97,6 +97,8 @@ do_initialize(RootUri, Capabilities, {ConfigPath, Config}) ->
   ExcludePaths = els_utils:resolve_paths(ExcludePathsSpecs, RootPath, true),
   lager:info("Excluded OTP Applications: ~p", [OtpAppsExclude]),
   CodeReload = maps:get("code_reload", Config, disabled),
+  Runtime = maps:get("runtime", Config, #{}),
+  CtRunTest = maps:get("ct-run-test", Config, #{}),
```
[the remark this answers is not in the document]
robertoaloi: I know, but this would imply renaming also the lens itself (and the `show-behaviour-usages` one. I didn't want to do it in the same PR. Created #672 for renaming the lenses.

src/els_distribution.erl
robertoaloi: I guess that is part of the "TODO: Do not just spawn". I want to connect to the node and monitor the connection, but as a follow up. Formalized that TODO in #673 as a follow up.

rebar.config
robertoaloi: The plugin is there so that we can use Erlang LS with Erlang LS itself (since it's a rebar3 project itself).

The purpose of the plugin is to run an agent (a `gen_server`) which has the rebar3 state (paths and the alike) in its own state, so that we can RPC commands to it, effectively using the Erlang API of `rebar3`.

Let's look at how, for example, running CT tests would work for a `rebar3` project without such a plugin.

One option would be to simply start a new Erlang node and run something like `ct:run_test/1` (basically what we had so far, but with a RPC in the mix). This would work for very simple cases, but it would fail as soon as the users would start setting `ct_opts` in their rebar.config or rely on hooks and the alike. We would have to re-implement half of the rebar3 CT provider for things to work correctly.

Another option would be to just `os:cmd` the `rebar3` commands, using the CLI interface rather than the Erlang one. But how would we get back the failing stacktrace in case of errors? Things are just outputted to stdout without our hook (it's not possible to pass custom hooks via the CLI interface of rebar3). We would have to play with group leaders and the alike.

A third option would be to use the `rebar3 shell` as the _runtime node_. The problem there is that the shell also starts apps and running CT tests results in the shell itself being killed.

So, the plugin is a compromise. It's annoying that we would require a rebar3 user to install it to work with the language server, but I guess that's still better than re-implementing half of rebar3 in Erlang LS. For projects which already have the concept of a _running node_, things would just work as long as you have a MFA which can reply via RPC. But I'm still not sure how it would work for a non-rebar3 project which does not have the concept of an _agent_. Open to suggestions, though.

robertoaloi: Maybe we should not default to the rebar3 implementation and keep the basic CT RPC on the server side as a default.

robertoaloi: I will proceed and merge for now. We can eventually follow-up.

Issue #460, opened 2020-02-22
Add go-to-definition for parse transforms


robertoaloi:

maxno-kivra:
This would be lovely ❤️

PR #682, opened 2020-04-20
Detect parse transforms properly.

kape1395:
Compile options in the module is a list, and several parse transforms can be specified there.

### DIFF
--- a/priv/code_navigation/src/diagnostics_parse_transform_usage_list.erl
+++ b/priv/code_navigation/src/diagnostics_parse_transform_usage_list.erl
@@ -1,0 +1,8 @@
+-module(diagnostics_parse_transform_usage_list).
+
+-export([ main/1 ]).
+
+-compile([{parse_transform, diagnostics_parse_transform}, debug_info]).
+
+main(Args) ->
+  ok.
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -245,7 +245,9 @@ attribute(Tree) ->
       [poi(Pos, behaviour, Behaviour)];
     {callback, {callback, {{F, A}, _}}} ->
       [poi(Pos, callback, {F, A})];
-    {compile, {compile, CompileOpts}} ->
+    {compile, {compile, {parse_transform, ParseTransform}}} ->
+      [poi(Pos, parse_transform, ParseTransform)];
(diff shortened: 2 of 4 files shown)


michalmuskala:
AFAIK the `-compile` attribute can contain both - either a bare option or a list of options.

kape1395:
I added a testcase (`els_diagnostics_SUITE:compiler_with_parse_transform_list`), but it somehow interferes with tests `compiler_with_parse_transform` and `compiler_with_parse_transform_included`.

Without my fixes in esl_parser, these three tests are passing (one should fail). If I comment tests `compiler_with_parse_transform` and `compiler_with_parse_transform_included`, then `compiler_with_parse_transform_list` fails as expected. Not sure, where to look for that.

robertoaloi:
@kape1395 You are absolutely right, thanks for spending the time to share that finding. This is what I think happens:

* When you run the new test without your fix in isolation it fails. This is because the parser cannot find the parse transform so later on, when the compiler diagnostics run, the compiler is not aware of the dependency.
* When you run the whole test suite, though, the parser in the previous test case correctly finds the parse transform and loads the parse transform into memory. The new test case passes, because the parse transform is already there.

So, what I think we should do:

* Fix the suite, so that the parse transform is purged between test cases
* Add a test case in the `els_code_navigation_SUITE` where we look for the definition of a parse transform in case it is provided within a list

Let me know if you want to give it a try or if you would like some help. And thanks again for the report, really appreciated!

kape1395:
I fixed the test cases, now they reproduce the error correctly.

Regarding:
> Add a test case in the els_code_navigation_SUITE where we look for the definition of a parse transform in case it is provided within a list

I found no such suite :/ And it looks like navigation to a parse transform is not working or implemented.

robertoaloi:
Thanks. We can probably refactor those purges, but since the code compilation part of the code is going to change due to the introduction of a runtime node, this is perfectly fine for now. Don't worry about the additional test, I was under the impression navigation was already there, but if it's not we can tackle that separately. Thanks!

robertoaloi:
@kape1395 You were indeed right. There is a ticket for navigating to definition for parse transforms, but it's not implemented yet: https://github.com/erlang-ls/erlang_ls/issues/460

PR #686, opened 2020-04-23
Windows CI

filmor:


filmor:
Not actually ready for review, just want to see whether the checks run this way...

robertoaloi:
> Not actually ready for review, just want to see whether the checks run this way...

Converted it to a draft.

.github/workflows/build.yml
```diff
@@ -1,51 +1,69 @@
...
+
     runs-on: ${{ matrix.platform }}
+
     container:
       image: erlang:${{ matrix.otp-version }}
+
     steps:
+    - name: Install Erlang
+      runs-on: windows-latest
+      run: choco install erlang
```
robertoaloi, line 26: Is the idea (in the final version) to re-use the versions of Erlang from the `otp-version` field for Windows as well?

.github/workflows/build.yml
```diff
@@ -1,51 +1,69 @@
...
       uses: actions/cache@v1
       with:
         path: ~/.cache/rebar3/rebar3_*.plt
         key: ${{ runner.os }}-dialyzer-${{ hashFiles(format('{0}{1}', github.workspace, '/rebar.config')) }}
         restore-keys: |
           ${{ runner.os }}-dialyzer-
+
     - name: Compile
-      run: rebar3 compile
+      run: escript rebar3 compile
```
robertoaloi, line 48: Is there any way we can avoid prepending `escript`? Maybe automate the steps from https://www.rebar3.org/docs/getting-started and add them to the provisioning?

.github/workflows/build.yml
filmor: It is, I just need to get anything to run right now...

robertoaloi: Sure. And thanks for the help!

robertoaloi:
@alanz @filmor They mentioned at the Code Beam V that they started running CI for Windows for relx and see what I found:

https://github.com/erlware/relx/blob/master/.github/workflows/wintest.yml#L1

filmor:
I'll give this another go. It doesn't use the tool-cache, yet, will look into that as well :)

robertoaloi:
@filmor I am closing this one for inactivity (it also has conflicts). I am trying to add a simple Windows CI pipeline as well, but that seems to be more complicated than expected. PR will be opened shortly.

PR #687, opened 2020-04-27
Fix indexing on Windows

filmor:
### Description

The indexed data was wrong on Windows (crashes also the tests) due to a mismatch between `file:///c:/some/path` and `file:///c%3a/some/path` URIs. With this, the common test suite finally runs through.


filmor:
@robertoaloi This should be good to go now, I'll try to get Windows CI running, but until then you can take my word for it, that with this patch the test-suite passes ;)

robertoaloi:
Thanks!

PR #690, opened 2020-04-27
[#684] Introduce distribution and group leader servers

### LINKED ISSUE #684: Code lens for running tests should not include percentage in BEGIN message
When running CT test suites, we don't know in advance how long they will take. Better not to include percentages and get a _spinner_ instead.

References:

* https://www.gnu.org/software/emacs/manual/html_node/elisp/Progress.html
* https://github.com/emacs-lsp/lsp-mode/issues/1599

robertoaloi:
### Description

This PR refines the architecture of the _runtime_ node by introducing a _distribution server_ process which monitors the runtime node. For now no automatic action (e.g. a restart) in case the runtime node goes down, but a message is logged on the server-side. Some issues related to the interaction with the runtime node via background jobs have been solved. Specifically:

* Now background jobs allow _infinite progress_ (to use the protocol terminology) for tasks of which the duration is not known in advance. Periodic reports are sent to the client during the task, so that the user has a better feedback about the ongoing progress (e.g. via a spinning wheel in Emacs)
* The restart strategy for background jobs has been relaxed, to avoid the whole server to crash on a background job issue
* I/O produced by the RPC calls is captured via dedicated group leaders

The CT code lens (the first use case for the runtime node) has been integrated with the new solution. 

Fixes #684 .


test/prop_statem.erl
```diff
@@ -198,8 +198,10 @@ did_open(Uri, LanguageId, Version, Text) ->
 did_open_args(_S) ->
   [els_proper_gen:uri(), <<"erlang">>, 0, els_proper_gen:tokens()].
 
-did_open_pre(#{connected := Connected} = _S) ->
-  Connected.
+did_open_pre(#{ connected := Connected
```
robertoaloi, line 201: The protocol establishes that the client cannot send any of these requests before the `initialized`, so preventing this from happening in the property-based tests. This is to avoid some flakyness due to the fact the indexing now happens after the `initialized` and these calls effectively by-pass the server forcing indexing of the file. If the DB is not ready when the `did_open` is sent, things would crash in the background.

src/els_group_leader_server.erl
```diff
@@ -0,0 +1,126 @@
+%%==============================================================================
```
jfacorro, line 1: It's nice to have a remote node, but things seem to start getting really complicated now. 😄

src/els_command_ct_run_test.erl
```diff
@@ -46,8 +47,8 @@ publish_result(Uri, Line, Severity, Message) ->
   els_diagnostics_provider:publish(Uri, [D]),
   ok.
 
--spec run_test(atom(), atom()) -> any().
+-spec run_test(atom(), atom()) -> {ok, binary()} | {error, any()}.
```
jfacorro, line 50: Should we add this contract in the documentation for the function configured to run tests in the remote node?

alanz: I think we are going to have to come up with some clear specs/guidelines/protocols for how the LS node interacts with the runtime one.

Can that be defined via bahaviours?

src/els_group_leader_server.erl
robertoaloi: It is indeed a massive complication to the architecture, especially now that other portions of the code are not using it (e.g. code reload, compilation). It feels very much like when we introduced Mnesia and persistent storage. I suspect it will take a while until we stabilize again. But please report any issue you find with the new setup.

PR #731, opened 2020-05-15
Handle message as string, not as pattern.

kape1395:
The notification sending was failing with badarg
if the notification text had ~p or similar.


robertoaloi:
Well spotted, thanks!

PR #739, opened 2020-05-15
[#550] Introduce OTP 23.0 pipeline

### LINKED ISSUE #550: Add OTP 23 pipeline
PR #536 adds support for features introduced in OTP 23. Let's add a pipeline for OTP 23, so we can eventually test it.

robertoaloi:
### Description

Fixes #550 .


robertoaloi:
```
src/els_uri.erl:34: http_uri:decode/1 is deprecated and will be removed in OTP 25; use use uri_string functions instead
```

Did they deprecate `http_uri:decode/1` without providing an actual alternative?

See https://github.com/erlang/otp/pull/2492 and https://github.com/for-GET/jesse/pull/86/files ?

O_o

PR #743, opened 2020-05-15
[#725] Do not crash if the Uri contains spaces

### LINKED ISSUE #725: Server does not start on emacs for root path with spaces in it.
See https://github.com/emacs-lsp/lsp-mode/issues/1657

robertoaloi:
### Description

Fixes #725 .

### DIFF
--- a/src/els_uri.erl
+++ b/src/els_uri.erl
@@ -31,11 +31,10 @@ module(Uri) ->
 
 -spec path(uri()) -> path().
 path(Uri) ->
-  Uri1 = http_uri:decode(Uri),
   #{ host := Host
    , path := Path
    , scheme := <<"file">>
-  } = uri_string:parse(Uri1),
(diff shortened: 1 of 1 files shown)


filmor:
This broke Windows support again. I won't be submitting any more changes until I have the CI going.

robertoaloi:
Ouch, sorry about that. It's a pain to maintain support for Windows without being able to test this in CI. Maybe we should check with the OTP if they could help providing a GitHub action to support Windows? It should be useful for them as well...

filmor:
No worries, this was mainly an FYI. I'm looking into setting up a `setup` action for Windows (and maybe macOS as well) and will update the CI here when done. One issue I recognised, btw, is that "failure" of the test suites doesn't manifest itself in readable errors, its mostly locking up (probably due to waiting for indexed content of a different name).

PR #755, opened 2020-05-26
Fix crash from asking erl_lint format our own error messages

alanz:
The diagnostic processing calles 'Module:format_error/1' to render diagnostics.
For included files we render these ourselves, and wrap them in a context
message. Do not pass the wrapped messages on to the original fomatter again.

### DIFF
--- a/src/els_compiler_diagnostics.erl
+++ b/src/els_compiler_diagnostics.erl
@@ -153,9 +153,8 @@ diagnostic(_Path, MessagePath, Range, Document, Module, Desc0, Severity) ->
   %% The compiler message is related to an included file. Replace the
   %% original location with the location of the file inclusion.
   %% And re-route the format_error call to this module as a no-op
-  Desc1 = lists:flatten(Module:format_error(Desc0)),
-  Desc = lists:flatten(io_lib:format("Issue in included file (~p): ~s",
-                                     [Line, Desc1])),
+  Desc1 = Module:format_error(Desc0),
+  Desc = io_lib:format("Issue in included file (~p): ~s", [Line, Desc1]),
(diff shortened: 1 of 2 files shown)


src/els_compiler_diagnostics.erl
```diff
@@ -149,8 +152,11 @@ diagnostic(_Path, MessagePath, Range, Document, Module, Desc0, Severity) ->
   InclusionRange = inclusion_range(MessagePath, Document),
   %% The compiler message is related to an included file. Replace the
   %% original location with the location of the file inclusion.
-  Desc = io_lib:format("Issue in included file (~p): ~p", [Line, Desc0]),
-  diagnostic(InclusionRange, Module, Desc, Severity).
+  %% And re-route the format_error call to this module as a no-op
+  Desc1 = lists:flatten(Module:format_error(Desc0)),
```
robertoaloi, line 156: `diagnostic/4` already does the `format_error` and the flattening already, so there should be no need to calculate `Desc1` here and to perform the extra flatten below, right? Can't we just switch the `Module` with `?MODULE`?

alanz: That can work, but you end up with the original error message as per the testsuite update below. To me the new one looks better.

robertoaloi: I think the following will give you the exact error message you get in your test suite:

```
  Desc1 = Module:format_error(Desc0),
  Desc = io_lib:format("Issue in included file (~p): ~s", [Line, Desc1]),
  diagnostic(InclusionRange, ?MODULE, Desc, Severity).
```

alanz: Sorry, I misread the comment as being about formatting the message here at all, vs spurious flatten calls.

PR #763, opened 2020-06-09
[#615] report a warning for a missing or malformed config file

### LINKED ISSUE #615: No warning for malformed erlang_ls.config file
**Describe the bug**

The logic for loading `erlang_ls.config` files is to take a list of locations, attempt to load from them in order, until one succeeds.

It does not distinguish between a file not existing, and one existing but having a YAML syntax error in it.

So making a typo in your config can cause it to load a completely different config file.

**Expected behavior**

I would expect it to report a diagnostic for the first config file it finds, if it has a syntax error in it.

**Context**
 - `erlang_ls` version (tag/sha): 8ff322fddcf162ed0af494704fd9fd31b3ed3268

alanz:
1. If a config file is specified via the config_path setting, send a warning
message if the file is not found.

2. Once a config file is found, if it does not parse as a yaml file, send a
warning.

Fixes #615 .

### DIFF
--- a/src/els_config.erl
+++ b/src/els_config.erl
@@ -174,11 +174,21 @@ handle_cast(_Msg, State) -> {noreply, State}.
 config_paths( RootPath
             , #{<<"erlang">> := #{<<"config_path">> := ConfigPath0}}) ->
   ConfigPath = els_utils:to_list(ConfigPath0),
-  lists:append([ possible_config_paths(ConfigPath)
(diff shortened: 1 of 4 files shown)


robertoaloi:
@alanz What should we do with this one?

alanz:
I think it is useful, will rebase if you agree.

robertoaloi:
@alanz I'm going to implement something equivalent, so closing this one.

Issue #762, opened 2020-06-07
Configuration "include_dirs" doesn't work with 'include_lib" that includes a relative path


amiramix:
**Describe the bug**

The project consists of a few OTP releases with applications, dependencies, and libraries in various folders. The configuration file is placed in the root folder for those releases, e.g.:

erlang_ls.config
erlector_rel/_erlector/lib/... (a few custom OTP applications)
erlector_rel/_erlector/lib/erlector/src/etr_api.erl (a specific file where erlang_ls can't find include)
erlector_rel/_erlector/lib/yolf/include/yolf.hrl (a specific file included in etr_api.erl above)
erlector_re/_erlector/deps/... (few rebar dependencies)

The configuration file is as follows

```
g@venus:~/work % cat erlang_ls.config 
apps_dirs:
  - "erlector_rel/_erlector/deps/*"
deps_dirs:
  - "erlector_rel/_erlector/lib/*"
include_dirs:
  - "erlector_rel/_erlector/lib"
  - "erlector_rel/_erlector/deps"
```

The 'etr_api.erl' has the following include in line 44 which erlang_ls reports as not valid:

    -include_lib("yolf/include/yolf.hrl").

Diagnostic logs show the following request:

    "textDocument": {
      "uri": "file:///home/g/work/erlector_rel/_erlector/lib/erlector/src/etr_api.erl"
    },

And the diagnostic error:

    {
      "range": {
        "start": {
          "line": 43,
          "character": 0
        },
        "end": {
          "line": 44,
          "character": 0
        }
      },
      "severity": 1,
      "source": "Compiler",
      "message": "can\u0027t find include lib \"yolf/include/yolf.hrl\""
    },

The interesting bit is that when hovering over the error the diagnostic dialog shows that the file is present:

can't find include lib "yolf/include/yolf.hrl"
/home/g/work/erlector_rel/_erlector/lib/yolf/include/yolf.hrl
Size: 2.08 kB
Type: Erlang Header (Erlang)
Modified: 06/06/20 13:06
Created: 06/06/20 13:06 

    g@venus:~/work % ls -l erlector_rel/_erlector/lib/yolf/include/yolf.hrl 
    -rw-r--r--  1 g  g  2080  6 Jun 13:06 erlector_rel/_erlector/lib/yolf/include/yolf.hrl

**To Reproduce**

Add "include_lib" to an .erl file that includes the file from a relative folder using a path (see description above).

**Expected behavior**

[...]

PR #764, opened 2020-06-10
[#762] add include dir option of deps for compiler diagnostics

luos:
If an erlang file has a hrl included the compiler diagnostics raises an
error because it is not included in the arguments for the compile call.

Fix needs that the dependency directory is configured correctly in the `erlang_ls.config` file. 

Related #762

### DIFF
--- a/priv/code_navigation/deps/some_dep/include/some_dep.hrl
+++ b/priv/code_navigation/deps/some_dep/include/some_dep.hrl
@@ -1,0 +1,2 @@
+-record(some_dep_record, {field1, field2}).
+
--- a/priv/code_navigation/src/diagnostics_compiler_with_include_of_third_party_dep_header.erl
+++ b/priv/code_navigation/src/diagnostics_compiler_with_include_of_third_party_dep_header.erl
@@ -1,0 +1,11 @@
+-module(diagnostics_compiler_with_include_of_third_party_dep_header).
+
+-include_lib("some_dep/include/some_dep.hrl").
+
+-export([ main/1 ]).
+
+main(_Args) ->
+  #some_dep_record{
+    field1 = 1,
+    field2 = 2
+  }.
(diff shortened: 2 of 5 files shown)


luos:
Hi,

Should I add an example project under `priv`?

alanz:
> Should I add an example project under priv?

If possible do it under `tests/els_diagnostics_SUITE_data/`.  We only use `priv` for stuff that is so broken that rebar cant use it.

*Or*, modify one of the existing projects under `priv`.  But probably better to add a fresh one, for the specific problem.

luos:
Okay, thanks for the answer, added a test.

robertoaloi:
Hi @luos This feature should already be available, but is configuration is probably a bit tricky. Something like this should work for your case:

```
include_dirs:
  - "deps/*/include"
  - "deps"
```

But we could consider this to avoid the manual step. Any thought @jfacorro ?

Also, this should all go away once we integrate with the BSP part.

luos:
Okay, thanks, I'll check it out and report back if it works with that settings.

Yes, it would be better to automagically pick up Erlang MK / Rebar style structure, but if it needs to be configured that is fine as well. :-)

Thanks

robertoaloi:
@luos This PR has been sitting here for a huge amount of time, so sorry about that! Is this change something you still need? Should we solve conflicts?

luos:
Hi,

Let me check it again, I believe maybe it was solved in one of the newer versions as I have not run into this since.

Thanks,

luos:
Hi,

Sorry for the late reply. It still does not work for me. 

Pull this repo: 
https://github.com/rabbitmq/rabbitmq-metronome/

Pull the deps by running `make` or `gmake`, probably it will get stuck but the relevant dep is downloaded fine. 

My `erlang_ls.config` file is the following:

```
include_dirs:
  - "deps/*/include"
  - "deps"
  - "include"
  ```
  
  The error I get is this:
  
kép

Thanks

**Comment by robertoaloi** [comment]
Hi! The following should work:

```
deps_dirs:
  - "deps/*"
include_dirs:
  - "deps"
  - "deps/*/include"
  - "include"
```

I know, it's awkward right now. One entry is for the `include`, the other one for `include_lib`

robertoaloi:
Closing for lack of activity. The proposed solution should work without requiring changes. If that's not the case, feel free to reopen.

PR #766, opened 2020-06-11
[#765] Add commit message template.

### LINKED ISSUE #765: process : add a git commit message template
Our [contributor guidelines](https://github.com/erlang-ls/erlang_ls/blob/master/.github/CONTRIBUTING.md#commit-messages) request a particular format for commit messages.

Add a template to the repo to nudge contributors toward using this.

alanz:
This helps nudge contibutors to follow the contribution guidelines, as per
https://github.com/erlang-ls/erlang_ls/blob/master/.github/CONTRIBUTING.md#commit-messages

Fixes #765 .

### DIFF
--- a/.gitmessage
+++ b/.gitmessage
@@ -12,4 +12,4 @@
 # - Prefix every commit message with the issue number in the form [#NNN].
 # - Use imperative form (prefer Move function... over Moves function).
 # - Avoid past tense.
-#  -Use a final dot.
+# - Use a final dot.
--- a/.gitmessage
+++ b/.gitmessage
@@ -1,0 +1,15 @@
+# [#NNN] (If applied, this commit will...) <subject> (Max 72 char)
+# |<----   Preferably using up to 50 chars   --->|<------------------->|
+# Example:
+# [#765] Add commit message template.
+
+# Explain why this change is being made
+# |<----   Try To Limit Each Line to a Maximum Of 72 Characters   ---->|
+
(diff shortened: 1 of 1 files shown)


.gitmessage
```diff
@@ -0,0 +1,15 @@
...
+# Explain why this change is being made
+# |<----   Try To Limit Each Line to a Maximum Of 72 Characters   ---->|
+
+# From https://github.com/erlang-ls/erlang_ls/blob/master/.github/CONTRIBUTING.md#commit-messages
+#
+# - Use descriptive commit messages.
+# - Prefix every commit message with the issue number in the form [#NNN].
+# - Use imperative form (prefer Move function... over Moves function).
+# - Avoid past tense.
+#  -Use a final dot.
```
robertoaloi, line 15: Nit: missing space after the dash.

.gitmessage
```diff
@@ -0,0 +1,15 @@
+# [#NNN] (If applied, this commit will...) <subject> (Max 72 char)
```
robertoaloi, line 1: This template would need to be manually installed by the user or referred to via a `-t` option, right? We should somehow inform the users about the existence of this file.

robertoaloi: Maybe we could simply have a linter via a git commit hook or a GitHub action...

alanz: I think having both. It is better to use the template when composing the original commit, that is when you need the prompt. Otherwise it needs amend, force push, etc.
Ideally we should configure it as a feature of the git repo, but I think it does need to be set up per repo, via `git config commit.template .gitmessage`

PR #776, opened 2020-07-08
[#775] epp error messages for hrl files

### LINKED ISSUE #775: The diagnostics process crashes for non-existent macro in an include file
**Describe the bug**

Attempting to open or index a header file with an undefined macro in it leads to a crash dump.

**To Reproduce**

Create a `.hrl` file with the following contents

```erlang
-if(?MODULE == module_foo).  % TODO get rid of
-else.
-error("including nonexistent_macro.hrl is not allowed").
-endif.
```
Open it using erlang_ls, and look at the log files

**Expected behavior**

A diagnostic saying the macro is undefined

**Actual behavior**

A crash log

```
2020-07-08 14:26:15 =SUPERVISOR REPORT====
     Supervisor: {local,els_background_job_sup}
     Context:    child_terminated
     Reason:     {badarg,[{erlang,apply,[4,format_error,[{undefined,'MODULE',none}]],[]},{file,format_error,1,[{file,"file.erl"},{line,159}]},{els_compiler_diagnostics,diagnostic,4,[{file,"/Users/alanzimm/repos/erlang_ls/erlang_ls/src/els_compiler_diagnostics.erl"},{line,163}]},{els_compiler_diagnostics,'-parse/1-lc$^0/1-0-',1,[{file,"/Users/alanzimm/repos/erlang_ls/erlang_ls/src/els_compiler_diagnostics.erl"},{line,95}]},{els_compiler_diagnostics,parse,1,[{file,"/Users/alanzimm/repos/erlang_ls/erlang_ls/src/els_compiler_diagnostics.erl"},{line,96}]},{els_background_job,handle_info,2,[{file,"/Users/alanzimm/repos/erlang_ls/erlang_ls/src/els_background_job.erl"},{line,155}]},{gen_server,try_dispatch,4,[{file,"gen_server.erl"},{line,637}]},{gen_server,handle_msg,6,[{file,"gen_server.erl"},{line,711}]}]}
     Offender:   [{pid,<0.338.0>},{id,els_background_job},{mfargs,{els_background_job,start_link,undefined}},{restart_type,temporary},{shutdown,5000},{child_type,worker}]
```
A clear and concise description of what happens instead.

**Context**
 - `erlang_ls` version (tag/sha): 0fa9f260cdf88c73f6be34baa3af9e52183c4434
 - Editor used: emacs 26.3 (macos)
 - LSP client used: lsp-mode

alanz:
### Description

Generate diagnostics for preprocessor failure on a `.hrl` file

When parsing a header file with `epp:parse`, there are different ways errors can
be reported.

Make sure that all of them are correctly turned into diagnostics.

Fixes #775 .


src/els_compiler_diagnostics.erl
```diff
@@ -92,11 +92,27 @@ parse(Uri) ->
...
+-spec epp_diagnostic(integer(), module(), any()) ->
+        els_diagnostics:diagnostic().
+epp_diagnostic(Line, epp, {error, Desc}) ->
+  diagnostic(range(Line), epp, Desc, ?DIAGNOSTIC_ERROR);
+epp_diagnostic(Line, epp, {error, Line, Desc}) ->
+  diagnostic(range(Line), epp, Desc, ?DIAGNOSTIC_ERROR);
+epp_diagnostic(Line, epp, Desc) ->
+  diagnostic(range(Line), epp, Desc, ?DIAGNOSTIC_ERROR);
+epp_diagnostic(Line, erl_parse, Desc) ->
+  diagnostic(range(Line), erl_parse, Desc, ?DIAGNOSTIC_ERROR).
```
michalmuskala, line 114: I think we only need to handle the `{error, Line, Desc}` case specifically. It seems to be a bug in `epp:format_error/1` that it crashes for that. The other cases should work just fine [looking at the implementation](https://github.com/erlang/otp/blob/4f580ad1bbc18d798dd3391bab7c714902f9b996/lib/stdlib/src/epp.erl#L189-L233). Especially `{error, Desc}` that is formatted by epp differently than it would be in here.

michalmuskala: Reported upstream https://bugs.erlang.org/browse/ERL-1310

alanz: The comments in line 101-104 are actual cases seen, from the test.  So the other cases *are* needed, I think.

michalmuskala: The other cases should work just fine with what `diagnostic` already does - i.e. calling `Mod:format_error(Reason)`. It's just that this one seems to crash on that call.

PR #780, opened 2020-08-10
[#777] Compiler diagnostics do not find transitive dependencies.

### LINKED ISSUE #777: compiler diagnostics do not find transitive dependencies.
**Describe the bug**

If a file implements a behaviour, and that behaviour in turn implements another one, the diagnostics `load_dependency` function does not load the second one, so we report spurious warnings.

**Context**
 - `erlang_ls` version (tag/sha): 9e5034e9359971e5ffa61771b0d70d47fc5b2209
 - Editor used:
 - LSP client used:

alanz:
If a file implements a behaviour, and that behaviour in turn implements another
one, the diagnostics load_dependency function does not load the second one, so
we report spurious warnings.

Fixes #777.


alanz:
Note: I have not added tests as I was unable to reproduce the problem in the erlang_ls test setup.

I have confirmed that there are files in our private code base that show up with warnings without this patch, and load fine with it, so I am confident it does what is required.

robertoaloi:
@alanz Shouldn't this be reproducible by having a module which uses a behaviour which, in turns, uses a behaviour?

alanz:
@robertoaloi I had a test like that, but it loads fine. 

Our specific case comes about from a monorepo, and a behaviour in one app using a behaviour in another.  And after spending days trying to reproduce it, I thought to just try for a merge :).

alanz:
And looking more closely, renaming to protect the innocent, is is something like

```
./appa/src/appa.erl
./appb/src/behaviour_b.erl
./appc/src/behaviour_c.erl
./appc/src/behaviour_d.erl
```

`appa` uses `behaviour_b`

`behaviour_b` uses `behaviour_c` and `behaviour_d`

The problem shows up that `appa` has a warning that it cant load `behaviour_b`.

robertoaloi:
@alanz Could you please let me have a minimal example (privately if you prefer), so I can try and reproduce it? I would like to understand the root cause before a blind merge.

alanz:
@robertoaloi  FYI I am still struggling to reproduce this issue.  I have updated the diff to include what I *thought* was the problem, abstracted from the work codebase.  But not able to reproduce the problem here.  But it *does* show up in the real code base.

I have spent far too much time on this blessed thing.

robertoaloi:
@alanz Closing this since we don't have a reproducible example and a bunch of problems regarding compiler diagnostics have been solved in the past month. Feel free to reopen if you manage to reproduce the error.

PR #784, opened 2020-09-13
[#771] Do not ask for text on file save

### LINKED ISSUE #771: Do not ask for text on file save
With version 0.4.1, we set the server capabilities to include text on save

Screenshot 2020-06-25 at 16 26 26

But we always process the files from disk.  Until such time as we actually use that text, we should not ask for it.

dr0pdb:
Signed-off-by: Saurav Tiwary <srv.twry@gmail.com>

### Description

This PR disable the `includeText` flag while saving a text document since the files are processed from the disk hence the text isn't required.

Fixes #771


robertoaloi (approved):

robertoaloi:
Thanks!

PR #787, opened 2020-09-18
Added loop detection into dependency discovery

srstrong:
### Description
If there is a dependency loop in include files, then els_diagnostics_utils:dependencies/1 would infinitely recurse.  

I've added a simple set of previously-visited uris to prevent such loops from occuring


robertoaloi:
Thanks!

robertoaloi:
As a plus, it would be good to have a test case which reproduces the loop, so we can avoid introducing a regression later on.

PR #791, opened 2020-09-30
[DAP] add ability to launch in integrated terminal

alanz:
And stub out some more functions.

`stepOut` currently does the same as `step`.

### DIFF
--- a/SPAWNFEST.md
+++ b/SPAWNFEST.md
@@ -113,6 +113,64 @@ While working on the project, we noticed a few issues with third-party projects 
 
 (require 'dap-mode)
 
+;; Log io to *Messages*. Seems to be the only option at the moment. Optional
+(setq dap-inhibit-io nil)
+(setq dap-print-io t)
+
+(defun dap-erlang--populate-start-file-args (conf)
+  "Populate CONF with the required arguments."
+  (-> conf
+      (dap--put-if-absent :dap-server-path '("els_dap"))
+      (dap--put-if-absent :request "launch")
+      (dap--put-if-absent :projectDir (lsp-find-session-folder (lsp-session) (buffer-file-name)))
+      (dap--put-if-absent :cwd (lsp-find-session-folder (lsp-session) (buffer-file-name)))
+      ))
+
+;; Add a Run Configuration for running 'rebar3 shell'
+(dap-register-debug-provider "Erlang" 'dap-erlang--populate-start-file-args)
+(dap-register-debug-template "Erlang rebar3 shell"
+                             (list :type "Erlang"
+                                   :program "rebar3"
+                                   :args "shell"
+                                   :name "Erlang::Run"))
+
+;; Add a Run Configuration for executing a given MFA
+(dap-register-debug-template "Erlang MFA"
+                             (list :type "Erlang"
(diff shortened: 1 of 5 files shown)


SPAWNFEST.md
```diff
@@ -111,6 +111,64 @@ While working on the project, we noticed a few issues with third-party projects
...
+                                   :function "fact"
+                                   ;; :function "dummy"
+                                   :args "[4]"
+                                   :name "Erlang::Terminal MFA"))
+
+
+
+
+
 (require 'dap-mode)
```
robertoaloi, line 172: This seems to be duplicated now

SPAWNFEST.md
```diff
@@ -111,6 +111,64 @@ While working on the project, we noticed a few issues with third-party projects
...
+                                   ;; :function "dummy"
+                                   :args "[3]"
+                                   :name "Erlang::Run MFA"))
+
+;; Add a Run Configuration for running 'rebar3 shell'
+;; in an integrated terminal in the client
+;; NOTE: set the projectnode hostname in two places
+(dap-register-debug-template "Erlang Terminal"
+                             (list :type "Erlang"
+                                   :runinterminal '("rebar3" "shell" "--name" "dapnode@alanzimm-mbp")
```
robertoaloi, line 151: We should not hard-code the name. Maybe use a variable instead? I know it's the `Spawnfest.md`, but better not to have hard-coded values left around, in case we move this to an actual sample config.

SPAWNFEST.md
```diff
@@ -111,6 +111,64 @@ While working on the project, we noticed a few issues with third-party projects
...
+                                   :projectnode "dapnode@alanzimm-mbp"
+                                   :name "Erlang::Terminal"))
+
+;; Add a Run Configuration for running 'rebar3 shell'
+;; in an integrated terminal in the client, running a given MFA
+;; NOTE: set the projectnode hostname in two places
+(dap-register-debug-template "Erlang Terminal MFA"
+                             (list :type "Erlang"
+                                   :runinterminal '("rebar3" "shell" "--name" "dapnode@alanzimm-mbp")
```
robertoaloi, line 160: Same comment as above.

src/els_distribution_server.erl
```diff
@@ -51,7 +51,10 @@ start_distribution(Name) ->
     {ok, _Pid} ->
       lager:info("Distribution enabled [name=~p]", [Name]);
     {error, {already_started, _Pid}} ->
-      lager:info("Distribution already enabled [name=~p]", [Name])
+      lager:info("Distribution already enabled [name=~p]", [Name]);
+    {error, {{shutdown, {failed_to_start_child, net_kernel, E1}}, E2}} ->
+      lager:info("Distribution shutdown [errs=~p]", [{E1, E2}]),
```
robertoaloi, line 56: Can we collapse this into a single message?

robertoaloi:
```
lager:info("Distribution shut down [name=~p] [errors=~p]" ...
```

robertoaloi (approved):
Some minor comments which would be good to address. Looks good otherwise.

PR #792, opened 2020-10-02
Assume invoked trigger kind

hauleth:
Not all LS clients send `context` as this is not required by the specs. So instead of failing to return any completions it will just assume manually triggered completion.

In addition it allows completion after just plain trigger character even when invoked manually.


robertoaloi:
It sounds reasonable, thanks!

robertoaloi:
The CI issue seems unrelated. Addressing it separately.

PR #801, opened 2020-10-11
[#800] Ensure the path is recognized by OTP 23.1 as an absolute one

### LINKED ISSUE #800: CI pipeline for OTP 23.1 is broken
In OTP 23.1 the CI pipeline is broken, due to a bunch of failing test suites. An example is the `els_diagnostics_SUITE`, where the wrong number of diagnostics is returned. This is due to the `include_paths` passed to the `compile:file/2` being empty.
After a small investigation, it turned out that the underlying cause is this bug fix for `uri_string:recompose/1` in OTP 23.1:

https://github.com/erlang/otp/commit/40efd733e0fc643668a368e3f2860583d7ceb521

The new code attempts to convert the `path` to an absolute one, whenever a `host` component is provided and a relative path is provided. In the case of Erlang LS, it tries to convert an already absolute path, adding an extra `/`. I suspect this could be an actual bug in the new implementation, but as a workaround, we can pass a `/` character instead of a binary. Will verify with OTP if this is a bug or not.

robertoaloi:
Reference: erlang/otp@40efd73

Fixes #800 .

### DIFF
--- a/src/els_uri.erl
+++ b/src/els_uri.erl
@@ -69,7 +69,7 @@ uri(Path) ->
     uri_string:recompose(#{
       scheme => <<"file">>,
       host => Host,
-      path => [<<"/">>, Path1]
+      path => [$/, Path1]
     })
   ).
 


robertoaloi:
Opened a PR upstream: https://github.com/erlang/otp/pull/2795

PR #803, opened 2020-10-17
[#305] Add Windows CI pipeline

### LINKED ISSUE #305: Add Windows platform to CI

robertoaloi:
### Description

There are still a few issues with the pipeline:

* The pre-compile hook fails
* rebar3 detects the error, but terminates with a zero status code

Fixes #305 .

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,7 +8,7 @@ on:
       - opened
       - synchronize
 jobs:
-  build:
+  linux:
     strategy:
       matrix:
         platform: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


alanz:
Is this still valid/relevant?

robertoaloi:
> Is this still valid/relevant?

I think so. A Windows pipeline would help avoid regressions for users.

.github/workflows/build.yml
```diff
@@ -56,3 +56,22 @@ jobs:
...
+      uses: actions/checkout@v2
+    - name: Install Erlang
+      run: choco install -y erlang --version ${{ matrix.otp-version }}
+    - name: Install rebar3
+      run: choco install -y rebar3 --version 3.13.1
+    - name: Compile
+      run: rebar3 compile
+    - name: Run CT Tests
+      run: rebar3 ct
+    - name: Run PropEr Tests
```
TheGeorge, line 76: should we upload the logs for debugging?

jfacorro:
Nice! I will use this for clojerl!

PR #807, opened 2020-11-19
[#770] Run mnesia in RAM-only mode

### LINKED ISSUE #770: Discussion : should we trash the db on startup?
At the moment, the startup process hangs while we wait for mnesia tables to be ready. For any decently sized code base, this wait can be quite long, in the order of minutes.

But we are now able to do indexing while the application is running, and index on-demand as a file is opened.  We also provide very clear feedback on the state of indexing vie `$/progress` messages.

So perhaps we will be better off just deleting the db on startup, so the IDE becomes responsive sooner, and the user can wait for indexing for full search capabilities.

Other possible alternatives are

a) use some other db instead of mnesia
b) use mnesia for core (small) information, and store the rest in files in the filesystem (perhaps using LSIF?).

`b)` is probably a very bad option, as it is basically rolling our own db.

alanz:
1. We already store everything in RAM, with disc copies only.

2. The time to load tables on startup is comparable to the initial indexing time.

3. During indexing erlang_ls is still useful, while waiting for tables it does
not process any messages.

4. If we end up with multiple instances of erlang_ls running against the same
code base, we will not get contention for the mnesia database directory. (Note:
this can happen when running under vscode in remote mode).

Fixes #770

### DIFF
--- a/src/els_db.erl
+++ b/src/els_db.erl
@@ -5,8 +5,7 @@
         , clear_tables/0
         , delete/2
         , delete_object/1
-        , dump_tables/0
-        , install/2
+        , install/0
         , lookup/2
         , match/1
         , stop/0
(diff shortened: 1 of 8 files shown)


robertoaloi:
Hi @alanz Did you try the `https://github.com/erlang-ls/erlang_ls/tree/remove-mnesia` branch? There I'm just using ETS as a backend. Mnesia is already a stretch and, if we are also removing persistency, it really doesn't make sense to use Mnesia at all. I would rather focus on that branch and move to ETS (with an optional config to use DETS) than switching to memory-only tables. Thoughts?

robertoaloi:
ETS-only version up for review: https://github.com/erlang-ls/erlang_ls/pull/809
That should supersede this PR.

robertoaloi:
Closing this as superseded by #809

PR #809, opened 2020-11-20
[#770] Switch back from Mnesia to simple ETS

### LINKED ISSUE #770: Discussion : should we trash the db on startup?

robertoaloi:
### Description

For now, let's switch back to plain ETS. We can then follow up by adding persistence (via DETS) if necessary. In that case, we could have a dedicated option in the `erlang_ls.config` file to enable it. Before adding persistency, we should really check if it's worth, performance-wise.

All write operations to the ETS tables are serialized via a `gen_server`. Reads happen straight from the ETS table, which has `read_concurrency` enabled, even if it shouldn't matter much in Erlang LS.

Fixes #770 .

### DIFF
--- a/src/els_dt_document.erl
+++ b/src/els_dt_document.erl
@@ -72,12 +72,7 @@ name() -> ?MODULE.
 
 -spec opts() -> proplists:proplist().
 opts() ->
-  [ {attributes        , record_info(fields, els_dt_document)}
-  , {disc_copies       , [node()]}
-  , {index             , []}
-  , {type              , set}
-  , {storage_properties, [{ets, []}]}
-  ].
+  [].
 
 %%==============================================================================
 %% API
@@ -118,7 +113,7 @@ to_item(#els_dt_document{ uri  = Uri
 -spec insert(item()) -> ok | {error, any()}.
 insert(Map) when is_map(Map) ->
   Record = from_item(Map),
-  els_db:write(Record).
+  els_db:write(name(), Record).
 
 -spec lookup(uri()) -> {ok, [item()]}.
 lookup(Uri) ->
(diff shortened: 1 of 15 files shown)


src/els_dt_document.erl
```diff
@@ -72,12 +72,7 @@ name() -> ?MODULE.
 
 -spec opts() -> proplists:proplist().
 opts() ->
-  [ {attributes        , record_info(fields, els_dt_document)}
-  , {disc_copies       , [node()]}
-  , {index             , []}
-  , {type              , set}
-  , {storage_properties, [{ets, []}]}
-  ].
+  [].
```
alanz, line 75: Is `set` the default?  Should we be explicit?

alanz (approved):
LGTM

Given we now use unique node-names, I suggest removing the SHA value, and using the base of the project root instead.

robertoaloi:
> LGTM
> 
> Given we now use unique node-names, I suggest removing the SHA value, and using the base of the project root instead.

Agreed, and we can also sync the DAP node name, which is today based on the SHA of the CWD. Will do it separately, though.

src/els_dt_document.erl
robertoaloi: Yes, `set` is the default type. We can be explicit, will do it as a separate PR (I want to change something else).

PR #810, opened 2020-11-26
[#770] Be explicit on table type

robertoaloi:
### Description

`set` is the default type for tables, but let's be explicit.


alanz (approved):
LGTM

PR #811, opened 2020-11-26
Make doc chunk search use els db instead of code path

garazdawi:
I recently started using erlang_ls and found an itch with it that I needed to scratch.

Before this change, any els hover request used only the path of the running VM to find the doc chunks. This made using `otp_path` to point out which OTP version to use not work for hover requests.

With this change, the hover provider now looks using els db for the .beam and .chunk file in the local filesystem.

I'm not sure how to add tests for this, any suggestions? I looked around but it seems like very few other testcases change the `otp_path`.


robertoaloi:
This makes a lot of sense. Testing in this area is still very weak and things will probably change once we start using a "build server", but for now I believe this is a long due improvement. Thanks!

PR #813, opened 2020-11-27
[#812] Use project basename for ELS and DAP node names

### LINKED ISSUE #812: Refactor and harmonize node names for distribution
For a given project, ELS and DAP node names are generated using a SHA of the RootUri and OtpPath. Furthermore, the SHA used by ELS and DAP for the same project is different, for no reason.

Now that node names are unique using a `unique_integer`, it would be much more convenient to use the project basename as part of the node name (eg when looking for a specific project node).

* Replace the SHA with the project basename
* Have DAP and ELS use the same name format

robertoaloi:
### Description

Fixes #812

### DIFF
--- a/src/els_distribution_server.erl
+++ b/src/els_distribution_server.erl
@@ -168,10 +168,9 @@ ensure_epmd() ->
   0 = els_utils:cmd("epmd", ["-daemon"]),
   ok.
 
--spec node_name(string(), binary()) -> atom().
-node_name(Prefix, Binary) ->
-  <<SHA:160/integer>> = crypto:hash(sha, Binary),
-  Int = erlang:unique_integer([positive]),
-  Id = lists:flatten(io_lib:format("~40.16.0b_~p", [SHA, Int])),
+-spec node_name(binary(), binary()) -> atom().
+node_name(Prefix, Name) ->
+  Int = erlang:phash2(erlang:timestamp()),
+  Id = lists:flatten(io_lib:format("~s_~s_~p", [Prefix, Name, Int])),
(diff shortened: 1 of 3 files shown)


alanz (approved):
LGTM.

src/els_distribution_server.erl
```diff
@@ -168,10 +168,9 @@ ensure_epmd() ->
...
   ok.
 
--spec node_name(string(), binary()) -> atom().
-node_name(Prefix, Binary) ->
-  <<SHA:160/integer>> = crypto:hash(sha, Binary),
-  Int = erlang:unique_integer([positive]),
-  Id = lists:flatten(io_lib:format("~40.16.0b_~p", [SHA, Int])),
+-spec node_name(binary(), binary()) -> atom().
+node_name(Prefix, Name) ->
+  Int = erlang:phash2(erlang:timestamp()),
```
alanz, line 173: If we used a string representation instead of the phash2, either as a time or of the underlying integers, we could see which process was "older". Or is it some meaningless large negative number?

alanz (approved):

src/els_distribution_server.erl
robertoaloi: The `monotonic_time` was returning a negative number, that's why I switched for this. `erlang:timestamp/0` returns a tuple. The "age" of a process should still be visible via `ps`. Also, I suspect the port numbers displayed by `epmd -names` are somewhat incremental.

alanz: ok, happy with this as is then.

PR #814, opened 2020-11-30
Rename xref diagnostics to crossref

alanz:
This is an internal erlang_ls diagnostic, based on the results of indexing.

The name 'xref' causes people to think it is a version of the 'xref' tool.

This also opens the way to provide an integration of the actual 'xref' tool at a
fututure date.


robertoaloi (approved):

robertoaloi:
Thanks @alanz . Would you mind updating the docs in erlang-ls.github.io to reflect the change?

PR #816, opened 2020-11-30
[#404] Add experimental diagnostics to detect unused included files

### LINKED ISSUE #404: Detect unused include/include_lib
I would like to detect unused include/include_lib.

If an include/include_lib directive is unused in a module, it should be marked as a warning.

**Additional context**
This could be implemented by trying to compile the same file multiple times, excluding one include/include_lib at a time. Since this could be expensive, it could be triggered on demand by the user.

robertoaloi:
### Description

This approach is not particularly performant and it detects a few false positives, but - considering this is disabled by default and currently undocumented -, I believe it's a step in the right direction. We can refine the algorithm incrementally.

Fixes #404 .

### DIFF
--- a/include/erlang_ls.hrl
+++ b/include/erlang_ls.hrl
@@ -579,6 +579,7 @@
                    | include_lib
                    | macro
                    | module
+                   | parse_transform
                    | record
                    | record_access
                    | record_expr
--- a/src/els_compiler_diagnostics.erl
+++ b/src/els_compiler_diagnostics.erl
@@ -16,6 +16,10 @@
 %% identity function for our own diagnostics
 -export([ format_error/1 ]).
 
+-export([ inclusion_range/2
+        , inclusion_range/3
+        ]).
+
 %%==============================================================================
 %% Includes
 %%==============================================================================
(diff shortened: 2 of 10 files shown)


include/erlang_ls.hrl
```diff
@@ -579,6 +579,7 @@
                    | include_lib
                    | macro
                    | module
+                   | parse_transform
```
robertoaloi, line 582: This was already  available as a `poi_kind()` but missing in the type definition. The new code highlighted the issue.

src/els_compiler_diagnostics.erl
```diff
@@ -241,7 +245,9 @@ include_lib_id(Path) ->
   Length     = length(Components),
   End        = Length - 1,
   Beginning  = max(1, Length - 2),
-  filename:join(lists:sublist(Components, Beginning, End)).
+  [H|T]      = lists:sublist(Components, Beginning, End),
+  %% Strip the app version number from the path
```
robertoaloi, line 249: Not stripping the eventual app number was causing the warning to be displayed on the first line of the module (fallback).

PR #819, opened 2020-12-02
[#818] Allow navigation to user defined types and records from within types

### LINKED ISSUE #818: Record definitions cannot be navigated from types
**Describe the bug**
Discovered while working on #815 

Given the following:
```erlang
-record(foo, {bar, baz}).
-type foo() :: #foo{}.
```
It is not possible to navigate to the definition of `#foo{}`.

**Expected behavior**
It should be possible to navigate to the definition of `#foo{}`.

**Context**
 - `erlang_ls` version (tag/sha): [#815 branch](https://github.com/onno-vos-dev/erlang_ls/tree/815-record-definitions-cannot-be-navigated-from-type-specs)
 - Editor used: Emacs
 - LSP client used: 20201129.1832

onno-vos-dev:
Fixes #818

### DIFF
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -99,6 +99,8 @@ find_attribute_pois(Tree, Tokens) ->
               || {{F, A}, {atom, Pos, _}} <- lists:zip(Exports, Atoms)
             ],
           [find_attribute_tokens(Tokens), ExportTypeEntries];
+        {type, Type} ->
+          type_to_poi(Type);
         _ -> []
       catch
         throw:syntax_error ->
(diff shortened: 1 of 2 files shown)


robertoaloi:
Bedankt! As a follow up, we should include the other cases listed in `do_find_spec_points_of_interest` (e.g. `remote_type`) and refactor the code so that your new function is used when finding type applications from specs.

onno-vos-dev:
@robertoaloi I was considering the refactoring but wanted to keep this diff to a minimal. I'll follow up on this over the next few days 👍

Issue #154, opened 2019-10-26
Include link to OTP documentation on hover


robertoaloi:
When hovering  on OTP library functions, a link to the relevant official OTP documentation page should be included at the bottom of the type specs / edoc.

The location of the documentation should be configurable (local man pages, local html pages) and maybe default to remote OTP pages.

robertoaloi:
@dgud What do you think about having this as a fallback in the case a user is using OTP < 23 ? Relates to #174

dgud:
This sounds good to me, but for me the emacs html rendering and usage is still to bad, I have tried to use it with the already available 'erldoc.el' but I couldn't make anchors work in emacs-25 with the builtin browser.

 #174 might be a bit much work when in a year everybody will be on 23
( except you guys :-)

alanz:
@dgud 
> but for me the emacs html rendering and usage is still to bad
What version of emacs are you running?  Are you running it in a terminal, or as a gui?

dgud:
I have used emacs25 in a gui, i.e. X-window or the gui variant on windows, 
now I run emacs26 (since there are markdown rendering problems in 25) 
have not tested html handling with that version.

alanz:
@dgud I think the markdown stuff in the gui needs emacs 26, which I am running (on debian and mac) and it works for me.

dgud:
I understand that but I'm afraid that it will lead to a bad first impressions.
Ubuntu LTS (i.e. currently 18.04) which is what most linux users use (at least on Ericsson)
have only emacs25 as a package.

Now I'm working from home (for other reasons) using windows and WSL which uses ubuntu 18.04 so first thing I had todo was to fix emacs26.

Now Ubuntu 20.04 should be ready soon and I guess that will have emacs26 so I guess that problem is soon solved.

robertoaloi:
Emacs 27 is almost out btw :)

PR #830, opened 2020-12-11
[#631] Show function clauses on hover

### LINKED ISSUE #631: Show function clauses on hover

robertoaloi:
Screenshot 2020-12-11 at 20 27 42

Screenshot 2020-12-11 at 20 44 26

* Facelift for docs on hover
* Include function clauses on hover
* Include clickable links to OTP docs for OTP exported functions on hover
* Refactor docs handling into separate module
* Separate docs fetching from formatting
* Refactor test code for hover into separate modules

Fixes #631 and #154 .


michalmuskala:
The list of clauses should probably be limited - there can be functions with hundreds if not thousands of clauses, printing them all would probably be slow, and probably not very informative.

robertoaloi:
> The list of clauses should probably be limited - there can be functions with hundreds if not thousands of clauses, printing them all would probably be slow, and probably not very informative.
> 
> E.g. Elixir limits clauses printed to 10 when diffing patterns in `function_clause` errors.

Makes sense, I can limit that.

PR #832, opened 2020-12-18
Fix column numbers

garazdawi:
I've been reviewing https://github.com/erlang/otp/pull/2664 and found that erlang ls breaks some abstractions when it comes to handling `erl_anno` annotations. This PR fixes those.

I also start using the "fixed" epp:open/1 instead of the undocumented epp:open/5 for parsing escripts. https://github.com/erlang/otp/pull/2664 has not been merged yet, so the API may still change, but I thought we may as well fix it now.


robertoaloi:
Amazing, thanks!

PR #833, opened 2020-12-18
Add column numbers

garazdawi:
This PR adds rudimentary support for giving error location information based on the column number returned by the erlang compiler.

This PR depends on https://github.com/erlang/otp/pull/2664 and https://github.com/erlang-ls/erlang_ls/pull/832.

I decided to use the poi that the column is pointing to as the object to mark as the compiler only emits the starting column and not the end column.

Using the poi is not ideal as sometimes it can select a much larger section than it should, the most obvious example is when an attribute is broken somehow. It is however better than trying to come up with some other heuristic about which range it is that is responsible for the warning.

I've put this PR as a draft as I'm not happy about the usage of `poi`s. So I'm looking to see if anyone has any better suggestions? :)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         platform: [ubuntu-latest]
-        otp-version: [21.3, 22.2, 23.1]
+        otp-version: [21, 22, 23, 24]
     runs-on: ${{ matrix.platform }}
     container:
       image: erlang:${{ matrix.otp-version }}
--- a/apps/els_core/src/els_escript.erl
+++ b/apps/els_core/src/els_escript.erl
@@ -149,17 +149,7 @@ parse_source(S, File, Fd, StartLine, HeaderSz) ->
   _ = io:get_line(Fd, ''),
   Encoding = epp:set_encoding(Fd),
   {ok, _} = file:position(Fd, HeaderSz),
-  {ok, Epp} =
-        case erlang:function_exported(epp, open, 5) of
-            true ->
-                %% Pre OTP-24 function to call for escripts
-                %% Done via apply in order to silence dialyzer
-                apply(epp, open, [File, Fd, StartLine,
-                                  IncludePath, PreDefMacros]);
-            false ->
-                epp:open([{fd, Fd}, {name, File}, {location, StartLine},
-                          {includes, IncludePath}, {macros, PreDefMacros}])
-        end,
+  {ok, Epp} = epp_open(File, Fd, StartLine, IncludePath, PreDefMacros),
   _ = [io:setopts(Fd, [{encoding, Encoding}]) || Encoding =/= none],
   {ok, FileForm} = epp:parse_erl_form(Epp),
   OptModRes = epp:parse_erl_form(Epp),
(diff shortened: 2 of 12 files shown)


robertoaloi:
Merged #832 so you may want to rebase on top of it.

maxnordlund:
Hi, I noticed that the new changes also allows for passing [options to `erl_scan`](https://erlang.org/doc/man/erl_scan.html#type-option). One of which is `text`, which is stored inside the `erl_anno` field of the tokens, and subsequently the AST (although I suspect some may get lost in the process).

Anyway, this might be enough to calculate the end position afterwards. An alternative would be to tokenize ourselves, add an end location to each token, and then continue parsing. Not sure if it's any better though.

garazdawi:
> I noticed that the new changes also allows for passing options to erl_scan. One of which is text, which is stored inside the erl_anno field of the tokens, and subsequently the AST (although I suspect some may get lost in the process).

The PR does add that option, but IIRC we have removed it when testing and finalizing the PR as it was very incomplete and buggy. We are discussing adding something similar though to help with determining what text makes up an AST term.

robertoaloi:
@garazdawi What should we do with this one?

garazdawi:
Completely forgot about it! I'll find some time this week to look at it again.

Do you have any opinions about if I should target the poi? For very long attributes (eg. `-spec`), the highlighted part becomes very large.

gomoripeti:
I wonder how the quoted source code in compiler error messages for multi-line expressions will look like when Richard finishes his effort ongoing in https://github.com/erlang/otp/pull/3026.

I wouldn't mind highlighting the whole spec when the error message relates to the whole attribute but if there is a compiler error for eg an integer that does not have its own POI, the whole function clause is highlighted? Maybe some simple heuristic, like if the POI at point starts in a previous line (or even if it's multi-line) fall back to single line highlighting?

garazdawi:
I've pushed updates that makes columns numbers work for OTP-24. All tests pass for OTP-23, but some fail in OTP-24.

A problem that I noticed while doing this though is that in OTP-24 [we changed](https://github.com/erlang/otp/commit/9e1e41a99bc959ec22b1fe705ffb7385f54981a8#diff-ce50385a52bd00ba661c7cc3af59a473b66ae6c4820bc6d4cde79b9dc963bec9R373) where the anno of some things refers to. For most places it does not matter much, but the anno of `record` and `record_field` does matter as it is used by erlang_ls to figure out highlighting, rename, references, and such. In this PR the testcases that work with records currently break (on OTP-24) because of this and I don't know how to solve it in erlang_ls. I talked to @uabboli about this and I think we agreed that for OTP-24 RC2 we will revert the part of the change that modifies `record` and `record_field` to again take the `anno` from the `#` or possibly the record name.

Regarding the usage of pois as the mechanism to figure out what to highlight I think the current approach works pretty well. We may have to tweak it a bit going forwards, but this is a step in the right direction. If erlang/otp#4529 and erlang/otp#3026 make it into Erlang this will all become a lot easier to do.

I'm happy with this PR, so feel free to merge it if you agree.

garazdawi:
Rebased on latest main and fixed some issues that popped up along the way.

All tests pass with the current master of erlang.

OTP-24.0-rc2 is scheduled to be released next week and will contain the fixes to erl_parse that I referenced above.

garazdawi:
Updated the PR to also test with OTP 24 and fixed a couple of compatibility issues with running against both 21, 22, 23, and 24.

Let me know if you would like the OTP 24 fixes in a separate PR.

robertoaloi:
Amazing stuff, thanks a lot @garazdawi !

PR #834, opened 2020-12-18
Fix macro entry opaque ranges

garazdawi:
Fixes a bunch of problems when it comes to ranges, there are still more left. I'll try to get a PR for those as well.

### DIFF
--- a/priv/code_navigation/src/code_navigation_types.erl
+++ b/priv/code_navigation/src/code_navigation_types.erl
@@ -7,3 +7,5 @@
 -opaque opaque_type_a() :: atom().
 
 -export_type([ opaque_type_a/0 ]).
+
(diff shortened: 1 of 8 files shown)


robertoaloi:
Amazing, thanks! There are a couple of things we should do about POIs and ranges:

* Avoid the multiple representation of a "range" (right now there are three different ones in the code)
* Introduce the concept of view/edit range for POIs. These could be used, respectively, to trigger go to definition and to determine renaming (right now the relevant code is scattered across the codebase)
* Refactor the test suites. My idea would be to generate the LSIF (https://code.visualstudio.com/blogs/2019/02/19/lsif) from a sample code base and to compare the results found by the server with the information in the LSIF. That would avoid having to manually check line/column numbers in tests.

PR #837, opened 2020-12-22
Add tests for highlight of records, spec, behaviour and more

garazdawi:
Add test for various highlight ranges. Also fixed various bugs found when implementing the tests.

As you said in the other PR, we really need a better way to test this. Also, the way that the ranges is done is very fragile as you can just insert a space somewhere and it will break. i.e.

```erlang
fun 
                
lists

:

foldl

/

3.
```

The poi range for this incorrect, but the above is correct (though unusual) code. Not to mention if we start putting strange unicode characters in the code... Maybe I'll see what I can do about that after the holidays.

This PR at least is an improvement :)

### DIFF
--- a/elvis.config
+++ b/elvis.config
@@ -6,6 +6,7 @@
          , ruleset => erl_files
          , rules   => [ {elvis_style, god_modules, #{ignore => [ els_client
                                                                , els_definition_SUITE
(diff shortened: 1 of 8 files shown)


robertoaloi:
🙏

PR #845, opened 2020-12-30
Support jumping from record field to record definition

gomoripeti:
### Description

Add new poi for a field within a record expression. Same way as for record
access field, target range of "jump to definition" is the record name (instead of the field name) in
the record declaration

In the new `record_field` poi kind I chose to use field name as id and record name as data (is that right?).

Next step would be to jump to field within the record declaration. (Currently `record` poi does not contain range of fields. Would the right way be to extend `record` poi with field ranges or add a separate new `record_declaration_field` poi?)
(This is my first try at erlang_ls, so doing step by step)


robertoaloi:
Hi @gomoripeti, it's great you're contributing! I like the idea behind this. The main thing to keep in mind is that the _POIs_ are kept in memory, so we could check how much memory does it take to store the record fields for something like OTP + a medium sized project.
One possibility (something I considered, but didn't implement yet) is to only store a subset of all POIs for _dependencies_, but it could be confusing for the final user.

src/els_parser.erl
```diff
@@ -397,11 +394,29 @@ record_expr(Tree) ->
...
       []
   end.
 
+-spec record_expr_fields(atom(), tree()) -> [poi()].
+record_expr_fields(Record, RecordExprTree) ->
+  Fields = erl_syntax:record_expr_fields(RecordExprTree),
+  [begin
+     NameNode = erl_syntax:record_field_name(FieldNode),
+     NameAtom = record_field_name_atom(NameNode),
+     poi(erl_syntax:get_pos(NameNode), record_field, NameAtom, Record)
```
robertoaloi, line 409: To answer your question in the description, an alternative could be to have a composite `Id`, something like `{RecordName, FieldName}`, mostly for consistency with the other cases. The `Data` field is currently only used for storing the `pretty_printed` version of a spec.

gomoripeti:
thanks @robertoaloi it's good to know the memory usage needs to be kept in mind.

Checking ets memory for a random project (most of the mem is taken up by otp modules in els_dt_document table) adding additional `record_field`/`record_def_field` pois gives an additional ~10%.
However by descending into subtrees the parser also creates an `atom` poi for each record field. If I omitted creating those atoms (ie `record_field`/`record_def_field` poi replaces an `atom` poi with the exact same range) the mem increase becomes minor.
My question is what are all the uses of the `atom` poi and should the code be updated to replicate those behaviours (if some occurrences are replaced by `record_field` pois)?
- completion for standalone atoms would of course include record field names
- when cursor is on a standalone atom, should record fields be highlighted as well?
- when cursor is on a record field, should standalone atoms be highlighted as well?
- what other uses of atom poi?

Another question about record access: currently `#rec.foo` is a single `record_access` poi (plus an `atom` poi `foo`). What if this would be replaced by a `record_expr` poi for `#rec.` and a `record_field` poi for `foo`, to make things more consistent?

robertoaloi:
Hi @gomoripeti! I suspect that `atoms` are only there for completion purposes. Thinking about it, I believe we could save lot of memory by not storing pois for atoms (and actually variables), but calculating them on-demand when needed (for the current and included buffers). After all, there's very little point in knowing about atoms and variables across the whole project. In that case, it would make total sense, instead, to store record fields since they could be used in a renaming operation. What do you think?

Regarding your other questions:

> when cursor is on a standalone atom, should record fields be highlighted as well?
That's what I would probably expect as a user

> when cursor is on a record field, should standalone atoms be highlighted as well?
Good question. I would go for consistency with the previous point and say yes

robertoaloi:
@gomoripeti If you prefer, we could merge this PR as is and work on the atoms/variables POI elimination separately. What do you think?

gomoripeti:
hi @robertoaloi thanks for your answers. I have a few commits that address the atom+record_field highlighting that I will push shortly.
I would keep eliminating all atom and variable pois for another PR (but Im happy to look at that)

gomoripeti:
hi @robertoaloi I pushed all small changes I had in mind. If it looks good I can squash it (or remove unwanted commits)
one more thing that I would handle outside of this PR is indexing record_field references, as I have to check how much memory overhead it might have

include/erlang_ls.hrl
```diff
@@ -590,14 +590,16 @@
...
+                   | record_def_field
                    | record_expr
+                   | record_field
                    | spec
                    | type_application
                    | type_definition
                    | variable.
 -type poi_range() :: #{ from := pos(), to := pos() }.
 -type poi_id()    :: atom()
+                   | {atom(), atom()} %% record_def_field, record_field
```
robertoaloi, line 602: (outside of the PR scope) We should probably create aliases for these ids (eg `poi_id_record_field()`).

robertoaloi (approved):

robertoaloi:
Nice stuff, thanks!

include/erlang_ls.hrl
gomoripeti: good idea

PR #850, opened 2021-01-05
Provide elvis_path option

define-null:
Added option for configuring elvis config path

### DIFF
--- a/src/els_config.erl
+++ b/src/els_config.erl
@@ -48,7 +48,8 @@
                | plt_path
                | root_uri
                | search_paths
-               | code_reload.
+               | code_reload
+               | elvis_config_path.
 
 -type path()  :: file:filename().
 -type state() :: #{ apps_dirs        => [path()]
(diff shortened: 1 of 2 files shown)


src/els_config.erl
```diff
@@ -99,6 +100,7 @@ do_initialize(RootUri, Capabilities, {ConfigPath, Config}) ->
   CodeReload = maps:get("code_reload", Config, disabled),
   Runtime = maps:get("runtime", Config, #{}),
   CtRunTest = maps:get("ct-run-test", Config, #{}),
+  ElvisPath = maps:get("elvis_path", Config, undefined),
```
robertoaloi, line 103: This maybe interpreted as the path to the `elvis` executable/escript. Would `elvis_config_path` be a more precise name?

define-null: Good point, fixed.

robertoaloi:
Thanks!

PR #851, opened 2021-01-06
Implement code_path_extra_dirs config key

zsoci:
### Description

When code_path_extra_dirs key is set in erlang_ls.config, the list of wildcard directories will be added with code:add_path/1.
This solves the case when during compile, a parse_transform module wants to use another module that is impossible to track down by erlang_ls.


robertoaloi:
Thanks!

PR #852, opened 2021-01-06
Handle macros in patterns

### LINKED ISSUE #805: Go to definition fails with functions that have a parametrised macro in their head.
**Describe the bug**
As per the title.

**To Reproduce**
Given the snippet below, the `hello/1` function throws a `Definition not found`, whereas the `hi/1` function is navigated to as expected, and so is `howdy/1`.
```
-module(erlang_ls_goto).

-export([hello/1]).
-export([hi/1]).
-export([howdy/1]).

-define(HELLO(World), {hello, World}).
-define(HOWDY, howdy).

hello(?HELLO(World)) ->
    World.

hi({hi, World}) ->
    World.

howdy(?HOWDY) ->
    howdy.

```

**Expected behaviour**
Go to definition should navigate to a function with a parametrised macro in its head.

**Actual behaviour**
`Definition not found`

**Context**
 - `erlang_ls` version (tag/sha): 0.5.0
 - Editor used: `nvim`
 - LSP client used: `coc.nvim`

gomoripeti:
The dodger used to replace a macro with a local function call. That works in
expressions but not in patterns like function heads or catch patterns. Instead
of a function call a tuple syntax is used which works in more places.
(`?MACRO(Arg1, Arg2)` is temporarily replaced by `{'a-magic-atom', MACRO, Arg1, Arg2}`
 instead of old `'a-magic-atom'(MACRO, Arg1, Arg2)`)

This change is motivated by the commonly used `?EXCEPTION` macro

```
try
  ...
catch
  ?EXCEPTION(Class, Reason, StackToken) ->
    ...
end.
```

Open source examples:
https://github.com/devinus/poolboy/commit/5d40cc517edc9bb8ee70756544167b63f66662f0
https://github.com/eproxus/meck/commit/558e925b48ce257b12e381080c851dc49c87d7bb

I only noticed when looking at the tests that there is already a ticket which is addressed by this change:
Fixes #805 . 

I ran this modified `els_dodger` and `epp_dodger` on most of the OTP modules, and they either return the same result or `epp_dodger` returns some errors while `els_dodger` doesn't.

I'm willing to submit this change upstream to OTP, but erlang_ls has a better turnaround time and more added value to start here first :)

### DIFF
--- a/src/els_dodger.erl
+++ b/src/els_dodger.erl
@@ -791,16 +791,27 @@ macro(L, {Type, _, A}, Rest, As, Opt) ->
 macro_call([{'(', _}, {')', _}], L, {_, Ln, _} = N, Rest, As, Opt) ->
   {Open, Close} = parentheses(As),
   scan_macros_1([], Rest,
-                lists:reverse(Open ++ [{atom, L, ?macro_call},
-                                       {'(', L}, N, {')', Ln}] ++ Close,
+                %% {'?macro_call', N }
+                lists:reverse(Open ++ [{'{', L},
+                                       {atom, L, ?macro_call},
+                                       {',', L},
+                                       N,
+                                       {'}', Ln}] ++ Close,
                               As), Opt);
 macro_call([{'(', _} | Args], L, {_, Ln, _}=N, Rest, As, Opt) ->
   {Open, Close} = parentheses(As),
+  %% drop closing parenthesis
+  {')', _} = lists:last(Args), %% assert
+  Args1 = lists:droplast(Args),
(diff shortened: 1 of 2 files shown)


robertoaloi:
Nice one, thanks! @garazdawi FYI in case someone from the OTP team is looking into `epp_dodger` and especially https://bugs.erlang.org/browse/ERL-1429

onno-vos-dev:
@gomoripeti Just wanted to say THANK YOU! This has been a huge problem for me recently so thank you! <3

PR #853, opened 2021-01-07
Dont crash in case compiler option is not found in a beam

zsoci:
### Description
In case module was compiled with the 'deterministic' flag, the compile_info key does not have the 'option'  part so it is 'undefined'
Fixes # 
'undefined' is treated as a list and used with ++ that crashes els_compiler_diagnostics:load_dependencies(317).


robertoaloi:
Thanks!

PR #854, opened 2021-01-07
db_dir env var is not used any more

gomoripeti:
### Description

Just a minor cleanup in test. Additionally the [Internals](https://erlang-ls.github.io/internals/) page also talks about Mnesia. Maybe that page could be deleted or just mention the database is only stored in memory?


robertoaloi:
Nice catch! Yes, that documentation page is obsolete and should be removed.

PR #855, opened 2021-01-12
Implement using long names for distribution

zsoci:
### Description

Add use_long_names boolean variable to erlang_ls.config
Add cookie string variable to erlang_ls.config

This change will allow using long node names for the runtime node


robertoaloi:
Looks good, thanks!

PR #857, opened 2021-01-15
Fix macro rename deleting the next character

gomoripeti:
The range of a macro poi includes the leading '?'. However the columns were not
converted to protocol-range, so the values were shifted one char to the right
resulting in the '?' being preserved but the char after the macro being deleted.


robertoaloi:
Thanks!

PR #858, opened 2021-01-18
Update Makefile to build needed .plt for typer tests

alanz:
The `suggest_spec` test fails if the appropriate PLT does not exist.

Update the makefile so that `make ci` builds the PLT, so the tests can be run
locally and pass.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,7 @@ install: all
 	@ cp _build/default/bin/erlang_ls /usr/local/bin
 
 ci:
+	dialyzer --build_plt --apps erts kernel stdlib
 	rebar3 do compile, ct, proper --cover --constraint_tries 100, dialyzer, xref, cover, edoc
 
 coveralls:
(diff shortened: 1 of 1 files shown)


gomoripeti:
great!
however wouldn't this build the PLT over and over again each time `make ci` is called? (which takes 45 seconds for me)
what about something like
```
$HOME/.dialyzer_plt:
    dialyzer --build_plt --apps erts kernel stdlib

ci: $HOME/.dialyzer_plt
    rebar3 do ...
````

or
```
ci:
    dialyzer --check_plt || dialyzer --build_plt --apps erts kernel stdlib
    rebar3 do ...
```

robertoaloi:
I would also second @gomoripeti 's approach. As a note, the `make ci` is just used during development, since the GitHub pipeline uses rebar3 targets directly.

alanz:
>  `make ci` is just used during development

I know, but it should mimic what the github CI does, so developers can be confident that their pipelines will pass, and have a means to debug the problem in the code if not.

PR #863, opened 2021-01-21
code navigation: find bifs also if referenced without module name

### LINKED ISSUE #860: "fun atom_to_list/1" not found by crossref
**Describe the bug**

```erlang
    S = fun atom_to_list/1,
    S = fun erlang:atom_to_list/1,
```

The `crossref` diagnostics report a "not found" error for the first occurrence of `atom_to_list`, not the second.

**Context**
 - `erlang_ls` version (tag/sha): 90a68d8b40138ba06402c17699aa079e1593fcad
 - Editor used:
 - LSP client used:

al-khanji:
### Make `els_code_navigation` find non-fully qualified BIFs

Fixes #860  .

### DIFF
--- a/src/els_code_navigation.erl
+++ b/src/els_code_navigation.erl
@@ -31,11 +31,16 @@ goto_definition( _Uri
     {error, Error} -> {error, Error}
   end;
 goto_definition( Uri
-               , #{ kind := Kind, id := {F, A}}
+               , #{ kind := Kind, id := {F, A}} = POI
                ) when Kind =:= application;
                       Kind =:= implicit_fun;
                       Kind =:= export_entry ->
-  find(Uri, function, {F, A});
+  case erl_internal:bif(F, A) of
+    true ->
+      goto_definition(Uri, POI#{id := {erlang, F, A}});
+    false ->
+      find(Uri, function, {F, A})
+  end;
 goto_definition( _Uri
                , #{ kind := Kind, id := Module }
                ) when Kind =:= atom;


alanz (commented):
Looks good.

Can you add a test case to els_diagnostics_SUITE that shows it works?

gomoripeti:
thinking about `{no_auto_import,[{F,A}, ...]}` compiler option (http://erlang.org/doc/man/compile.html#file-2), maybe a heuristic that covers some edge cases could be what is also described in the documentation of the above compiler option
> As from R14A and forward, the compiler resolves calls without module prefix to local or imported functions before trying with auto-imported BIFs.

so in the below example I'd like to go to the local `error/1` function definition
```
f(A) ->
  error(A).

error(_) ->
  ok.
```

(I was also thinking about first looking for a `no_auto_import` option, but that can be in a header file, rebar.config or even command line option, so erlang_ls might not know about it)

michalmuskala:
Unfortunately, it's even more complex. There are 2 types of auto-imported BIFs, one can be shadowed without `no_auto_import`, the other requires the use of `no_auto_import` for shadowing.

For BIFs where `erl_internal:old_bif` returns true, shadowing is a hard error, while for the rest it's merely a warning.

al-khanji:
I will revise the diff. I'll at least make it so it first looks for a local function before trying to find a BIF. I'll see if I can also figure out how to check for `no_auto_import` and incorporate that.

@alanz I will also add the test, thanks for the reminder.

al-khanji:
This auto-closed when I renamed master -> main to track upstream setup, follow-up is in https://github.com/erlang-ls/erlang_ls/pull/876

PR #876, opened 2021-01-27
Make els_code_navigation find non-fully qualified BIFs

### LINKED ISSUE #860: "fun atom_to_list/1" not found by crossref

al-khanji:
### Make els_code_navigation find non-fully qualified BIFs

Add indexing of no_auto_import compile options. Use this information when deciding whether a fun reference is to a BIF.

Fixes #860.

### DIFF
--- a/apps/els_lsp/priv/code_navigation/src/diagnostics_autoimport.erl
+++ b/apps/els_lsp/priv/code_navigation/src/diagnostics_autoimport.erl
@@ -1,0 +1,6 @@
+-module(diagnostics_autoimport).
+
+-export([main/1]).
+
+main(_Args) ->
+    fun atom_to_list/1.
--- a/apps/els_lsp/priv/code_navigation/src/diagnostics_autoimport_disabled.erl
+++ b/apps/els_lsp/priv/code_navigation/src/diagnostics_autoimport_disabled.erl
@@ -1,0 +1,7 @@
+-module(diagnostics_autoimport_disabled).
+-compile(no_auto_import).
+
+-export([main/1]).
+
+main(_Args) ->
+    fun atom_to_list/1.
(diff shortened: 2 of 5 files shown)


al-khanji:
Need to add tests, hence a draft.

src/els_code_navigation.erl
```diff
@@ -74,6 +86,55 @@ goto_definition(_Uri, #{ kind := parse_transform, id := Module }) ->
...
+  case {OldBif, Bif} of
+    %% Cannot be shadowed, always imported
+    {true, true} ->
+      true;
+    %% It's not a BIF at all
+    {false, false} ->
+      false;
+    %% The hard case
+    {_, _} ->
+      is_bif_auto_import_disabled(Uri, F, A)
```
gomoripeti, line 102: my personal opinion (but I am not an erlang_ls maintainer) is that you don't need to check if auto_import is disabled, just jump to the bif at this point
- it can be disabled also in a header file, or rebar.config or erlc command line option of which erlang_ls might not even know
- auto_import is usually disabled to silence warnings about a bif clashing with a local function - but at this point no local function is found so it is highly likely there won't be a no_auto_import directive

what I mean is that this algorithm can not be 100% correct (because erlang_ls does not have all the info) so instead let's use a simple heuristic that already covers most of the practical cases (and also not 100% correct)

btw very nice thorough code covering all the edge cases

robertoaloi:
@al-khanji This PR is still marked as a _draft_. Let me know when you want me to have a look.

al-khanji:
> @al-khanji This PR is still marked as a _draft_. Let me know when you want me to have a look.

When you get a chance please take a look, I dropped POI parsing and just check whether we find a bif of any sort.

gomoripeti (approved):
very nice!

robertoaloi:
Thanks!

PR #877, opened 2021-01-29
Fix DAP startup and add function breakpoints

TheGeorge:
- start els_config and initialize it during DAP startup
- add handler and capability for setFunctionBreakpoints

### DIFF
--- a/src/els_dap_general_provider.erl
+++ b/src/els_dap_general_provider.erl
@@ -66,7 +66,7 @@ init() ->
 handle_request({<<"initialize">>, _Params}, State) ->
   %% quick fix to satisfy els_config initialization
   {ok, RootPath} = file:get_cwd(),
-  RootUri = unicode:characters_to_binary(["file://", RootPath]),
+  RootUri = els_uri:uri(els_utils:to_binary(RootPath)),
   InitOptions = #{},
   ok = els_config:initialize(RootUri, capabilities(), InitOptions),
   {capabilities(), State};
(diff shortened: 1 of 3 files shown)


src/els_dap_general_provider.erl
```diff
@@ -64,6 +64,11 @@ init() ->
 
 -spec handle_request(request(), state()) -> {result(), state()}.
 handle_request({<<"initialize">>, _Params}, State) ->
+  %% quick fix to satisfy els_config initialization
+  {ok, RootPath} = file:get_cwd(),
+  RootUri = unicode:characters_to_binary(["file://", RootPath]),
```
robertoaloi, line 69: You can use `els_uri:uri(els_utils:to_binary(Cwd))` to ensure compatibility with Windows.

src/els_dap_general_provider.erl
```diff
@@ -316,3 +351,25 @@ frame_by_id(FrameId, Threads) ->
 -spec format_mfa(module(), atom(), integer()) -> binary().
 format_mfa(M, F, A) ->
   els_utils:to_binary(io_lib:format("~p:~p/~p", [M, F, A])).
+
+-spec parse_mfa(string()) -> {module(), atom(), non_neg_integer()} | error.
+parse_mfa(MFABinary) ->
+  MFA = unicode:characters_to_list(MFABinary),
+  ?LOG_DEBUG("tokens for ~p: ~w", [MFA,  erl_scan:string(MFA)]),
```
robertoaloi, line 358: Nit: we could store the output of `erl_scan:string(MFA)` into a variable, instead of performing the operation twice.

TheGeorge: That was leftover debug print, let me remove these two lines

robertoaloi:
Thanks!

Issue #822, opened 2020-12-06
I would like M-. to go to variable definition - I think this is a feature request, not a bug


banjomet:
TLDR: M-. works for functions, but not for variables? At least for me.

**Is your feature request related to a problem? Please describe.**
When on a variable, hitting M-. should go and find the first definition of the variable within the function.

**Describe the solution you'd like**
It should look up the variable in the function.

**Describe alternatives you've considered**
Using another erlang/emacs library that can handle that case, but I am trying to keep it simple.

**Additional context**
``` erlang
%% @doc Defines a life cycle of a single agent
-spec loop(agent(), arenas(), sim_params(), config()) -> ok.
loop(Agent "M-. on 'Agent' below should make cursor jump up to here", Arenas, SP, Cf) ->
    case mas_misc_util:behaviour_proxy(Agent, SP, Cf) of
        migration ->
            loop(Agent "M-. here on 'Agent' should move cursor to Agent variable definition" , mas_conc_port:emigrate(dict:fetch(migration, Arenas), Agent), SP, Cf);
        Activity ->
            ArenaPid = dict:fetch(Activity, Arenas),
            case mas_conc_arena:call(ArenaPid, Agent) of
                close ->
                    ok;
                the_end ->
                    mas_concurrent:send_result(Agent);
                NewAgent ->
                    loop(NewAgent, Arenas, SP, Cf)
            end
    end.


```

banjomet:
Just to add, I just installed vscode to see if erlang-ls supports in in vscode, and it does. So this seems to be emacs specific. I am sorry for not detailing that, I made the assumption that people would know I am talking about emacs.

robertoaloi:
Hi @banjomet , _go to definition_ is currently not implemented, so thanks for this feature request. I wonder how it works in VS Code given it's not implemented... 🤔

banjomet:
@robertoaloi hello! I was wondering if I could work on implementing it then. Would that be okay?

robertoaloi:
@banjomet Absolutely! If you want to give it a try, you're more than welcome! Feel free to reach out if you need help!

alanz:
I had a discussion with @richcarl about this recently, and understand it can be done by using the output from [erl_syntax:annotate_bindings/1](http://erlang.org/doc/man/erl_syntax_lib.html#annotate_bindings-1).

PR #878, opened 2021-01-30
Add diagnostics highlighting already bound variables in patterns

gomoripeti:
Inspired by the infamous "pinning operator" and the great blog post of Roberto
on how to write a diagnostics backend. The backend is disabled by default
because of the mixed feedback the pinning-operator idea received.

### DIFF
--- a/priv/code_navigation/src/diagnostics_bound_var_in_pattern.erl
+++ b/priv/code_navigation/src/diagnostics_bound_var_in_pattern.erl
@@ -1,0 +1,20 @@
+-module(diagnostics_bound_var_in_pattern).
+
+-export([f/1, g/1, h/2]).
+
+f(Var1) ->
+  Var1 = 1.
+
+g(Var2) ->
+  case a:b() of
+    {ok, Var2} -> ok;
+    _ -> error
+  end.
+
+h(Var3, Var4) ->
+  try a:b() of
+    {New, Var3} ->
+      New
+  catch Var4 ->
+      error
+  end.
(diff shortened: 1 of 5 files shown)


robertoaloi:
Ahah, amazing idea! My main feedback would be that the most of the logic could be factored out, sot that it could be reused for implementing #822. In principle, we shouldn't index variables as part of the standard indexing procedure (i.e. for the whole project), but for the current buffer only.

Happy to merge it as is and to tackle eventual refactoring separately. Thanks!

robertoaloi:
Just tried this in both Emacs and VS Code. In Emacs, I like that the hint is not obtrusive. Same in VS Code, where the diagnostic only marks a portion of the variable name, though. Not sure if that's expected.

Screenshot 2021-01-30 at 09 50 28

richcarl:
That was quick, great work! Regardless of whether anyone wants the explicit pinning annotations, I don't se why anyone would oppose having this highlighting on by default.

dgud:
Agree with @richcarl here, great work and make it default even non configure-able.

robertoaloi:
@richcarl @dgud The "disabled by default" approach is what we tend to use for all new features. That's to allow contributors and early adopters that use `main` to try out the feature in their daily code bases. If a feature results stable enough and users enjoy it, than it gets enabled by default prior to a release.

gomoripeti:
thanks for the positive feedback :)
I don't know why VSCode only marks a portion, erlang_ls should send the whole length.
Reworking handling variables sounds good to me. It would be enough to parse only the current function or folding_range for things like jump-to-def or references.

PR #891, opened 2021-02-02
[dap] implement evaluate and stacktrace callbacks for dap

TheGeorge:
- implementation for evaluate callback:
  - get values when overing over variables
  - support expressions for watch and repl
- complete stack frames

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -66,7 +66,7 @@
            , {plt_apps, all_deps}
              %% Depending on the OTP version, erl_types (used by
              %% els_typer), is either part of hipe or dialyzer.
-           , {plt_extra_apps, [dialyzer, hipe, mnesia, common_test]}
+           , {plt_extra_apps, [dialyzer, hipe, mnesia, common_test, debugger]}
            ]}.
 
 {xref_checks, [ undefined_function_calls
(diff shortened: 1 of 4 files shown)


TheGeorge:
This is gonna be the last "low-hanging fruit" PR for supporting DAP features, I'll concentrate on tests after this one

TheGeorge:
Screen Shot 2021-02-02 at 6 15 02 PM

robertoaloi:
> This is gonna be the last "low-hanging fruit" PR for supporting DAP features, I'll concentrate on tests after this one

Sounds good! Given that DAP does not interfere with LSP I thought it was good to get all the plumbing in and then to start polishing things and making them production ready. Another thing it would be extremely good is an overview of how the interaction actually works, in VS Code and Emacs. From my experience, once we get that information out (we can do it as an "article" in https://erlang-ls.github.io/), users (including myself) start to play with a feature and come back with suggestions, feature requests and bug fixes. So, the sooner we get that, the better it is.

TheGeorge:
The current plumming of the DAP as special case of the LS is quite dirty. I'm trying to factor the DAP components into a separate app. There are also several corner cases that make the adapter crash, that need to be ironed out. But yeah, excited to get this moving forward

PR #895, opened 2021-02-03
refactor erlang_ls into umbrella project

TheGeorge:
This moves erlang_ls to apps/ and splits it into:
- els_framework - general components shared by dap and ls
- els_dap - debug adapter
- erlang_ls - language server

Tests are passing for me locally, and both erlang_ls and els_dap work when trying out manually with vscode.

This allows to remove several hacks for making the DAP work.

Furthermore, this changes the application that holds the configuration in the application environment from erlang_ls to els_framework.

There are still refactoring steps necessary, that splits erlang_ls and framework and dap clean, but that can be follow-up work:
- els_config is very erlang_ls specific, but is needed by dap for working with els_distribution_server
- the server module (either els_server or els_dap_server) is now set as part of the els_framework, this should be an argument to the init function in the associated supervisors
- more stuff I probably havent discovered yet


TheGeorge:
The upload docs workflow is broken

robertoaloi:
Thanks for starting this! I wonder if it would be clearer to rename the apps as:

* els_dap
* els_lsp
* els_core

apps/els_dap/erlang_ls.config
```diff
@@ -0,0 +1,2 @@
+include_dirs:
```
robertoaloi, line 1: Do we really need per-app configs?

apps/els_dap/src/els_dap.app.src
```diff
@@ -0,0 +1,16 @@
+{application, els_dap, [
+    {description, "Erlang Debug Adapter"},
```
robertoaloi, line 2: Nit: "Erlang LS - Debug Adapter Protocol"

apps/els_dap/src/els_dap.erl
```diff
@@ -14,18 +14,28 @@
...
+-define(APP, els_dap).
+-define(FRAMEWORK, els_framework).
 -define(DEFAULT_LOGGING_LEVEL, "debug").
 
 -spec main([any()]) -> ok.
 main(Args) ->
   application:load(getopt),
   application:load(?APP),
+  application:load(?FRAMEWORK),
+  Config1 = application:get_all_env(?FRAMEWORK),
```
robertoaloi, line 26: Leftovers from debugging? Or do you actually want to log these?

apps/els_dap/src/els_dap.erl
```diff
@@ -14,18 +14,28 @@
...
   ok = parse_args(Args),
+  Config2 = application:get_all_env(?FRAMEWORK),
+  application:set_env(?FRAMEWORK, server, els_dap_server),
   configure_logging(),
   application:ensure_all_started(?APP),
+  % application:ensure_all_started(?ELS),
```
robertoaloi, line 32: Leftover, I assume.

apps/els_dap/src/els_dap.erl
```diff
@@ -128,6 +137,7 @@ configure_logging() ->
                                              , "[", level, "] "
                                              , msg, "\n"
                                              ]
+                              %  , single_line => false
```
robertoaloi, line 140: Leftover?

apps/els_framework/src/els_framework.app.src
```diff
@@ -0,0 +1,31 @@
+{application, els_framework, [
+    {description, "Erlang Language Server Framework"},
+    {vsn, "0.1.0"},
+    {registered, []},
+    {applications, [
```
robertoaloi, line 5: Some of these apps seem specific to the LSP parts (eg `docsh`, `rebar3_format` or `tdiff`).

apps/els_framework/src/els_stdio.erl
```diff
@@ -19,19 +19,15 @@
...
 start_listener(Cb) ->
-  {ok, IoDevice} = application:get_env(erlang_ls, io_device),
+  {ok, IoDevice} = application:get_env(els_framework, io_device),
   {ok, proc_lib:spawn_link(?MODULE, init, [{Cb, IoDevice}])}.
 
 -spec init({function(), any()}) -> no_return().
 init({Cb, IoDevice}) ->
   ?LOG_INFO("Starting stdio server..."),
   ok = io:setopts(IoDevice, [binary]),
-  case application:get_env(erlang_ls, dap) of
```
robertoaloi, line 29: \o/

apps/erlang_ls/src/els_app.erl
```diff
@@ -22,12 +22,7 @@
 -spec start(normal, any()) -> {ok, pid()}.
 start(_StartType, _StartArgs) ->
   ok = application:set_env(elvis_core, no_output, true),
-  case application:get_env(erlang_ls, dap) of
```
robertoaloi, line 25: \o/

apps/erlang_ls/src/erlang_ls.app.src
```diff
@@ -6,25 +6,11 @@
...
-      , yamerl
-      , docsh
-      , elvis_core
-      , redbug
-      , ephemeral
-      , rebar3_format
-      , tdiff
-      , getopt
-      , uuid
+      , els_framework
```
robertoaloi, line 9: See my comment for the other `.app.src` file above.

apps/erlang_ls/test/els_test_utils.erl
```diff
@@ -55,6 +55,7 @@ init_per_suite(Config) ->
                            , ?TEST_APP]),
   RootUri = els_uri:uri(RootPath),
   application:load(erlang_ls),
+  application:load(els_framework),
```
robertoaloi, line 58: Do we still need to load `erlang_ls`?

rebar.config
```diff
@@ -33,12 +33,15 @@
 
 {escript_emu_args, "%%! -connect_all false\n" }.
 {escript_incl_extra, [{"erlang_ls/priv/snippets/*", "_build/default/lib/"}]}.
+{escript_main_app, erlang_ls}.
```
robertoaloi, line 36: Found this in the rebar3 docs and I was planning to do exactly the same. Good stuff!

robertoaloi:
> The upload docs workflow is broken, since the docs are now in `apps/erlang_ls/docs`, `apps/els_framework/docs`, and `apps/els_dap/docs`
> 
> Seems like we can just såecify multiple paths in `build.yml`

Yes, that should be enough.

TheGeorge:
> Thanks for starting this! I wonder if it would be clearer to rename the apps as:
> 
> * els_dap
> * els_lsp
> * els_core

Yeah, let's change it to that

apps/els_dap/erlang_ls.config
TheGeorge: The LS will not find includes otherwise, at least not for VSCode.

My hunch is, that the LS does not understand the Umbrella structure well.

apps/els_dap/src/els_dap.erl
TheGeorge: yeah, this needs cleanup

apps/els_dap/src/els_dap.erl
TheGeorge: Yes, single-line comments are too hard to parse for me when debugging :)

apps/els_framework/src/els_framework.app.src
TheGeorge: I just put all external deps here, this needs sorting for sure

apps/erlang_ls/test/els_test_utils.erl
TheGeorge: No, there is no application config in either els_dap or erlang_ls anymore

rebar.config
TheGeorge: This let's us also change erlang_ls app to els_lsp

TheGeorge:
@robertoaloi thanks for the quick review, I'll clean the PR up and change the names as you suggested.

apps/els_dap/erlang_ls.config
gomoripeti: I use a config something like this for rebar3 umbrella projects (some of these might be default for erlang_ls already)
```
apps_dirs:
  - "apps/*"
deps_dirs:
  - "_build/*/lib/*"
include_dirs:
 # for include
  - "include"
  - "apps/*/include"
  - "_build/*/lib/*/include"
# for include_lib
  - "apps"
  - "_build/*/lib/"
```

TheGeorge: There are only defaults set for the base-case of one application:
```
  Paths = [ els_utils:resolve_paths( [ [RootPath, Dir, "src"]
                                     , [RootPath, Dir, "test"]
                                     , [RootPath, Dir, "include"]
                                     ]
                                   , RootPath
                                   , Recursive
                                   )
```
And there is an lsp instance started for each project app in vscode (plus one for the umbrella app)

TheGeorge: alright, having a central config for the umbrella works with:
```
apps_dirs:
    - "apps/*"
deps_dirs:
  - "_build/default/lib/*"
include_dirs:
  - "apps"
  - "apps/*/include"
  - "_build/default/lib/"
  ```
  
  It's not entirely correct, as the difference between `-include_lib` and `-include` are not picked up by the LS

**Comment by TheGeorge** [comment]
So this should fix the last issues:
- both erlang_ls and dap are working for me and starting up.
- tests pass

I'm a bit at a loss regarding the uploading of the edocs artifacts though

**Comment by TheGeorge** [comment]
> > The upload docs workflow is broken, since the docs are now in `apps/erlang_ls/docs`, `apps/els_framework/docs`, and `apps/els_dap/docs`
> > Seems like we can just såecify multiple paths in `build.yml`
> 
> Yes, that should be enough.

Took way longer to fix this than I'm proud of, but these are features of v2 (we used v1) I updated the version we use

File: apps/els_core/src/els_core.app.src [hunk 12]
```diff
@@ -0,0 +1,25 @@
...
+    {registered, []},
+    {applications, [
+        kernel,
+        stdlib,
+        ranch,
+        jsx,
+        cowlib,
+        yamerl,
+        redbug,
+        ephemeral
```
robertoaloi, line 13: Nit: I believe `ephemeral` is also only used by LSP (in the refactoring provider).

erlang_ls.config
```diff
@@ -0,0 +1,10 @@
+apps_dirs:
```
robertoaloi, line 1: I would not add this file to version control. Many users (including myself) customize it according to their own setup. The need for setting paths should disappear once we get the BSP part finished. Maybe include it as a `erlang_ls.config.sample`?

robertoaloi: Nit: Also, indentation is off for the two lists.

robertoaloi (approved):

erlang_ls.config
robertoaloi: Fixed as a follow up.

apps/els_core/src/els_core.app.src
robertoaloi: Fixed as a follow up.

TheGeorge:
Ah I'd thought you'd squash when merging :)

PR #896, opened 2021-02-03
Support references from function_clause POI

gomoripeti:
Finding references when pointer is on the first clause works, because element at
position is a `function` POI. However when the pointer is on another clause of a
multi-clause function, element at position is a `function_clause` POI.

### DIFF
--- a/test/els_references_SUITE.erl
+++ b/test/els_references_SUITE.erl
@@ -14,6 +14,7 @@
 -export([ application_local/1
         , application_remote/1
         , function_definition/1
+        , function_multiple_clauses/1
         , fun_local/1
         , fun_remote/1
         , export_entry/1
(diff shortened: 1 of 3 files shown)


test/els_references_SUITE.erl
```diff
@@ -112,6 +113,18 @@ function_definition(Config) ->
   assert_locations(Locations, ExpectedLocations),
   ok.
 
+-spec function_multiple_clauses(config()) -> ok.
+function_multiple_clauses(Config) ->
+  Uri = ?config(hover_docs_uri, Config),
```
gomoripeti, line 118: sorry for shamelessly reusing test modules of another feature :)

robertoaloi:
Thanks, I wanted to do this for such a long time! Another thing that often happens to me is to try and find references while hovering the `specs`. How would you feel if that would also had the same behaviour?

gomoripeti:
finding references from spec sounds good!
Currently the spec poi range is the whole form (ie from the beginning of `-spec` to the terminating dot) The simple way would be to find-references from the whole thing. Another option is to parse the spec and only find refs when the point is on the function name. Which one do you think is more user friendly?

robertoaloi:
> Currently the spec poi range is the whole form (ie from the beginning of `-spec` to the terminating dot) The simple way would be to find-references from the whole thing. Another option is to parse the spec and only find refs when the point is on the function name. Which one do you think is more user friendly?

As a user, I would probably expect `find-references` only from the function name. Doing so from a `type poi` within the specs, for example, could give the false impression that we can `find-references` from a reference itself.

PR #901, opened 2021-02-07
Find POIs recursively in type, spec and record attributes

### LINKED ISSUE #815: Record definitions cannot be navigated from type specs
Given the following type specs:

```
-spec my_function(my_type(), #my_record{}) -> ok.
```

It is not possible to navigate to the definition of `#my_record{}`.

gomoripeti:
### Description

Use the generic fold and subtree mechanism to descend into nested structures and
find all types and record types.

### Bacground

I started out fixing jumping to type and record definitions from record definitions. Then a bit of scope creep happened, as I realised the body of type and spec attributes used to be parsed on a special shallow way only. Additionally I made preparations to find macros in all kinds of weird places in attributes.

Fixes #815 #818 even more

### DIFF
--- a/apps/els_lsp/src/els_parser.erl
+++ b/apps/els_lsp/src/els_parser.erl
@@ -65,6 +65,9 @@ parse_form(IoDevice, StartLocation, Parser, _Options) ->
     {eof, _EndLocation} = Eof -> Eof
   end.
 
+%% @Doc Find POIs in attributes additionally using tokens to add location info
+%% missing from the syntax tree. Other attributes which don't need tokens are
+%% processed in `attribute/1'.
 -spec find_attribute_pois(erl_syntax:syntaxTree(), [erl_scan:token()]) ->
    [poi()].
 find_attribute_pois(Tree, Tokens) ->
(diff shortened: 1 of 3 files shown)


apps/els_lsp/src/els_parser.erl
```diff
@@ -568,14 +569,84 @@ subtrees(Tree, record_field) ->
...
+       AttrName =:= include_lib ->
+  [];
+attribute_subtrees(AttrName, [Arg])
+  when AttrName =:= callback;
+       AttrName =:= spec ->
+  {_FA, DefinitionClauses} = erl_syntax:concrete(Arg),
+  [DefinitionClauses];
+attribute_subtrees(AttrName, [Arg])
+  when AttrName =:= type;
+       AttrName =:= opque ->
```
robertoaloi, line 639: Typo: opque

gomoripeti: thanks, I realize I havent added tests for opaque and callback, will do

robertoaloi:
This is amazing, thanks!

PR #902, opened 2021-02-08
[dap] add initial set of test for DAP

TheGeorge:
Almost all of the DAP specific logic is in `els_dap_general_provider`

This adds tests for
- command-line args parsing 
- init
- launching project node through mfa
- setting break points
- extracting variables


TheGeorge:
More to come, but let's keep this reviewable

TheGeorge:
- fixed missing atom_to_binary in pre-OTP 23

TheGeorge:
So for some reason it is not possible to make a connection to the remote node in CI.

TheGeorge:
Alright, we actually need to manually start `epmd` if we want to use distribution

TheGeorge:
Ok, final issue is starting epmd on windows, but https://bugs.erlang.org/browse/ERL-994 shows we can just do
`erl -sname a -noinput -eval "halt(0)."`

TheGeorge:
@robertoaloi some jobs seem to be stuck (finished, but didn't report as finished yet) is it possible to rerun them?

michalmuskala:
I've triggered a re-build of gh actions

TheGeorge:
> I've triggered a re-build of gh actions

How did you do this, if I can ask :-)

michalmuskala:
There's a button in top right corner, but it might be showing up only for people with commit bits to the repository

Screenshot 2021-02-09 at 14 41 40

TheGeorge:
@michalmuskala probably I don't have the rights to do this.

Hmmm still shaving "Waiting for status to be reported", I don't really understand what's missing here

TheGeorge:
Ok, digging a bit, it seems to be an issue when changing the builds. The branch protection rules expect certain builds, but they never get executed (see https://github.community/t/build-12-x-expected-waiting-for-status-to-be-reported/17576 for an example).

But here, I only added a step to the build rules, the names are still the same. I'd expect that to still work without needing to adjust the repo settings. 

Let me revert the changes to the build.yml and do the epmd start in Ct instead to see if that effects things.

TheGeorge:
Seems like the repo configuration is not correct and the branch protection rules don't match with jobs in build.yml

michalmuskala:
Yeah, I don't have access to those settings

robertoaloi:
> Seems like the repo configuration is not correct and the branch protection rules don't match with jobs in build.yml
> 
> @robertoaloi can you take a look?

Sure, sorry for the late reply. For some reason, one of the repo settings contained a duplicated setting for those two versions. Removed them and everything seems fine now.

apps/els_core/src/els_utils.erl
```diff
@@ -29,12 +29,12 @@
...
 %%==============================================================================
 
--spec cmd(string(), [string()]) -> integer().
+-spec cmd(string(), [string()]) -> integer() | no_return().
 cmd(Cmd, Args) ->
   cmd(Cmd, Args, []).
 
 % @doc Replacement for os:cmd that allows for spaces in args and paths
--spec cmd(string(), [string()], string()) -> integer().
+-spec cmd(string(), [string()], string()) -> integer() | no_return().
```
robertoaloi, line 37: Why is this needed?

apps/els_dap/test/els_dap_SUITE.erl
```diff
@@ -0,0 +1,118 @@
+-module(els_dap_SUITE).
+
+-include("els_dap.hrl").
+
+%% CT Callbacks
+-export([
```
robertoaloi, line 6: Nit: we use 2-space indentation in the project.

apps/els_dap/test/els_dap_test_utils.erl
```diff
@@ -0,0 +1,136 @@
+-module(els_dap_test_utils).
```
robertoaloi, line 1: It would be good to refactor these to avoid the duplication.

apps/els_dap/test/els_dap_general_provider_SUITE.erl
```diff
@@ -0,0 +1,408 @@
...
+-spec suite() -> [tuple()].
+suite() ->
+    [{timetrap, {seconds, 30}}].
+
+-spec all() -> [atom()].
+all() ->
+    els_dap_test_utils:all(?MODULE).
+
+-spec groups() -> [atom()].
+groups() ->
```
robertoaloi, line 51: Nit: not needed.

apps/els_dap/test/els_dap_general_provider_SUITE.erl
```diff
@@ -0,0 +1,408 @@
...
+init_per_suite(Config) ->
+    Config.
+
+-spec end_per_suite(config()) -> ok.
+end_per_suite(_Config) ->
+    ok.
+
+-spec init_per_testcase(atom(), config()) -> config().
+init_per_testcase(TestCase, Config) when
+    TestCase =:= undefined orelse
```
robertoaloi, line 64: We could use an auxiliary function here, instead of calling back the `init_per_testcase/2` with an `undefined` atom.

robertoaloi (approved):

PR #904, opened 2021-02-10
Account for quoted atoms in POI ranges

### LINKED ISSUE #847: Ranges/highlighting misaligned for quoted atoms
**Describe the bug**

The range for atoms is calculated based on `atom_to_list/1`, which misses the additional length of quotes.

**To Reproduce**

Ranges are used for multiple things, example based on what is highlighted.
`'foo!'` - only the first 4 chars (`'foo`) are highlighted
`'foo'` - only the first 3 chars (`'fo`) are highlighted

**Expected behavior**

Range/highlight should be from opening quote to closing quote

**Actual behavior**

Range/highlight is 2 chars shorter.

**Possible fix**

For atoms with special characters replacing `atom_to_list/1` could help, but for atoms which don't need quoting this might be more problematic as (iiuc) text representation of expressions is lost during parsing. (This later case might not be that common to address)

**Context**
 - `erlang_ls` version (tag/sha): master at e346973 
 - Editor used: emacs
 - LSP client used: emacs lsp-mode

keynslug:
### Description

Some atoms in the Erlang source code must be quoted in order to not confuse the parser:

> https://erlang.org/doc/reference_manual/data_types.html#atom

> An atom is to be enclosed in single quotes (') if it does not begin with a lower-case letter or if it contains other characters than alphanumeric characters, underscore (_), or @.

Use io_lib facilities instead of `erlang:atom_to_list/1` which respect that requirement.

Fixes #847.

### Downsides

* To make "source code representation" of an atom this PR employs `io_lib:write/1`, which I believe performs somewhat slower than original `erlang:atom_to_list/1`, though the effects should be neglidgible.

* This approach still won't handle needlessly quoted atoms like `'atom'`.


robertoaloi (approved):

robertoaloi:
Thanks! LGTM, @gomoripeti would you also like to have a look, since you reported the original issue?

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -177,9 +177,9 @@ atoms(Document, Prefix) ->
   Unique = lists:usort(Atoms),
   filter_by_prefix(Prefix, Unique, fun to_binary/1, fun item_kind_atom/1).
 
--spec item_kind_atom(binary()) -> map().
+-spec item_kind_atom(atom()) -> map().
```
gomoripeti, line 180: this last commit I don't fully understand
`filter_by_prefix(Prefix, Unique, fun to_binary/1, fun item_kind_atom/1).`
this converts the atom in the `id` of the POI with `to_binary` into a binary, then `item_kind_atom/1` is definitely called with a binary. It is the wrong binary though because it's missing the quotes, right?
so maybe `to_binary/1` could be fixed to call `io:write` instead? what do you think @keynslug ?

(otoh `item_kind_field` is called with a binary from `all_record_fields/2` but with an atom from `record_fields/2`, that's wring indeed, so maybe the later should be fixed to wrap it in a `to_binary` first)

gomoripeti (commented):
looks pretty good
I might be missing something about the last commit though

keynslug:
> I might be missing something about the last commit though

Yeah. I guess completions demand some more tests, in line with those in the highlight suite. I'm happy to hack on them this weekend, unless you consider this is going to be out of scope of the underlying issue.

apps/els_lsp/src/els_completion_provider.erl
keynslug: Thanks for a thoughtful review. Guess you're right, I've addressed these concerns in  dda2734 .

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -412,9 +406,11 @@ completion_item(#{kind := Kind, id := {F, A}}, true)
...
    };
-completion_item(#{kind := Kind, id := Name}, _)
-  when Kind =:= record;
-       Kind =:= define ->
+completion_item(#{kind := Kind = record, id := Name}, _) ->
+  #{ label            => atom_to_label(Name)
+   , kind             => completion_item_kind(Kind)
+   };
+completion_item(#{kind := Kind = define, id := Name}, _) ->
   #{ label            => atom_to_binary(Name, utf8)
```
robertoaloi, line 414: Shouldn't we also use `atom_to_label/1` here?

robertoaloi (approved):

apps/els_lsp/src/els_completion_provider.erl
keynslug: I think that would give you suggestions in the form of `?'MACRO'`, which isn't quite right. Initially I did what you're suggesting here, and that broke a couple of testcases.

robertoaloi: That makes sense!

robertoaloi:
Thanks!

PR #913, opened 2021-02-15
fix --version for escripts and have dap version consistent with erlan…

TheGeorge:
```
erlang_ls --version
```

is currently broken


robertoaloi:
Thanks!

PR #914, opened 2021-02-16
[dap] improve dap startup and shutdown

TheGeorge:
- shutdown dap and send events when project node goes down
- ensure, that we don't double monitor nodes
- shutdown project node when we get a disconnect request
- add config for cookie and startup timeout

### DIFF
--- a/apps/els_core/src/els_distribution_server.erl
+++ b/apps/els_core/src/els_distribution_server.erl
@@ -155,11 +155,11 @@ start(Node) ->
   wait_connect_and_monitor(Node),
   ok.
 
--spec wait_connect_and_monitor(atom()) -> ok.
+-spec wait_connect_and_monitor(atom()) ->  ok | error.
 wait_connect_and_monitor(Node) ->
   wait_connect_and_monitor(Node, ?WAIT_ATTEMPTS).
 
--spec wait_connect_and_monitor(Node :: atom(), Attempts :: pos_integer()) -> ok.
+-spec wait_connect_and_monitor(Node :: atom(), Attempts :: pos_integer()) ->  ok | error.
 wait_connect_and_monitor(Node, Attempts) ->
   wait_connect_and_monitor(Node, Attempts, Attempts).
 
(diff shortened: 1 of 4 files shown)


TheGeorge:
it made sense to merge #915 and this one into one PR.

TheGeorge:

apps/els_dap/src/els_dap_general_provider.erl
```diff
@@ -692,3 +715,27 @@ do_function_breaks(Node, Module, FBreaks, Breaks) ->
...
+
+-spec ensure_connected(node(), timeout()) -> ok.
+ensure_connected(Node, Timeout) ->
+  case is_node_connected(Node) of
+    true -> ok;
+    false ->
+      % connect and monitore project node
+      case els_distribution_server:wait_connect_and_monitor(Node, Timeout) of
+        ok -> inject_dap_agent(Node);
+        _ -> stop_debugger()
```
robertoaloi, line 727: This could be just `error`, right?

apps/els_core/src/els_distribution_server.erl
```diff
@@ -155,24 +155,32 @@ start(Node) ->
...
+wait_connect_and_monitor(Node, Attempts, MaxAttempts) ->
   timer:sleep(?WAIT_INTERVAL),
   case connect_and_monitor(Node) of
     ok ->
       ok;
     error ->
       ?LOG_WARNING( "Trying to connect to node ~p (~p/~p)"
-                  , [Node, ?WAIT_ATTEMPTS - Attempts + 1, ?WAIT_ATTEMPTS]),
-      wait_connect_and_monitor(Node, Attempts - 1)
+                  , [Node, MaxAttempts - Attempts + 1, MaxAttempts]),
```
robertoaloi, line 182: Out of curiosity, why do we need this change?

apps/els_dap/src/els_dap_general_provider.erl
robertoaloi, line 727: This could just be `error`, right?

robertoaloi (approved):

apps/els_core/src/els_distribution_server.erl
TheGeorge: I added a config parameter, that makes the connection timeout configurable. This change displays the connection attemps correctly (e.g. 12/30)

apps/els_dap/src/els_dap_general_provider.erl
TheGeorge: In this case we could crash, and let the supervisor handle the error behavior.

PR #915, opened 2021-02-17
[dap] implement shutdown procedure correctly

TheGeorge:
- shutdown dap and send events when project node goes down
- ensure, that we don't double monitor nodes
- shutdown project node when we get a disconnect request


TheGeorge:
merged with #914

PR #918, opened 2021-02-19
different symbol kind for exported functions

TheGeorge:
This adds a lense that shows if a function is exported:
```
exported (0 references)
```

```
local (12 references)
```


TheGeorge:
I'm not so sure if this is the right way to display this information. This makes the whole editor window more noisy, and showing  the number of references also increases the time it takes to display everything significantly for larger projects.

robertoaloi:
> I'm not so sure if this is the right way to display this information. This makes the whole editor window more noisy, and showing the number of references also increases the time it takes to display everything significantly for larger projects.

I was wondering the same. Maybe we can start by showing if a function is exported or not and drop the reference count?

TheGeorge:
I was wondering if there is not a better way, e.g. have exported functions
as interface symbols instead of functions. Then its up to the editor to
display these kinds differently.

robertoaloi:
> I was wondering if there is not a better way, e.g. have exported functions as interface symbols instead of functions. Then its up to the editor to display these kinds differently.

In Emacs, I have some theme magic to display the functions in a different colour, maybe that's the way to go.

TheGeorge:
> > I was wondering if there is not a better way, e.g. have exported functions as interface symbols instead of functions. Then its up to the editor to display these kinds differently.
> 
> In Emacs, I have some theme magic to display the functions in a different colour, maybe that's the way to go.

I didn't find anything in vsode, but that's really just a short comming of the editor setup. I'd say we use different symbol kinds, it's the way that makes most sense from the LSP perspective. The editor is responsible for using this information correctly, and we get some of this in vscode (and seems it's better supported in emacs) 

Let me try this and see if there is some theme setup that vscode has as well.

alanz:
FYI there is an LSP spec issue about this, which is currently on their backlog. See https://github.com/microsoft/language-server-protocol/issues/1046

TheGeorge:
I was experimenting with this, but yeah, it currently only has deprecated as field.

PR #920, opened 2021-02-19
Fix infinite loop when parsing macro in wild attribute name

### LINKED ISSUE #919: Memory leak? Or to much memory.
**Describe the bug**

Update to master today 2021-02-19, and now beam.smp takes **26GB** of memory according to
windows performance monitor, it crashes soon thereafter.

**To Reproduce**
emacs ~/src/otp/lib/observer/test/crashdump_viewer_SUITE.erl
(on windows with wsl, don't know if that matter)

erlang_ls.config used:

diagnostics:
  enabled:
    - bound_var_in_pattern

apps_dirs:
        - "lib/*"
        - "erts/preloaded/"

**Expected behavior**

I used the master branch from beginning of January and that worked fine,
used a lot of memory but not this much.. 

**Actual behavior**
Maybe using below 2GB of memory, so it can be used on 32b machine.

**Context**
 - `erlang_ls` version (tag/sha):   df50bd4b9bd3bf4596e
 - Editor used:  emacs-26.3
 - LSP client used: emacs default ?

gomoripeti:
### Description

The infinite loop was triggered during indexing OTP by [`syntax_tools_SUITE_data/syntax_tools_test.erl`](https://github.com/erlang/otp/blob/master/lib/syntax_tools/test/syntax_tools_SUITE_data/syntax_tools_test.erl#L17) containing
```
-?attrib([foo2/2]).
```

While running the indexing on OTP I also found that a macro in a spec name caused a crash preventing parsing the whole file.

This was revealed by another test file in syntax_tools (`syntax_tools_SUITE_data/type_specs.erl`) containing

```
-spec ?MODULE:b() -> integer() | t().
```

Fixes #919

### DIFF
--- a/apps/els_lsp/src/els_parser.erl
+++ b/apps/els_lsp/src/els_parser.erl
@@ -128,16 +128,26 @@ analyze_attribute(Tree) ->
       [FATree | _] = erl_syntax:tuple_elements(ArgTuple),
       Definition = [], %% ignore definition
       %% concrete will throw an error if `FATRee' contains any macro
-      {AttrName, {AttrName, {erl_syntax:concrete(FATree), Definition}}};
+      try erl_syntax:concrete(FATree) of
+        FA ->
+          {AttrName, {AttrName, {FA, Definition}}}
+      catch _:_ ->
+          throw(syntax_error)
+      end;
     AttrName when AttrName =:= opaque;
                   AttrName =:= type ->
       [ArgTuple] = erl_syntax:attribute_arguments(Tree),
       [TypeTree, _, ArgsListTree] = erl_syntax:tuple_elements(ArgTuple),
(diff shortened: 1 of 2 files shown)


apps/els_lsp/src/els_parser.erl
```diff
@@ -637,7 +647,7 @@ attribute_name_atom(Tree) ->
     atom ->
       erl_syntax:atom_value(NameNode);
     _ ->
-      Tree
+      NameNode
```
gomoripeti, line 650: This is the essential part of the fix :)

dgud:
Drop in to confirm that the fix works.

robertoaloi:
Amazing, thanks! I wonder if we should really index content of the `_data` directories.

PR #925, opened 2021-02-23
Handle unicode in function clauses when pretty printing

gomoripeti:
### Description

Each `function_clause` POI contains the pretty printed text of the function head
and guards. The pretty printing used to crash if any of those contained unicode
characters.

Found while running indexing through all of OTP. I am afraid this is just one spot, as source code is handled as list of bytes in many places, but Im not an expert in unicode and source code encodings and printable ranges and friends.


robertoaloi:
Thanks! Pinging @richcarl as expert in all of those things, in case he wants to have a look.

richcarl:
Looks right to me.

PR #926, opened 2021-02-23
Fix extracting function spec snippets

### LINKED ISSUE #922: No spec in hover docs since 0.10.0
**Describe the bug**
No type specs appear in hover docs even if function has specs. Important to note I use OTP 21 which doesn't have shell_docs. I dont know where else the signatures are used where they might be missing

**To Reproduce**
Hover over a function call, which has a spec defined

**Expected behavior**
Hover docs should show type spec (as well as function clauses and documentation)

**Actual behavior**
Only function clauses and documentation shown, but no spec

**Context**
 - `erlang_ls` version (tag/sha): 1fd8c94137b or 0.10.0
 - Editor used: emacs
 - LSP client used: emacs lsp-mode
 - Erlang/OTP: 21.3

**Some details**

`els_docs` uses specs as a fallback when shell_docs are not available (https://github.com/erlang-ls/erlang_ls/blob/main/apps/els_lsp/src/els_docs.erl#L64) Signarutes are taken from the data field of `spec` POIs, but since https://github.com/erlang-ls/erlang_ls/pull/909 the format of spec syntax Tree changed and `erl_prettypr` crashes on it. Hence the data in all specs is always empty.

I see few ideas to fix this:
- reimplement/copy/paste relevant part of `erl_prettypr`
- somehow reuse the formatting provider (although that one works on files, not text snippets)
- use the range info in specs to retrieve original spec text from source file (instead of reformatting the parsed syntax tree)

The last option sounds the easiest for me, but there can be other considerations.

gomoripeti:
The `spec` POI has no data value any more. Hopefully it was not used elsewhere.

Fixes #922

### DIFF
--- a/apps/els_core/src/els_text.erl
+++ b/apps/els_core/src/els_text.erl
@@ -6,6 +6,7 @@
 -export([ last_token/1
         , line/2
         , line/3
+        , range/3
         , tokens/1
         ]).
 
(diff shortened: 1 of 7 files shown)


robertoaloi:
Nice! Are they any other use cases for the `poi/4` function, which take the extra `Data`? If not, we should probably clean it up. Of course that's out of scope for this PR.

gomoripeti:
I used erlang_ls to find references of `poi/4` :) there are a few more left (function, function_clause, type_definition, record)

PR #927, opened 2021-02-24
[dap] add logpoints and fix watchlist

TheGeorge:
This implements debugger based logpoints and fixes the watchlist.

Logpoints don't stop but instead print a message on the console. In practice this is an additional field for the breakpoint.
This implementation
- breaks on the breakpoint
- evaluates the expression
- creates an output event for the console
- continues

Evaluation triggers the callback from the attached project and that was exactly what was messing with the watch list as well. (and this is fixed as a side-effect)

### DIFF
--- a/apps/els_dap/src/els_dap_breakpoints.erl
+++ b/apps/els_dap/src/els_dap_breakpoints.erl
@@ -1,0 +1,98 @@
+-module(els_dap_breakpoints).
+-export([
+    build_source_breakpoints/1,
+    get_function_breaks/2,
+    get_line_breaks/2,
+    do_line_breakpoints/4,
+    do_function_breaks/4,
+    type/3
+]).
+
+%%==============================================================================
+%% Includes
+%%==============================================================================
+-include_lib("kernel/include/logger.hrl").
+
+%%==============================================================================
+%% Types
+%%==============================================================================
+
+-type breakpoints() :: #{
+    module() => #{
+        line => #{
+            line() => line_breaks()
+        },
+        function => [function_break()]
+    }
+}.
+-type line() :: non_neg_integer().
+-type line_breaks() ::
+    regular
+    | {log, expression()}.
+-type expression() :: string().
+-type function_break() :: {atom(), non_neg_integer()}.
+
+-export_type([breakpoints/0]).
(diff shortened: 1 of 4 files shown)


TheGeorge:
I also moved some breakpoint handling code out into a separate module.

apps/els_dap/src/els_dap_general_provider.erl
```diff
@@ -739,3 +729,17 @@ stop_debugger() ->
...
   lists:member(Node, erlang:nodes(connected)).
+
+-spec safe_eval(node(), pid(), string(), update | no_update) -> term().
+safe_eval(ProjectNode, Debugged, Expression, Update) ->
+  {ok, Meta} = els_dap_rpc:get_meta(ProjectNode, Debugged),
+  Command = els_utils:to_list(Expression),
+  Return = els_dap_rpc:meta_eval(ProjectNode, Meta, Command),
+  case Update of
+    update -> ok;
+    no_update ->
```
TheGeorge, line 740: This is to prevent an endless update loop. Evaluating something in the context of the meta process triggers notifying the attached process.

For variable watching, we don't want to do this, because the client will trigger updating through evaluate request (which evaluate variables in the meta process and trigger an update, causing a loop).

This is the same for logging. Furthermore, we already resumed the process, and entering handle_info with int_cb is assumed to only happen on breakpoints.

TheGeorge:

apps/els_dap/src/els_dap_breakpoints.erl
```diff
@@ -0,0 +1,98 @@
+-module(els_dap_breakpoints).
```
robertoaloi, line 1: Can we adopt the same style as the rest of project? Comma-first, 2 spaces, etc.

TheGeorge:
Yes I can make a formatting commit.

Do we have a formatter or checker for this in place?

TheGeorge:
please squash when merging this

robertoaloi:
> Do we have a formatter or checker for this in place?

Unfortunately, not yet.

robertoaloi:
And I forgot to squash, just saw your message 😢

PR #930, opened 2021-03-01
Use els_utils:lookup_document for lookup

### LINKED ISSUE #928: Cannot find definition for functions defined in OTP
**Describe the bug**
I suspect a regression happened due to the introduction of https://github.com/erlang-ls/erlang_ls/pull/881 even though I have not been able to confirm it. Something happened between 0.9.0 and 0.10.0 at least as this issue is not present in 0.9.0. In many modules that I operate I get random warnings that Erlang LS fails to find the definition of functions that are defined in OTP. Due to this, hot code reload is broken and my workflow (in which I heavily depend on hot code reloading for my day to day coding) is broken. 

**To Reproduce**
I've failed to produce a small enough testcase which confirms the issue and it seems to happen occasionally but not every time. Jumping to the `ordsets` module and re-loading the module which previously complained about `ordsets:subtract/2` not being found fixes the issue which seems to point to a lack of POIs from OTP. 

```
Cannot find definition for function ordsets:subtract/2
```

**Expected behavior**
Expect Erlang LS to find the definition of ordsets:subtract/2. 

**Context**
 - `erlang_ls` version (tag/sha): 0.10.0
 - Editor used: Visual Code Studio 1.53.2
 - LSP client used: ? (Extension used)

**erlang_ls.config**
```
otp_path: "<otp_path>"
plt_path: "<plt_path>"
diagnostics:
  enabled:
    - compiler
    - crossref
    - dialyzer
    - elvis
    - bound_var_in_pattern
    - unused_includes
    - unused_macros
deps_dirs:
  - "lib/*"
include_dirs:
  - "lib/*/include"
  - "lib/*/src"
  - "lib"
macros:
  - name: APPLICATION
code_reload:
  node: kred@<hostname>
```

sgillis:
### Description

Use els_utils:lookup_document for document lookups to make sure that files that were not yet indexed are indexed on request. Related to https://github.com/erlang-ls/erlang_ls/pull/881

Fixes #928

### DIFF
--- a/apps/els_lsp/src/els_diagnostics_utils.erl
+++ b/apps/els_lsp/src/els_diagnostics_utils.erl
@@ -31,23 +31,27 @@ included_uris(Document) ->
 dependencies([], Acc, _AlreadyProcessed) ->
   Acc;
 dependencies([Uri|Uris], Acc, AlreadyProcessed) ->
-  {ok, Document} = els_utils:lookup_document(Uri),
-  Behaviours = els_dt_document:pois(Document, [behaviour]),
-  ParseTransforms = els_dt_document:pois(Document, [parse_transform]),
-  IncludedUris = included_uris(Document),
-  FilteredIncludedUris = exclude_already_processed( IncludedUris
-                                                  , AlreadyProcessed
-                                                  ),
-  PTUris = lists:usort(
-             lists:flatten(
-               [pt_deps(Id) || #{id := Id} <- ParseTransforms])),
-  FilteredPTUris = exclude_already_processed( PTUris
-                                            , AlreadyProcessed
-                                            ),
-  dependencies( Uris ++ FilteredIncludedUris ++ FilteredPTUris
-              , Acc ++ [Id || #{id := Id} <- Behaviours ++ ParseTransforms]
-                ++ [els_uri:module(FPTUri) || FPTUri <- FilteredPTUris]
-              , sets:add_element(Uri, AlreadyProcessed)).
+  case els_utils:lookup_document(Uri) of
+    {ok, Document} ->
+      Behaviours = els_dt_document:pois(Document, [behaviour]),
+      ParseTransforms = els_dt_document:pois(Document, [parse_transform]),
+      IncludedUris = included_uris(Document),
+      FilteredIncludedUris = exclude_already_processed( IncludedUris
+                                                      , AlreadyProcessed
+                                                      ),
+      PTUris = lists:usort(
+                 lists:flatten(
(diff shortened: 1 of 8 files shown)


gomoripeti (approved):
thanks a lot

apps/els_lsp/src/els_unused_includes_diagnostics.erl
```diff
@@ -87,26 +83,21 @@ expand_includes(Uri) ->
...
-                            , Graph
-                            , IncludedUris),
-      expand_includes( Uris ++ NonVisitedIncludedUris
-                     , NewGraph
-                     , sets:add_element(Uri, Visited));
-    {ok, []} ->
-      ?LOG_WARNING("Failed lookup while expanding includes [uri=~p]", [Uri]),
-      []
-  end.
+  {ok, Document} = els_utils:lookup_document(Uri),
```
robertoaloi, line 86: What would happen if the included path is incorrect or if it does not exist yet? Or, more likely, if the current project is not configured correctly, so the included file is not found? That could cause the provider to crash, right?

apps/els_lsp/src/els_diagnostics_utils.erl
```diff
@@ -31,28 +31,23 @@ included_uris(Document) ->
...
-                                                ),
-      dependencies( Uris ++ FilteredIncludedUris ++ FilteredPTUris
-                  , Acc ++ [Id || #{id := Id} <- Behaviours ++ ParseTransforms]
-                    ++ [els_uri:module(FPTUri) || FPTUri <- FilteredPTUris]
-                  , sets:add_element(Uri, AlreadyProcessed));
-    Error ->
-      ?LOG_INFO("Lookup failed [Error=~p]", [Error]),
-      []
-  end.
+  {ok, Document} = els_utils:lookup_document(Uri),
```
robertoaloi, line 34: Same here. If a project is misconfigured, some dependencies will not be found and a hard pattern match here would result in a less stable server. Maybe we want to modify the `els_utils:lookup_document/1` so that if it fails after indexing it returns an error and keep the case in the callers?

robertoaloi (changes requested):
Hi! Even if in principle this makes sense, I am a bit worried about what would happen in case the project is misconfigured (which can happen often among new users). This change would cause crashes, which would result in providers stopping and the server malfunctioning. See comments below for a possible solution.

sgillis:
@robertoaloi Makes sense, I'll change that.

robertoaloi:
Nice, thanks!

robertoaloi (approved):

PR #934, opened 2021-03-08
[#933] Fix support for snippets

### LINKED ISSUE #933: Snippets do not work any longer
It looks like snippets (as explained in https://erlang-ls.github.io/articles/snippets-are-here/) do not work any longer.

robertoaloi:
### Description

Fixes #933 .

### DIFF
--- a/apps/els_lsp/src/els_snippets_server.erl
+++ b/apps/els_lsp/src/els_snippets_server.erl
@@ -29,7 +29,6 @@
 %%==============================================================================
 -define(SERVER, ?MODULE).
 -define(TABLE, snippets).
--define(CUSTOM_DIR, ).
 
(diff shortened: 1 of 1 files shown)


onno-vos-dev (approved):

onno-vos-dev:
No need to do it in this PR but would be nice if a test can be added to avoid regression in the future. Not sure how tricky it would be?

robertoaloi:
> No need to do it in this PR but would be nice if a test can be added to avoid regression in the future. Not sure how tricky it would be?

Indeed. I looked at the current test and it just checks whether the snippets retrieval mechanism works, without looking at specific snippets. It was retrieving an empty list of snippets and it considered it fine. 🥇 

I will follow up on that.

PR #940, opened 2021-03-13
[#939] Do not depend on the cowlib dependency

### LINKED ISSUE #939: Do not depend on cowlib
Erlang LS only uses one single function from `cowlib`, to parse HTTP headers. This function is unfortunately not available in OTP, but it feels overkill to include the whole `cowlib` library - which has a significant size - just for this.

* Copy the parsing function from `cowlib`, which is allowed by its current license (ISC)
* Include full Copyright and [LICENCE](https://github.com/ninenines/cowlib/blob/master/LICENSE)
* Remove `cowlib` dependency

robertoaloi:
### Description

Fixes #939 .


onno-vos-dev (approved):
Assuming you're a copy paste professional

PR #943, opened 2021-03-13
[#921] Fix wrong config file in PropEr tests

### LINKED ISSUE #921: Proper tests execution filled with warning
When run as part of CI, the PropEr tests pass, but the output is filled with:

```
=WARNING REPORT==== 22-Feb-2021::14:32:41.867939 ===
Could not read config file: path="/tmp/erlang_ls.config" class=throw error={yamerl_exception,
                                                                            [{yamerl_parsing_error,
                                                                              error,
                                                                              "Failed to open file \"/tmp/erlang_ls.config\": no such file or directory",
                                                                              undefined,
                                                                              undefined,
                                                                              file_open_failure,
                                                                              undefined,
                                                                              [{error,
                                                                                enoent}]}]}
```

I suspect the following commit introduced this behaviour: b0ca70fd

pablocostass:
In b0ca70f the config file that is used in the PropEr tests was renamed, causing the warning logs described in the issue. This commit renames the file back to `erlang_ls.config`.

Partially fixes #921 (as I locally now get crash reports from Erlang LS, but the property still holds).

### DIFF
--- a/apps/els_lsp/test/prop_statem.erl
+++ b/apps/els_lsp/test/prop_statem.erl
@@ -353,7 +353,7 @@ setup() ->
   HaltFun = fun(_X) -> Self ! halt_called, ok end,
   meck:expect(els_utils, halt, HaltFun),
   application:ensure_all_started(els_lsp),
-  file:write_file("/tmp/els_lsp.config", <<"">>),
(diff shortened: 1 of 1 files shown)


robertoaloi:
Thanks!

PR #945, opened 2021-03-16
[#899] Self-describing specs.

### LINKED ISSUE #899: Present the pre-calculated spec itself instead of "add spec" when a spec is missing

robertoaloi:
Instead of showing an "Add spec" button for missing spec, preview the
actual calculated spec by TypEr. Also:

* Simplify the `els_code_lens` behaviour
* Add an optional `init` function for the lenses
* Relax supervision strategy

### Description

Fixes #899 .

### DIFF
--- a/apps/els_lsp/src/els_code_lens.erl
+++ b/apps/els_lsp/src/els_code_lens.erl
@@ -8,12 +8,13 @@
 %% Callback Functions
 %%==============================================================================
 
--callback command(poi()) -> els_command:command_id().
--callback command_args(els_dt_document:item(), poi()) -> [any()].
+-callback init(els_dt_document:item()) -> state().
+-callback command(els_dt_document:item(), poi(), state()) ->
+  els_command:command().
 -callback is_default() -> boolean().
 -callback pois(els_dt_document:item()) -> [poi()].
 -callback precondition(els_dt_document:item()) -> boolean().
--callback title(poi()) -> binary().
+-optional_callbacks([init/1]).
 
 %%==============================================================================
 %% API
@@ -41,8 +42,10 @@
                  , data    => any()
                  }.
 -type lens_id() :: binary().
+-type state() :: any().
 -export_type([ lens/0
              , lens_id/0
+             , state/0
              ]).
 
 %%==============================================================================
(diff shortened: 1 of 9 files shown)


apps/els_lsp/src/els_code_lens.erl
```diff
@@ -8,12 +8,13 @@
 %% Callback Functions
 %%==============================================================================
 
--callback command(poi()) -> els_command:command_id().
--callback command_args(els_dt_document:item(), poi()) -> [any()].
+-callback init(els_dt_document:item()) -> state().
```
robertoaloi, line 11: This new, optional callback gives us a chance to run `TypEr` only once per file.

apps/els_lsp/src/els_code_lens.erl
```diff
@@ -8,12 +8,13 @@
...
+-callback command(els_dt_document:item(), poi(), state()) ->
```
robertoaloi, line 12: There was very little gain in having two callback functions for a very couple functionality. Use a combined one, instead.

apps/els_lsp/src/els_code_lens.erl
```diff
@@ -83,13 +93,10 @@ lenses(Id, Document) ->
...
--spec make_lens(atom(), els_dt_document:item(), poi()) -> lens().
-make_lens(CbModule, Document, #{range := Range} = POI) ->
-  Command = els_command:make_command( CbModule:title(POI)
-                                    , CbModule:command(POI)
-                                    , CbModule:command_args(Document, POI)),
+-spec make_lens(atom(), els_dt_document:item(), poi(), state()) -> lens().
+make_lens(CbModule, Document, #{range := Range} = POI, State) ->
   #{ range   => els_protocol:range(Range)
-   , command => Command
+   , command => CbModule:command(Document, POI, State)
```
robertoaloi, line 99: One callback instead of three. The price to pay is to invoke the `make_command` from the function, which I think makes sense. \o/

apps/els_lsp/src/els_providers_sup.erl
```diff
@@ -45,8 +45,8 @@ start_link() ->
 -spec init([]) -> {ok, {supervisor:sup_flags(), [supervisor:child_spec()]}}.
 init([]) ->
   SupFlags = #{ strategy  => one_for_one
-              , intensity => 1
-              , period    => 5
+              , intensity => 5
```
robertoaloi, line 48: Relaxed a bit to avoid the whole server to stop if a provider is occasionally misbehaving.

Issue #948, opened 2021-03-17
It seems like "-compile(export_all)." is making problem to the erlang_ls parser


pickme467:
During an indexing process the server crashes with the following error:
```2021-03-17 09:26:58.210 [error] <0.7483.0>@els_indexing:try_index_file:129 Error indexing file [filename=/utils/src/helpers.erl, uri=file:///utils/src/helpers.erl] error:function_clause:[{els_parser,find_compile_options_pois,[export_all,[{'-',{16,1}},{atom,{16,2},compile},{'(',{16,9}},{atom,{16,10},export_all},{')',{16,20}},{dot,{16,21}}]],[{file,"/github/erlang_ls/src/els_parser.erl"},{line,117}]},{els_parser,parse_form,4,[{file,"/github/erlang_ls/src/els_parser.erl"},{line,55}]},{els_dodger,parse,5,[{file,"/github/erlang_ls/src/els_dodger.erl"},{line,343}]},{els_parser,parse_file,1,[{file,"/github/erlang_ls/src/els_parser.erl"},{line,28}]},{els_dt_document,new,4,[{file,"/github/erlang_ls/src/els_dt_document.erl"},{line,138}]},{els_indexing,index,3,[{file,"/github/erlang_ls/src/els_indexing.erl"},{line,62}]},{els_indexing,try_index_file,2,[{file,"/github/erlang_ls/src/els_indexing.erl"},{line,127}]},{els_indexing,index_file,1,[{file,"/github/erlang_ls/src/els_indexing.erl"},{line,52}]}]```

I have made a fix that is on my branch and I can just make a PR for this, but I'm not sure if my fix would follow the whole parsing idea. The fix is the following:
```
diff --git a/src/els_parser.erl b/src/els_parser.erl
index 6e87f8d..715fbd2 100644
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -112,9 +112,11 @@ find_attribute_pois(Tree, Tokens) ->
       []
   end.
 
--spec find_compile_options_pois([any()] | tuple(), [erl_scan:token()]) ->
-        [poi()].
-find_compile_options_pois(CompileOpts, Tokens) when is_tuple(CompileOpts) ->
+-spec find_compile_options_pois([any()] | tuple() | atom(),
+                                [erl_scan:token()]) -> [poi()].
+find_compile_options_pois(CompileOpts, Tokens)
+   when is_tuple(CompileOpts);
[...]

PR #949, opened 2021-03-17
Made els_parser accept -compile(export_all).

pickme467:
This is a fix to the issue:
https://github.com/erlang-ls/erlang_ls/issues/948

### DIFF
--- a/src/els_parser.erl
+++ b/src/els_parser.erl
@@ -112,9 +112,11 @@ find_attribute_pois(Tree, Tokens) ->
       []
   end.
 
--spec find_compile_options_pois([any()] | tuple(), [erl_scan:token()]) ->
-        [poi()].
-find_compile_options_pois(CompileOpts, Tokens) when is_tuple(CompileOpts) ->
+-spec find_compile_options_pois([any()] | tuple() | atom(),
+                                [erl_scan:token()]) -> [poi()].
+find_compile_options_pois(CompileOpts, Tokens)
+   when is_tuple(CompileOpts);
+        is_atom(CompileOpts) ->
   find_compile_options_pois([CompileOpts], Tokens);
 find_compile_options_pois(CompileOpts, Tokens) when is_list(CompileOpts) ->
   Fun = fun({parse_transform, PT}, Acc) ->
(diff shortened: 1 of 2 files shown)


gomoripeti (approved):
lgtm, maybe could you add a one-liner test case in els_parser_SUITE something like this https://github.com/erlang-ls/erlang_ls/blob/main/apps/els_lsp/test/els_parser_SUITE.erl#L71

pickme467:
Test added.

robertoaloi:
@pickme467 Should you rebase on top of the latest `main` and solve the conflict, so we can merge this one?

pickme467:
Hi Roberto,

With your request I realized that the erlang_ls repository doesn't use
master branch, but main branch instead. The error I found is probably a
very old one existing on the master branch. I ran my test on the main
branch and it passed even without my fix, so I guess the main branch is
free from this problem. I use erlang_ls compiled from scratch, so probably
doing the clone it defaulted to master branch and that's why I experienced
problems my fix was addressing. Now I recompiled from the main branch for
me to use it.

Greetings,

Pawel

pt., 26 mar 2021 o 17:04 Roberto Aloi ***@***.***> napisał(a):

robertoaloi:
All right, that makes sense. Thanks for the update!

pickme467:
I have one small proposal. Maybe you could just merge releases to  master
branch, so that clone would give a valuable version out of the box? I never
thought that master branch could be just unusable and that's the reason I
haven't checked if there is another branch used for stable versions.

Greetings,

Pawel

sob., 27 mar 2021 o 10:18 Roberto Aloi ***@***.***>
napisał(a):

robertoaloi:
Hi @pickme467!

Renaming `master` into `main` is an intended choice.
`main` is the default branch name for repositories on GitHub since October 2020:

https://github.blog/changelog/2020-10-01-the-default-branch-for-newly-created-repositories-is-now-main/

We renamed the default branch of Erlang LS to `main` since we agree that is not necessary to stick to a less inclusive branch name as `master`. Renaming your local branch should be straightforward:

https://www.git-tower.com/learn/git/faq/git-rename-master-to-main/

PR #954, opened 2021-03-22
[#822] Add goto variable definition

### LINKED ISSUE #822: I would like M-. to go to variable definition - I think this is a feature request, not a bug

plux:
### Description

Add navigation support for variable definitions.

This will naively try to find the definition of a variable by finding the
first occurrence of the variable in the function clause.

Fixes #822.


apps/els_lsp/src/els_code_navigation.erl
```diff
@@ -22,6 +22,30 @@
...
+    [] -> {error, not_in_function_clause};
+    FunRanges ->
+      FunRange = lists:last(FunRanges),
+      %% Find the first occurance of the variable in the function clause
+      [POI|_] = [P || P = #{range := Range, id := Id} <- VarPOIs,
+                      els_range:compare(Range, VarRange),
+                      els_range:compare(FunRange, Range),
+                      Id =:= VarId],
+      case POI of
+        Var -> {error, already_at_definition};
```
robertoaloi, line 45: Nit: I would probably return "self" rather than returning an error at this stage.

robertoaloi:
Nice, thanks!

robertoaloi:
@plux This won't find definitions in macros, but I guess that can be addressed as a follow-up improvement.

PR #955, opened 2021-03-22
Limit export completions to unexported definitions

plux:
### Description

Limit completions in `-export()` and `-export_type()` to unexported definitions


robertoaloi:
Thanks!

PR #962, opened 2021-03-30
Add support for renaming functions

plux:
### Description
Add support for renaming functions

### DIFF
--- a/apps/els_lsp/priv/code_navigation/src/rename_function.erl
+++ b/apps/els_lsp/priv/code_navigation/src/rename_function.erl
@@ -1,0 +1,35 @@
+-module(rename_function).
+-export([foo/1, 'Quoted_Atom'/1]).
+
+-spec foo(any()) -> any().
+foo(1) ->
+  1;
+foo(2) ->
+  2;
+foo(_) ->
+  anything.
+
+bar() ->
+  ok = rename_function:foo(1),
+  F = fun foo/1,
+  ok = F(),
+  F2 = fun rename_function:foo/1,
+  ok = F2(),
+  F3 = fun foo/1,
+  foo(F3).
+
+-spec 'Quoted_Atom'(any()) -> any(). %% TODO: Replace Quoted_Atom -> Quoted.Atom
+'Quoted_Atom'(1) ->
+  1;
+'Quoted_Atom'(2) ->
+  2;
+'Quoted_Atom'(_) ->
+  anything.
+
+baz() ->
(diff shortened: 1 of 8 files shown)


gomoripeti:
good point in https://github.com/erlang-ls/erlang_ls/pull/962/files#diff-8b95bce4d2827f81806bbfd9649c6aa71564e36f0b693ed981f0b66e37412e9fR222
I think Roberto also mentioned that there needs to be multiple ranges assossiated with a POI.
Although bit arbitrary maybe additional ranges could be stored in the data part, eg an application poi "mod:fun(...)" could have `#{mod_range => _, fun_range => _, colon_range => _}`

robertoaloi:
> good point in https://github.com/erlang-ls/erlang_ls/pull/962/files#diff-8b95bce4d2827f81806bbfd9649c6aa71564e36f0b693ed981f0b66e37412e9fR222
> I think Roberto also mentioned that there needs to be multiple ranges assossiated with a POI.
> Although bit arbitrary maybe additional ranges could be stored in the data part, eg an application poi "mod:fun(...)" could have `#{mod_range => _, fun_range => _, colon_range => _}`

But I guess we can merge this one and follow up, or?

robertoaloi (approved):

robertoaloi:
Your 2 PRs were conflicting, would you mind rebasing this on `main`?

plux:
> > good point in https://github.com/erlang-ls/erlang_ls/pull/962/files#diff-8b95bce4d2827f81806bbfd9649c6aa71564e36f0b693ed981f0b66e37412e9fR222
> > I think Roberto also mentioned that there needs to be multiple ranges assossiated with a POI.
> > Although bit arbitrary maybe additional ranges could be stored in the data part, eg an application poi "mod:fun(...)" could have `#{mod_range => _, fun_range => _, colon_range => _}`
> 
> But I guess we can merge this one and follow up, or?

Yeah I think this could be merged and then we can improve upon it.

plux:
> Your 2 PRs were conflicting, would you mind rebasing this on `main`?

Rebase done!

robertoaloi:
Thanks! You're on a roll!

robertoaloi:
One thing it would be nice (and hopefully easy) to add is the ability to rename a function starting from any of each function clauses (right now it only works from the top-level clause).

plux:
> One thing it would be nice (and hopefully easy) to add is the ability to rename a function starting from any of each function clauses (right now it only works from the top-level clause).

Good point! I will look into fixing that.

PR #968, opened 2021-04-05
Add support for renaming function from function clause

plux:
### Description

Add support for renaming function from function clause


robertoaloi (approved):

robertoaloi:
Nice, thanks for following up!

PR #969, opened 2021-04-05
Add completions for attributes

plux:
### Description

Add completion for most module attributes.

Also add nice context aware completions for
```
-behaviour(<tab>
-include("<tab>
-include_lib("<tab>
```
Try it out!

### DIFF
--- a/apps/els_lsp/src/els_completion_provider.erl
+++ b/apps/els_lsp/src/els_completion_provider.erl
@@ -3,6 +3,7 @@
 -behaviour(els_provider).
 
 -include("els_lsp.hrl").
+-include_lib("kernel/include/logger.hrl").
 
 -export([ handle_request/2
         , is_enabled/0
@@ -18,6 +19,9 @@
                     , line     := line()
                     , column   := column()
                     }.
+
+-type items() :: [item()].
+-type item() :: completion_item().
 -type state() :: any().
 
 %%==============================================================================
@@ -81,7 +85,7 @@ resolve(#{ <<"kind">> := ?COMPLETION_ITEM_KIND_FUNCTION
 resolve(CompletionItem) ->
   CompletionItem.
 
--spec find_completions(binary(), integer(), options()) -> [any()].
+-spec find_completions(binary(), integer(), options()) -> items().
 find_completions( Prefix
                , ?COMPLETION_TRIGGER_KIND_CHARACTER
                , #{ trigger  := <<":">>
@@ -104,6 +108,11 @@ find_completions( _Prefix
                , #{trigger := <<"?">>, document := Document}
                ) ->
   definitions(Document, define);
+find_completions( _Prefix
+                , ?COMPLETION_TRIGGER_KIND_CHARACTER
+                , #{trigger := <<"-">>, document := _Document, column := 1}
+                ) ->
+  attributes();
 find_completions( _Prefix
                , ?COMPLETION_TRIGGER_KIND_CHARACTER
                , #{trigger := <<"#">>, document := Document}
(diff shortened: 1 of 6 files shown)


apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -183,12 +209,188 @@ find_completions( Prefix
...
+  #{ label            => Path
+   , kind             => ?COMPLETION_ITEM_KIND_FILE
+   , insertTextFormat => ?INSERT_TEXT_FORMAT_PLAIN_TEXT
+   }.
+
+%%==============================================================================
+%% Snippets
+%%==============================================================================
+-spec snippet(atom()) -> item().
+snippet(attribute_behaviour) ->
```
robertoaloi, line 341: Should we use the snippet server for this? (can be done as a follow up). Look at `els_snippets_server` and `apps/els_lsp/priv/snippets/`

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -183,12 +209,188 @@ find_completions( Prefix
...
+  , snippet(attribute_record)
+  , snippet(attribute_type)
+  , snippet(attribute_vsn)
+  ].
+
+%%=============================================================================
+%% Include paths
+%%=============================================================================
+-spec paths_include(els_dt_document:item()) -> [binary()].
+paths_include(#{uri := Uri}) ->
```
robertoaloi, line 259: is this supposed to provide completion for paths in include/include_lib? That would be amazing, but I don't seem to get it to work, though.

robertoaloi:
Love it. The only part which does not seem to work is completion for paths (not sure why). I tried raising the log level to none, but I could not see any error...

apps/els_lsp/src/els_completion_provider.erl
plux: Yeah, that's the idea atleast :) Good catch!
It needs to be triggered right after `-include("`.
I've only tried it out on a limited number of projects on my own computer, so there could definitely be issues with the code on other platforms or projects.
When you say it doesn't work, does it hang or does it simply not find any completions?
Does it not work on the erlang_ls repo? 
May I ask to see your erlang_ls.config, so I can try to reproduce it?

apps/els_lsp/src/els_completion_provider.erl
plux: I guess so, I will look into it.
I'm not too familiar with the `els_snippets_server`, what would the advantage of using that be?

robertoaloi: Mainly to have all snippets in one place and not hard-coded in code. Snippets are loaded from a directory, which means they can be overwritten by the user (e.g. if you prefer a slightly different formatting than the default ones).

apps/els_lsp/src/els_completion_provider.erl
robertoaloi: It does not find any completions for me. Tried on the `erlang_ls` code base and the Klarna one. But since I still like the idea and it does not cause any harm, I'm keen to merge this in and to eventually follow up with a fix before the next release. I will also try to clean and restart everything on my side to see if it was an artifact of my setup.

robertoaloi: @plux Issue confirmed. Redbugging the code, I don't seem to be able to trigger the completions.
Steps I try:
* Type `-inc`
* Select `-include_lib` from the dropdown
* Wait -> Nothing
* Start typing the name of an existing app -> Nothing
* Move the first `"` and wait -> Nothing
* Try to retype `"` -> Nothing
* Trigger a manual `completion-at-point` or `company-complete` -> Nothing

According to Redbug, the completion is not triggered at all. What should trigger it? Maybe we're missing a `"` trigger character?

Tried in both Emacs and VS Code. Which editor are you using?

plux: That seems odd, needs to be investigated further. :)
Currently I didn't add `"` as a trigger character, so only manually triggering the completion is intended to work.
Adding `"` as a trigger character sounds like a good idea, I will try it out.
 
I'm using Emacs with `lsp-mode` and `company`.

apps/els_lsp/src/els_completion_provider.erl
plux: I will look into it!

PR #970, opened 2021-04-05
Supply completions for POIs in includes recursively

plux:
### Description

Supply completions for POIs in includes recursively


robertoaloi:
Thanks!

PR #972, opened 2021-04-08
Global parse_transforms through configuration

zwilias:
### Description

In some of our work projects, we declare `parse_transform`'s to be "globally" activated through `erl_opts` in `rebar.config`. While we're aware that this is somewhat tricky since that means they're also used to compile our dependencies and the usage is implicit, we have so far preferred this approach over adding a `-compile` attribute to every file we use this in, or adding some sort of global `hrl` file and including that in every module.

The parse transform we use happens to make invalid syntax valid, so our modules won't compile without the parse transform being applied.

_We could potentially go the route of adding a `-compile({parse_transform, cut}).` attribute to every module. However, that's a lot of impact, with the main pro being that our devs can then use `erlang-ls`._

An alternative solution would be to have `erlang-ls` support configuring such parse_transforms in the configuration file; much in the same way that it supports macro-declarations.

This PR adds the minimal code I could figure out to make this _work_.

It's missing any kind of test or documentation.

My goal here is mostly to figure out whether there's any interest in allowing a `parse_transforms` key in `erlang_ls.config`, and if so, whether the direction this implementation takes is good to explore further. If so, I can look into tests and docs.

### DIFF
--- a/apps/els_core/src/els_config.erl
+++ b/apps/els_core/src/els_config.erl
@@ -50,7 +50,8 @@
                | root_uri
                | search_paths
                | code_reload
-               | elvis_config_path.
+               | elvis_config_path
+               | parse_transforms.
 
 -type path()  :: file:filename().
 -type state() :: #{ apps_dirs        => [path()]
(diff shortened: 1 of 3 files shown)


zwilias:
I should probably get this into a working state first.

Issue #829, opened 2020-12-09
Add support for inlining macros


robertoaloi:
It should be possible to trigger the inlining of a macro via a code action when hovering a macro definition.

PR #975, opened 2021-04-08
Show macro definition on hover

plux:
### Description

Show macro definition on hover.

We now store the define value in the POI, this should be helpful when implementing #829.


robertoaloi:
😲

Issue #862, opened 2021-01-20
go to definition when already on the definition


alanz:
Invoking goto definition on a function takes you to the definition.

Invoking it when already on the definition returns an empty result.

I believe other language servers treat this as `textDocument/references` which has the same return type, and is useful functionality.

Perhaps we should do this too.

robertoaloi:
Fixed by #978

PR #978, opened 2021-04-09
return references if no definitions are found

TheGeorge:
### If no definitions can be found, try to return references instead

This is inline with what most other LS are doing, when we jump-to-definition on a definition (or no definition can be found), try to return references instead. If there are no references, give up.


TheGeorge:
This fixes #862

robertoaloi:
Thanks!

PR #983, opened 2021-04-12
Fix bogus bound variable warnings in fun expression heads

### LINKED ISSUE #982: Bound in pattern for named funs
**Describe the bug**

In the code below:

```erlang
foo() ->
  fun F(#{ level := Level }) ->
    Level
  end.
```
Level is marked as "Bound variable in pattern: Level" when it should not be. If I remove the `F` name it works as expected.

Not sure if this is something that comes from syntax_tools or how erlang-ls uses it.

**Context**
 - `erlang_ls` version (tag/sha): 36bf0815e35db5
 - Editor used: emacs
 - Erlang version: 24.0-rc2

gomoripeti:
This is a workaround for erl_syntax_lib not considering shadowing in fun
expressions.

There is a further issue with named funs where variables from the name and head
are not added to the env of the fun body. This is not fixed by the current
change.

Fixes #982.

### DIFF
--- a/apps/els_lsp/priv/code_navigation/src/diagnostics_bound_var_in_pattern.erl
+++ b/apps/els_lsp/priv/code_navigation/src/diagnostics_bound_var_in_pattern.erl
@@ -1,6 +1,6 @@
 -module(diagnostics_bound_var_in_pattern).
 
--export([f/1, g/1, h/2]).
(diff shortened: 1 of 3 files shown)


robertoaloi:
Thanks!

PR #985, opened 2021-04-16
#981: Clear stalled indicator on termination

### LINKED ISSUE #981: Cancelling a request may result in a stale progress indicator
When a request is cancelled by the client, a background job maybe killed. As a consequence, a progress report may become stale on the client (e.g. showing a `0 / 1` or similar in the status bar), since an `end` message was never sent from the server to the client.

Original discussion:

```
Jesper Eskilson Today at 10:53 AM
erlang_ls seems to be leaving a bunch of "stale" 0/1 progress indicators in the Emacs modeline. I tried to reproduce this, but of course when I tried to provoke it, it didn't happen. See screenshot: (edited) 

roberto.aloi  8 minutes ago
Interesting. I just checked my LSP logs to see what could be.

roberto.aloi  7 minutes ago
cat /tmp/logs | grep -A3 '"kind": "begin"' | grep token > /tmp/begin
cat /tmp/logs | grep -A2 '"kind": "end"' | grep token > /tmp/end
cat /tmp/begin | sort > /tmp/begin-sorted
cat /tmp/token-end | sort > /tmp/end-sorted
diff /tmp/begin-sorted /tmp/end-sorted

roberto.aloi  6 minutes ago
This shows two tasks which start, but do not end

roberto.aloi  4 minutes ago
What I think happens is that, upon cancellation, the background job gets killed. But the final "report" message is not sent to the client.

roberto.aloi  3 minutes ago
I assumed the client would clear those for cancelled requests, but of course the reports are originated by the server, so Erlang LS should take care of the mapping and cleaning. Well spotted!
```

zsoci:
### Description

Clear stalled indicator on termination

Fixes #981 .

### DIFF
--- a/apps/els_lsp/src/els_background_job.erl
+++ b/apps/els_lsp/src/els_background_job.erl
@@ -189,9 +189,13 @@ terminate(normal, #{ config := #{on_complete := OnComplete}
   ok;
 terminate(Reason, #{ config := #{on_error := OnError}
                    , internal_state := InternalState
+                   , token := Token
+                   , total := Total
+                   , progress_enabled := ProgressEnabled
                    }) ->
   ?LOG_WARNING( "Background job aborted. [reason=~p] [pid=~p]"
               , [Reason, self()]),
+  notify_end(Token, Total, ProgressEnabled),
   OnError(InternalState),
   ok.
 


TheGeorge:
Thanks for fixing this!

The change looks good to me. I rerun the failing tests (els_server_SUITE is a bit flaky)

zsoci:
Thanks! Seems it passed now.

robertoaloi:
> s (els_server_SUITE is a bit flaky)

@TheGeorge Do you know what the root cause of flakyness is?

TheGeorge:
No, I just observed it on other PRs as well.

PR #996, opened 2021-05-06
[dap] Add basic support for break/logpoint (hit)conditions.

hajduakos:
### Description

This PR adds basic support for **conditions** and **hitconditions** for break/logpoints in the debugger ([see specification](https://microsoft.github.io/debug-adapter-protocol/specification#Types_SourceBreakpoint)).

**Details**

- A break/logpoint with a **condition** only triggers if the condition expression **evaluates to true**.
- Hitconditions currently must be non-negative integers N, meaning that only **every Nth hit** of a break/logpoint will trigger.
- Note that it is possible to have both a condition and a hitcondition for a break/logpoint, in this case we are only counting hits where the condition is true, and we only trigger at every Nth such hit.
- Note that conditions and hitconditions work the same way both for breakpoints and logpoints. The only difference is that breakpoints will stop, while logpoints will print a message and continue.

**Limitations**

Error handling is currently limited:
- If a condition cannot be evaluated, or does not evaluate to a Boolean, a warning message is displayed in the output and the condition is treated as false.
- If the hitcondition cannot be evaluated, or does not evaluate to a non-negative integer, a warning message is displayed in the output, the hitcondition is disregarded (treated as a hit).

**Testing**

New unit tests were added (`els_dap_general_provider_SUITE.erl`) and manual end-to-end testing was also carried out using the [daptoy project](https://github.com/erlang-ls/daptoy), focusing on all the possible combinations of break/logpoints with conditions and/or hitconditions. Manual testing actually revealed an interesting aspect of the protocol: if a break/logpoint is set without a (hit)condition, then there is no (hit)condition key in the mapping. However, if there is a (hit)condition, but is deleted later, a key is present with an empty value. This is handled correctly, and a unit test was also added for this behavior.

### DIFF
--- a/apps/els_dap/src/els_dap_general_provider.erl
+++ b/apps/els_dap/src/els_dap_general_provider.erl
@@ -46,14 +46,16 @@
                          , frames := #{frame_id() => frame()}
                          }.
 -type thread_id()    :: integer().
+-type mode()         :: undefined | running | stepping.
 -type state()        :: #{ threads => #{thread_id() => thread()}
                          , project_node => atom()
                          , launch_params => #{}
                          , scope_bindings =>
                           #{pos_integer() => {binding_type(), bindings()}}
                          , breakpoints := els_dap_breakpoints:breakpoints()
+                         , hits => #{line() => non_neg_integer()}
                          , timeout := timeout()
-                         , mode := undefined | running | stepping
+                         , mode := mode()
                          }.
 -type bindings()     :: [{varname(), term()}].
 -type varname()      :: atom() | string().
@@ -74,6 +76,7 @@ init() ->
    , launch_params => #{}
    , scope_bindings => #{}
    , breakpoints => #{}
+   , hits => #{}
    , timeout => 30
    , mode => undefined}.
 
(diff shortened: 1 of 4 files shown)


apps/els_dap/src/els_dap_general_provider.erl
```diff
@@ -425,39 +447,73 @@ handle_info( {int_cb, ThreadPid}
...
-                        [source(Module, ProjectNode), Line, Return])
-        ),
-        els_dap_server:send_event( <<"output">>
-                                 , #{ <<"output">> => LogMessage }),
-        case Mode0 of
-          running ->
-            els_dap_rpc:continue(ProjectNode, ThreadPid);
+  Breakpt = els_dap_breakpoints:type(Breakpoints, Module, Line),
+  %% evaluate condition if exists, otherwise treat as 'true'
+  Condition = case Breakpt of
```
TheGeorge, line 452: Can we break the implementation of the handle_info clause into smaller functions? It's 100 loc for one clause currently. Each of the blocks could be a sub-function

TheGeorge (changes requested):
Looks in general really good to me, but I think the handle_info needs t obe broken up a bit.

apps/els_dap/src/els_dap_general_provider.erl
hajduakos: Thanks for the feedback! Yes, that makes sense, will do.

TheGeorge (approved):
LGTM

Issue #995, opened 2021-05-05
misleading behaviour of suggest-spec lens when ~/.dialyzer_plt is missing


hellmean:
**Describe the bug**
Whenever ~/.dializer_plt is missing and also a corresponding setting is not set in erlang_ls.config you see "Cannot extract specs (check logs for details)". Moreover when you click on this lens this very same message is added as an extra line just above the function definition (where otherwise the spec would be added).

**To Reproduce**
(Re)move  ~/.dializer_pl and check. 

**Expected behavior**
As per documentation of erlang_ls.config: "Path to the dialyzer PLT file. When none is provided the dialyzer diagnostics will not be available.". I.e. the lens should be disabled when no plt file is found. It would be also nice to look for one in _build directory. Since this is where rebar3 would generate one. Also the user should be somehow informed that the plt file is not found and this is why the lens is not working. And it should not add a new line with an error message.

**Actual behavior**
See above in the description section. Thanks!

**Context**
 - `erlang_ls` version (tag/sha): 
 - Editor used: VS Code 1.55.2
 - LSP client used: 0.15.0

PR #1000, opened 2021-05-09
[FEATURE] truncate suggested spec title

### LINKED ISSUE #960: Suggested specs can be arbitrarily large
**Describe the bug**

The `suggest-spec` code lense sometimes shows ridiculously long specs.

**To Reproduce**

For example:
image

**Expected behavior**
Code lens should not be several lines

**Actual behavior**
Code lens is 7 lines line, and could probably be arbitrarily large.

**Context**
 - `erlang_ls` version (tag/sha): 0.13.0
 - Editor used: Emacs
 - LSP client used: 20210328.1404

hellmean:
also use empty command with a spec lens when failed to extract spec

### Description

The only editor that seem to be affected is Emacs. When suggested spec is too long it is wrapped over multiple lines in the editor window. In vim (coc) and Vs Code it is always presented as one very long line. Nonetheless it is probably good to avoid arbitrary long lens titles. This truncates the title to 100 chars by default (the font of the lens is typically smaller than normal text, hence it can be more than 80 chars).

Also fixes a bug when TypEr fails and no spec is available the lens with the error message was still clickable and would be producing a new line of text.

Fixes #960 and #995 .

### DIFF
--- a/apps/els_lsp/src/els_code_lens_suggest_spec.erl
+++ b/apps/els_lsp/src/els_code_lens_suggest_spec.erl
@@ -19,7 +19,7 @@
 %%==============================================================================
 %% Type Definitions
 %%==============================================================================
--type state() :: #{'info' => els_typer:info() | 'no_info'}.
+-type state() :: els_typer:info() | 'no_info'.
 
 %%==============================================================================
 %% Callback functions for the els_code_lens behaviour
@@ -28,7 +28,7 @@
 init(#{uri := Uri} = _Document) ->
   try els_typer:get_info(Uri) of
     Info ->
-      #{info => Info}
+      Info
   catch C:E:S ->
       Fmt =
         "Cannot extract typer info.~n"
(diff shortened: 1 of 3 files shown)


apps/els_lsp/src/els_code_lens_suggest_spec.erl
```diff
@@ -66,13 +70,27 @@ pois(Document) ->
...
-      re:replace(Spec, ",", ", ", [global, {return, binary}])
+    #{id := {Function, Arity}} = POI,
+    Spec = els_typer:get_type_spec(Function, Arity, Info),
+    re:replace(Spec, ",", ", ", [global, {return, binary}]).
+
+-spec truncate_spec_title(binary(), number()) -> binary().
+truncate_spec_title(Spec, MaxLength) ->
+  {Encoding, _BOMLength} = unicode:bom_to_encoding(Spec),
+  CodedSpec = unicode:characters_to_list(Spec, Encoding),
+  Length = string:length(CodedSpec),
```
gomoripeti, line 83: I'm not an expert on unicode so trying to follow what's happening here:
- it looks like `string:length/1` also works on binaries and returns a unicode-aware length, do we need to convert to list first?
- are you sure there is a BOM in the begining of `Spec` , otherwise I'm getting strange results:
```
1> Spec = unicode:characters_to_binary([200]).
<<"È"/utf8>>
2> {Encoding, _BOMLength} = unicode:bom_to_encoding(Spec).
{latin1,0}
3> CodedSpec = unicode:characters_to_list(Spec, Encoding).
[195,136]
4> string:length(CodedSpec).
2
```
I would expect a length of 1

apps/els_lsp/src/els_code_lens_suggest_spec.erl
```diff
@@ -66,13 +70,27 @@ pois(Document) ->
...
+    true ->
+      Title = unicode:characters_to_binary(
+        string:slice(CodedSpec, 0, MaxLength - 3)
+      ),
+      <<Title/binary, "...">>;
+    false ->
+      Spec
   end.
+
+-spec spec_title_max_length() -> number().
```
gomoripeti, line 94: nitpick: can max length be also `float()` or only `integer`?

gomoripeti (commented):
nice 👍 , I just have a couple questions

apps/els_lsp/src/els_code_lens_suggest_spec.erl
hellmean: haha:) neither am I. I didn't understand how `unicode:characters_to_list/2` works.  You are absolutely right, I can  use the binary directly. Fixed.

apps/els_lsp/src/els_code_lens_suggest_spec.erl
hellmean: fixed

apps/els_lsp/src/els_typer.erl
```diff
@@ -61,7 +61,7 @@
 -record(info, {records = maps:new() :: erl_types:type_table(),
                functions = []       :: [func_info()],
                types = map__new()   :: map_dict()}).
--opaque info() :: #info{}.
+-type info() :: #info{}.
```
gomoripeti, line 64: does this need to be non-opaque?

gomoripeti (approved):

apps/els_lsp/src/els_typer.erl
hellmean: this type is being used in the spec definition of `els_code_lens_suggest_spec:get_type_spec/2`. And it seems that due to it being opaque dialyzer thinks that there might be a type conflict for this function. Not sure, though, why it worked previously.

robertoaloi:
Thanks @hellmean and @gomoripeti for the review!

PR #1001, opened 2021-05-09
Support record fields in reference provider

gomoripeti:
References are only searched in the document containing the definition and
documents that (maybe recursively) include the defining document. The scoping
for macro and record referencing is also changed to the same.

### DIFF
--- a/apps/els_core/src/els_utils.erl
+++ b/apps/els_core/src/els_utils.erl
@@ -8,6 +8,8 @@
         , fold_files/4
         , halt/1
         , lookup_document/1
+        , include_id/1
+        , include_lib_id/1
         , macro_string_to_term/1
         , project_relative/1
         , resolve_paths/3
@@ -88,7 +90,8 @@ cmd_path(Cmd) ->
       Epmd
   end.
 
--spec filename_to_atom(els_dt_document:id()) -> atom().
+%% @doc Convert an 'include'/'include_lib' POI ID to a document index ID
+-spec filename_to_atom(string()) -> atom().
 filename_to_atom(FileName) ->
   list_to_atom(filename:basename(FileName, filename:extension(FileName))).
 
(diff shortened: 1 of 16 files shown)


apps/els_core/src/els_utils.erl
```diff
@@ -141,6 +144,23 @@ lookup_document(Uri) ->
       end
   end.
 
+%% @doc Convert path to an 'include' POI id
+-spec include_id(file:filename_all()) -> string().
+include_id(Path) ->
+  els_utils:to_list(filename:basename(Path)).
```
TheGeorge, line 150: Do we need to convert to list(char()) here? filename:basename() should return a string

gomoripeti: good question, I hit this while testing
the new code here https://github.com/erlang-ls/erlang_ls/pull/1001/files#diff-2ae95695a32108b6ba4383bafbc080cea105dbd9ba88c6a45474444ef3773471R71
calls `include_id/1` with a binary from `els_uri:path/1` and `filename:basename/1` returns a binary if called with a binary. (Similarly `filename:join/1` returns a binary if any of the components are binary)

apps/els_lsp/src/els_references_provider.erl
```diff
@@ -69,14 +72,23 @@ find_references(Uri, #{ kind := Kind
...
+find_references(Uri, Poi = #{kind := Kind})
+  when Kind =:= record_expr;
+       Kind =:= record_field;
+       Kind =:= macro ->
+  case els_code_navigation:goto_definition(Uri, Poi) of
+    {ok, DefUri, DefPoi} ->
+      uri_pois_to_locations(
+        find_scoped_references_for_def(DefUri, DefPoi));
+    {error, _} ->
+      []
```
gomoripeti, line 90: FIXME: could still look for references in the current document, even if definition is not yet declared

robertoaloi:
Amazing, thanks!

PR #1002, opened 2021-05-10
Implement support for incremental sync

plux:
### Description

Implement support for incremental text synchronization.
This should improve the experience when editing large files since we no longer need to transmit the entire text buffer for every change, instead a set of changes are submitted and applied.
To avoid triggering indexing for every update I've added `els_index_buffer` which will accumulate changes and index the accumulated result of the changes.

This feature is disabled by default but can be enabled by adding the follwing line to erlang_ls.config:
```yaml
incremental_sync: true
```

I've only tested these changes with Emacs lsp-mode.
Would be great if people with other editor setups could also test it out.

### DIFF
--- a/apps/els_core/src/els_text.erl
+++ b/apps/els_core/src/els_text.erl
@@ -8,8 +8,15 @@
         , line/3
         , range/3
         , tokens/1
+        , apply_edits/2
         ]).
 
+-export_type([edit/0]).
+
+-include("els_core.hrl").
+
+-type edit()       :: {poi_range(), string()}.
+-type lines()      :: [string() | binary()].
 -type text()       :: binary().
 -type line_num()   :: non_neg_integer().
 -type column_num() :: pos_integer().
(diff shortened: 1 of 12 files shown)


apps/els_core/src/els_text.erl
```diff
@@ -65,3 +127,94 @@ line_starts(Text) ->
 pos(LineStarts, {LineNum, ColumnNum}) ->
   {LinePos, _} = lists:nth(LineNum, LineStarts),
   LinePos + ColumnNum.
+
+-ifdef(TEST).
+-include_lib("eunit/include/eunit.hrl").
+apply_edits_single_line_test_() ->
```
plux, line 133: Seems like eunit tests aren't run by CI. Should I convert these to CT?

apps/els_core/src/els_text.erl
```diff
@@ -52,6 +59,61 @@ last_token(Text) ->
     Tokens -> lists:last(Tokens)
   end.
 
+-spec apply_edits(text(), [edit()]) -> text().
+apply_edits(Text, []) ->
+  Text;
+apply_edits(Text, Edits) when is_binary(Text) ->
+  lines_to_bin(apply_edits(bin_to_lines(Text), 1, 0, lists:sort(Edits))).
```
michalmuskala, line 66: AFAIK you can't sort edits - this can result in different content. They have to be applied in the order they are received. VS Code most of the time will sort them, so they are easier to apply, but this is not guaranteed, especially from other editors.

See this note in [documentation](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_didChange):
```
	/**
	 * The actual content changes. The content changes describe single state
	 * changes to the document. So if there are two content changes c1 (at
	 * array index 0) and c2 (at array index 1) for a document in state S then
	 * c1 moves the document from S to S' and c2 from S' to S''. So c1 is
	 * computed on the state S and c2 is computed on the state S'.
	 *
	 * To mirror the content of a document using change events use the following
	 * approach:
	 * - start with the same initial content
	 * - apply the 'textDocument/didChange' notifications in the order you
	 *   receive them.
	 * - apply the `TextDocumentContentChangeEvent`s in a single notification
	 *   in the order you receive them.
	 */
```

plux: Thanks for pointing this out!
lsp-mode in Emacs only ever sent edits one at a time, so I never actually tested this in a "real world" scenario.
It makes the implementation a lot simpler actually :+1:

apps/els_core/src/els_text.erl
```diff
@@ -63,47 +63,40 @@ last_token(Text) ->
...
+           {#{from := {L, FromC}, to := {L, ToC}}, Insert}) ->
   %% One line edit
-  {Prefix, Rest} = lists:split(FromC - 1 + Offset, CurrLine),
+  {Prefix, Rest} = lists:split(FromC - 1, CurrLine),
   {_, Suffix} = lists:split(ToC - FromC, Rest),
-  NewOffset = Offset + length(Insert) - (ToC - FromC),
-  apply_edits([Prefix ++ Insert ++ Suffix|RestLines], L, NewOffset, Edits);
-apply_edits([CurrLine|RestLines], L, Offset,
-            [{#{from := {L, FromC}, to := {ToL, ToC}}, Insert}|Edits]) ->
+  [Prefix ++ Insert ++ Suffix|RestLines];
```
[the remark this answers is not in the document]
plux: Good point! Will fix this.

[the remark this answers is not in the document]
plux: Pushed a fix for this

apps/els_core/src/els_text.erl
robertoaloi: Either that or add a `eunit` target here: https://github.com/erlang-ls/erlang_ls/blob/main/.github/workflows/build.yml#L44

apps/els_lsp/src/els_index_buffer.erl
```diff
@@ -0,0 +1,110 @@
+%% @doc Buffer textedits to avoid spamming indexing for every keystroke
+%%
+%% FIXME: Currently implemented as a simple process.
+%%        Might be nicer to rewrite this as a gen_server
```
robertoaloi, line 4: I would definitely rewrite this as a `gen_server`, but we can do that as a follow up.

robertoaloi:
Thanks for this @plux . Looks good to me and considering that this is disabled by default, I'm happy to merge it. Should we enable `eunit` in CI and maybe squash first?

plux:
> Thanks for this @plux . Looks good to me and considering that this is disabled by default, I'm happy to merge it. Should we enable `eunit` in CI and maybe squash first?

Sounds good, I will fix that!

plux:
I decided to create a CT suite instead of the eunit tests.
Squashed the commits.

From my perspective this is ready to be merged. :+1:

robertoaloi (approved):

PR #1003, opened 2021-05-11
Show how to install erlang_ls

pierre-rouleau:
### Description

Add missing description to show how to install erlang_ls once `make` was done.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Compile the project:
 
     make
 
-Install erlang_ls built by `make`:
+To install the produced `erlang_ls` escript in `/usr/local/bin`:
 
     make install
 
--- a/README.md
+++ b/README.md
@@ -18,6 +18,10 @@ Compile the project:
 
     make
 
+Install erlang_ls built by `make`:
+
+    make install
+
 ## Command-line Arguments
 
 These are the command-line arguments that can be provided to the


pierre-rouleau:
Hello @robertoaloi I agree.  The reason I added this text was because I was looking for instructions on how to install erlang_ls myself.  I ran make and then ... nothing.   So I searched where erlang_ls was  and it was nowhere to be found.  So I googled a bit and found nothing.  Then I looked in the make file and I saw  the install target,  so I ran `make install` and now I found erlang_ls.   But I have not yet started using it.  I was going to investigate this next.

I would like to see a little bit more info somewhere on how to install.  Perhaps a help target in the make would provide extra info that could be adjusted to the running environment.  At this point I don't know enough to contribute much though.

robertoaloi:
Also, here is the User Documentation for the project: https://erlang-ls.github.io/
It contains per-editor installation procedures but some steps may still be missing. And we should probably invest in things like `asdf` to make the installation process even smoother.

robertoaloi:
> Is there someplace that describes the erlang_ls features I should expect?

All Erlang LS features are currently available in Emacs via `lsp-mode` and `lsp-ui`. The best list we currently have is the one at erlang-ls.github.io, but it's not completely up to date and needs some love.

Feel free to contribute to the docs if you notice a feature is missing. The repo is here: 

https://github.com/erlang-ls/erlang-ls.github.io/tree/mkdocs/docs

pierre-rouleau:
Thanks, will do!

PR #1004, opened 2021-05-12
Allow renaming function when pointing at function reference

plux:
### Description

Previously functions could only be renamed when pointing at the definition, now a function can be renamed when  pointing at a reference too.

### DIFF
--- a/apps/els_lsp/src/els_rename_provider.erl
+++ b/apps/els_lsp/src/els_rename_provider.erl
@@ -58,7 +58,9 @@ workspace_edits(Uri, [#{kind := 'define'} = POI| _], NewName) ->
 workspace_edits(Uri, [#{kind := Kind} = POI| _], NewName)
   when Kind =:= macro;
        Kind =:= application;
-       Kind =:= implicit_fun ->
+       Kind =:= implicit_fun;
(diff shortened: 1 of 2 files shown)


apps/els_lsp/src/els_rename_provider.erl
```diff
@@ -55,7 +55,10 @@ workspace_edits(Uri, [#{kind := function_clause} = POI| _], NewName) ->
   #{changes => changes(Uri, POI#{kind => function, id => {F, A}}, NewName)};
 workspace_edits(Uri, [#{kind := 'define'} = POI| _], NewName) ->
   #{changes => changes(Uri, POI, NewName)};
-workspace_edits(Uri, [#{kind := 'macro'} = POI| _], NewName) ->
+workspace_edits(Uri, [#{kind := Kind} = POI| _], NewName)
+  when Kind =:= macro;
+       Kind =:= application;
+       Kind =:= implicit_fun ->
```
gomoripeti, line 61: how about supporting `export_entry` as well?

gomoripeti (approved):

robertoaloi:
Thanks!

PR #1005, opened 2021-05-12
Add "Find References" support to record fields

ruippeixotog:
As titled:

https://user-images.githubusercontent.com/613493/118038889-99e72d80-b367-11eb-9c92-b27d97379b23.mov

### DIFF
--- a/apps/els_lsp/src/els_dt_references.erl
+++ b/apps/els_lsp/src/els_dt_references.erl
@@ -100,7 +100,8 @@ find_by(Pattern) ->
   {ok, Items} = els_db:match(name(), Pattern),
   {ok, [to_item(Item) || Item <- Items]}.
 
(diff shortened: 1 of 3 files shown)


gomoripeti:
hi @ruippeixotog there seems to be some overlap with this PR https://github.com/erlang-ls/erlang_ls/pull/1001

Why I changed the scoping of record (and record_field) references in that PR is because it was a bit inconvenient that searching for the refs of `#state{}` record showed me all the states even OTP code. It is less a problem with record field as there is less chance there is a field with the same name in another record, but it can still happen.

robertoaloi:
Is this superseded by #1001 and can therefore be closed or there is something missing?

ruippeixotog:
We can close this, thanks for reminding me!

PR #1006, opened 2021-05-13
Properly extract line number from dialyzer warning

plux:
### Description
Dialyzer diagnostics crashes with bad arith on OTP 24.

Seems like this was broken in OTP 24 as the warning now returns a line and
column. Use `erl_anno:line/1` which will handle both `Line` and `{Line, Col}`

Side note: Seems like there are no tests for dialyzer diagnostics.

### DIFF
--- a/apps/els_lsp/src/els_dialyzer_diagnostics.erl
+++ b/apps/els_lsp/src/els_dialyzer_diagnostics.erl
@@ -66,9 +66,10 @@ source() ->
 %%==============================================================================
 %% Internal Functions
 %%==============================================================================
--spec diagnostic({any(), {any(), integer()}, any()}) ->
+-spec diagnostic({any(), {any(), erl_anno:anno()}, any()}) ->
         els_diagnostics:diagnostic().
-diagnostic({_, {_, Line}, _} = Warning) ->
+diagnostic({_, {_, Anno}, _} = Warning) ->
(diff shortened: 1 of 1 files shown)


apps/els_lsp/src/els_dialyzer_diagnostics.erl
```diff
@@ -66,9 +66,10 @@ source() ->
 %%==============================================================================
 %% Internal Functions
 %%==============================================================================
--spec diagnostic({any(), {any(), integer()}, any()}) ->
+-spec diagnostic({any(), {any(), erl_anno:anno()}, any()}) ->
         els_diagnostics:diagnostic().
-diagnostic({_, {_, Line}, _} = Warning) ->
+diagnostic({_, {_, Anno}, _} = Warning) ->
+  Line     = erl_anno:line(Anno),
```
gomoripeti, line 72: improvement idea: what if column value would be utilized as well if available?
(I dont want to mention the smartness of the compiler diagnostics which looks for a matching POI to even set the end location)

plux: Good idea! 
Maybe we should merge this since dialyzer diagnostics are currently broken andimprove this later. o

plux: Fixed in #1007

PR #1007, opened 2021-05-14
Make more exact dialyzer diagnostics by utilizing column number

plux:
### Description

Dialyzer diagnostics will now use the same range calculation as compiler diagnostics.

Follow up to suggestion in PR of #1006

### DIFF
--- a/apps/els_lsp/test/els_server_SUITE.erl
+++ b/apps/els_lsp/test/els_server_SUITE.erl
@@ -108,6 +108,7 @@ wait_until_no_lens_jobs() ->
 
 -spec get_current_lens_jobs() -> [pid()].
 get_current_lens_jobs() ->
-  #{internal_state := InternalState} = sys:get_state(els_code_lens_provider),
+  #{internal_state := InternalState} =
+    sys:get_state(els_code_lens_provider, 30 * 1000),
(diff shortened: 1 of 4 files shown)


gomoripeti (approved):
awesome 🎉

apps/els_lsp/test/els_server_SUITE.erl
```diff
@@ -108,6 +108,7 @@ wait_until_no_lens_jobs() ->
 
 -spec get_current_lens_jobs() -> [pid()].
 get_current_lens_jobs() ->
-  #{internal_state := InternalState} = sys:get_state(els_code_lens_provider),
+  #{internal_state := InternalState} =
+    sys:get_state(els_code_lens_provider, 30 * 1000),
```
robertoaloi, line 112: Do we know why this takes so long?

robertoaloi (approved):

robertoaloi:
Thanks! We should probably figure out why the `sys:get_state/1,2` takes so long.

apps/els_lsp/test/els_server_SUITE.erl
plux: No we should look into it!

PR #1008, opened 2021-05-16
Fix navigation to macros with overloaded arity

michalmuskala:
Macro name should be identified by both name & arity (if available).
Macros can be overloaded on arity. Currently, if that happens, the
navigation is incorrect and points to the first definition with
particular name, regardless of the actual arity.

### DIFF
--- a/apps/els_lsp/src/els_completion_provider.erl
+++ b/apps/els_lsp/src/els_completion_provider.erl
@@ -687,10 +687,16 @@ completion_item(#{kind := Kind = record, id := Name}, Data, _) ->
    , data             => Data
    };
 completion_item(#{kind := Kind = define, id := Name}, Data, _) ->
-  #{ label            => atom_to_binary(Name, utf8)
+  #{ label            => macro_label(Name)
    , kind             => completion_item_kind(Kind)
    , data             => Data
    }.
+
+-spec macro_label(atom() | {atom(), non_neg_integer()}) -> binary().
+macro_label({Name, Arity}) ->
+  els_utils:to_binary(io_lib:format("~ts/~p", [Name, Arity]));
+macro_label(Name) ->
+  atom_to_binary(Name, utf8).
 
 -spec snippet_function_call(atom(), [{integer(), string()}]) -> binary().
 snippet_function_call(Function, Args0) ->
--- a/apps/els_lsp/src/els_completion_provider.erl
+++ b/apps/els_lsp/src/els_completion_provider.erl
@@ -668,7 +668,7 @@ completion_item(#{kind := Kind, id := {F, A}, data := ArgsNames}, Data, false)
   Label = io_lib:format("~p/~p", [F, A]),
   #{ label            => els_utils:to_binary(Label)
    , kind             => completion_item_kind(Kind)
-   , insertText       => snippet_function_call(F, ArgsNames)
+   , insertText       => snippet_function(F, ArgsNames)
    , insertTextFormat => ?INSERT_TEXT_FORMAT_SNIPPET
    , data             => Data
    };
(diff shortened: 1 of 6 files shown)


TheGeorge (approved):

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -687,11 +687,17 @@ completion_item(#{kind := Kind = record, id := Name}, Data, _) ->
    , data             => Data
    };
 completion_item(#{kind := Kind = define, id := Name}, Data, _) ->
-  #{ label            => atom_to_binary(Name, utf8)
+  #{ label            => macro_label(Name)
```
gomoripeti, line 690: Im not sure how completions work. If there is no explicit `insertText` will this insert the label text upon selection? If yes, wouldn't it be better to not insert `/arity`? (Either insert just the name or an opening paren, or snippet with placeholder for args)

gomoripeti (approved):

apps/els_lsp/src/els_completion_provider.erl
robertoaloi: I believe @gomoripeti 's interpretation is correct here.

michalmuskala: Ah, that makes sense. I wasn't really sure what the `label` does. I'll change it to only include the name in here.

michalmuskala:
I pushed another commit that makes macro completions similar to function completions with snippets for arguments, when the definition had arguments

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -686,18 +686,39 @@ completion_item(#{kind := Kind = record, id := Name}, Data, _) ->
...
+  atom_to_binary(Name, utf8).
+
+-spec snippet_function(atom(), [{integer(), string()}]) -> binary().
+snippet_function(Name, Args) ->
+  snippet_args(atom_to_label(Name), Args).
+
+-spec snippet_macro( atom() | {atom(), non_neg_integer()}
+                   , [{integer(), string()}]) -> binary().
+snippet_macro({Name0, _Arity}, Args) ->
+  Name = atom_to_binary(Name0, utf8),
```
gomoripeti, line 711: one more nitpick: is it possible that the macro name contains special characters and must be a quoted atom?
For example in this contrived case Name should be `atom_to_label(Name0)`
```
-define('foo!', bar).
```
otoh if macro name is a variable it mustn't be quoted as `'Foo'`
```
-define(Foo, bar)
```
(to be fair the contrived edge case was not handled before this change either, so I'm fine to ignore it for now...and blame the one who tries to use such macros)

gomoripeti (approved):
very nice 👏

apps/els_lsp/src/els_completion_provider.erl
michalmuskala: It seems the best way would be to get the original text from the "name" node and just use that. But it looks like this is not available to get from the `erl_syntax` representation

gomoripeti: (I sometimes have these momentary failures of "edge-casing") let's ignore quoted macros all together

Just an interesting finding that I was not aware
```
-define(M, 1).
-define('M', 2).
```
The compiler errors with "redefining macro 'M'". (It treats a macro name of `{var, _, Name}` and `{atom, _, Name}` the same [in epp](https://github.com/erlang/otp/blob/master/lib/stdlib/src/epp.erl#L936))

(it is `els_erlfmt_ast` that does not copy the `text` from the erlfmt nodes to erl_syntax nodes, I omitted copying because it is not used yet it erlang_ls. An alternative to getting original text of a node is to extract it from the document text based on node range)

michalmuskala: In this case, should we merge this PR and then perhaps open an issue to properly review the quoted macros case?

michalmuskala: See https://github.com/erlang-ls/erlang_ls/issues/1011

PR #1014, opened 2021-05-20
Allow config files to be called erlang_ls.yaml

elbrujohalcon:
### Description

This PR allows erlang-ls users to call their configuration files `erlang_ls.yml` instead of `erlang_ls.config` to avoid mixing them up with config files, as in AdRoll/rebar3_format#259

### DIFF
--- a/apps/els_core/src/els_config.erl
+++ b/apps/els_core/src/els_config.erl
@@ -24,7 +24,7 @@
 %% Macros
 %%==============================================================================
 -define(DEFAULT_CONFIG_FILE, "erlang_ls.config").
--define(ALTERNATIVE_CONFIG_FILE, "erlang_ls.yml").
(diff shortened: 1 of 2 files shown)


apps/els_core/src/els_config.erl
```diff
@@ -24,6 +24,7 @@
 %% Macros
 %%==============================================================================
 -define(DEFAULT_CONFIG_FILE, "erlang_ls.config").
+-define(ALTERNATIVE_CONFIG_FILE, "erlang_ls.yml").
```
robertoaloi, line 27: @elbrujohalcon Is `.yml` or `.yaml` the most common extension?

elbrujohalcon: According to [The Great Oracle](https://stackoverflow.com/questions/21059124/is-it-yaml-or-yml), I should use `yaml`. Will change it in a moment.

robertoaloi (approved):

PR #1015, opened 2021-05-21
[dap] support "attach"

RunyaoZhang:
Different from "launch" which will start a shell, "attach" will attach to an existing shell.
See https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Attach

### Description

Add handle_request for "attach"
Change handle_request for "disconnect" if it was "attach" don't stop the node, only stop breakpoints.

Tests see comment.

Fixes # .

### DIFF
--- a/apps/els_dap/src/els_dap_general_provider.erl
+++ b/apps/els_dap/src/els_dap_general_provider.erl
@@ -452,7 +452,8 @@ handle_request({<<"disconnect">>, _Params}
       els_dap_rpc:no_break(ProjectNode),
       [els_dap_rpc:continue(ProjectNode, Pid) ||
         {_ThreadID, #{pid := Pid}} <- maps:to_list(Threads)],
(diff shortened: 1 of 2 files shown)


robertoaloi:
Hi @RunyaoZhang and really thanks for this!
The branch seems to have some linting issues, would you mind looking into those?
Looks good to me, @TheGeorge do you have any feedback on this?
We may want to factor some duplication out, but that can be also done as a follow up.

RunyaoZhang:
For second diff test:
1 Set break point
2 Start shell `rebar3 shell --sname abc@localhost --setcookie COOKIE `
3 Attach
4 `daptoy_fact:fact(2).` ( hit break point )
5 Disconnect
6 Noticed continue (previously not)
7 Attach again
....

TheGeorge:
LGTM let's wait for CI

RunyaoZhang:
Add dap rpc int:n/1 and int: interpreted/0 to fix dialyzer.
(Without these 2 it still work.)
Test same as before.

TheGeorge:
Can you also do a PR to the VSCode plugin?

PR #1023, opened 2021-06-01
Prevent infinite recursion when enumerating document POIs

keynslug:
### Description

Before this commit recursion in `els_scope:included_pois/2` might become infinite. This can happen for a couple of reasons:
* header mistakenly includes itself,
* header ends up including itself transitively, though inclusion is guarded with `ifndef` and `define` to prevent compiler from
  complaining.
* header includes another header with exactly same filename but  located in a different directory, yet from the point of view of  `els_utils:find_header/1` it's the same header.

This commit introduces simple abstract functional graph concept with traversals over set of nodes reachable from a given node and employs it to gather all POIs reachable through inclusion in the `els_scope` module.

### DIFF
--- a/apps/els_lsp/src/els_diagnostics_utils.erl
+++ b/apps/els_lsp/src/els_diagnostics_utils.erl
@@ -8,7 +8,9 @@
 %%==============================================================================
 -export([ dependencies/1
         , included_uris/1
+        , included_documents/1
         , range/2
+        , traverse_include_graph/3
         ]).
 %%==============================================================================
 %% Includes
@@ -24,6 +26,20 @@ dependencies(Uri) ->
 included_uris(Document) ->
   POIs = els_dt_document:pois(Document, [include, include_lib]),
   included_uris([Id || #{id := Id} <- POIs], []).
+
+-spec included_documents(els_dt_document:item()) -> [els_dt_document:item()].
+included_documents(Document) ->
+  lists:filtermap(fun find_included_document/1, included_uris(Document)).
+
+-spec find_included_document(uri()) -> {true, els_dt_document:item()} | false.
+find_included_document(Uri) ->
+  case els_utils:lookup_document(Uri) of
+    {ok, IncludeDocument} ->
+      {true, IncludeDocument};
+    {error, _} ->
+      ?LOG_WARNING("Failed included document lookup [uri=~p]", [Uri]),
+      false
+  end.
 
 -spec range(els_dt_document:item() | undefined,
             erl_anno:anno() | none) -> poi_range().
@@ -59,6 +75,17 @@ range(Document, Anno) ->
           maps:get(range, hd(POIs))
       end
   end.
+
+-spec traverse_include_graph(AccFun, AccT, From) -> AccT when
+    AccFun :: fun((Included, Includer, AccT) -> AccT),
+    From :: els_dt_document:item(),
+    Included :: els_dt_document:item(),
+    Includer :: els_dt_document:item().
+traverse_include_graph(AccFun, Acc, From) ->
+    Graph = els_fungraph:new(
+      fun els_dt_document:uri/1,
+      fun included_documents/1),
+    els_fungraph:traverse(AccFun, Acc, From, Graph).
 
 %%==============================================================================
 %% Internal Functions
--- a/apps/els_lsp/src/els_fungraph.erl
+++ b/apps/els_lsp/src/els_fungraph.erl
@@ -1,0 +1,45 @@
+-module(els_fungraph).
+
+-type id_fun(NodeT) :: fun((NodeT) -> _NodeID).
+-type edges_fun(NodeT) :: fun((NodeT) -> list(NodeT)).
+
+-opaque graph(NodeT) :: {?MODULE, id_fun(NodeT), edges_fun(NodeT)}.
+
+-export_type([graph/1]).
+
+-export([new/2]).
+-export([traverse/4]).
+
+-spec new(id_fun(NodeT), edges_fun(NodeT)) -> graph(NodeT).
+new(IdFun, EdgesFun) ->
+  {?MODULE, IdFun, EdgesFun}.
+
+-type acc_fun(NodeT, AccT) ::
+  fun((_Node :: NodeT, _From :: NodeT, AccT) -> AccT).
+
+-spec traverse(acc_fun(NodeT, AccT), AccT, NodeT, graph(NodeT)) -> AccT.
+traverse(AccFun, Acc, From, G) ->
+  traverse(AccFun, Acc, [From], sets:new(), G).
+
+-spec traverse(acc_fun(NodeT, AccT), AccT, [NodeT], sets:set(), graph(NodeT)) ->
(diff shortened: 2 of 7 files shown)


apps/els_lsp/src/els_fungraph.erl
```diff
@@ -0,0 +1,37 @@
+-module(els_fungraph).
+
+-type gnode() :: #{id := _ID, _ => _}.
+-type edges_fun() :: fun((node()) -> list(node())).
```
gomoripeti, line 4: this might be a typo? `node() -> gnode()`

keynslug: Oops, you're right. Thanks for a quick review!

apps/els_lsp/src/els_fungraph.erl
```diff
@@ -0,0 +1,38 @@
...
+
+-type acc_fun(AccT) :: fun((gnode(), AccT) -> AccT).
+
+-spec traverse(acc_fun(AccT), AccT, _From :: gnode(), graph()) -> AccT.
+traverse(AccFun, Acc, From, G) ->
+  traverse(AccFun, Acc, [From], sets:new(), G).
+
+-spec traverse(acc_fun(AccT), AccT, _From :: gnode(), sets:set(_NodeID), graph()) -> AccT.
+traverse(AccFun, Acc, Nodes, Visited, G = {?MODULE, EdgesFun}) ->
+  AdjacentNodes = [
```
gomoripeti, line 25: do I see correctly that `AdjacentNodes` might have duplicates if multiple Nodes reference the same document? (so an usort could help to avoid duplicates in `AccNext`)

gomoripeti (commented):
awesome! thank you for the contribution.

I just had a quick look, will finish later.

apps/els_lsp/src/els_fungraph.erl
keynslug: Again, spot on! I fixed it by rewriting into a depth-first traversal, for a bit of extra efficiency (hopefully).

robertoaloi:
Hi and thanks for this! We do something similar in `els_unused_includes_diagnostics`, I wonder if we refactor things to have a single implementation of this.

apps/els_lsp/src/els_fungraph.erl
```diff
@@ -0,0 +1,63 @@
...
+    end,
+    {[], Visited},
+    EdgesFun(Node)
+  ),
+  AccNext = lists:foldl(AccFun, Acc, AdjacentNodes),
+  traverse(AccFun, AccNext, AdjacentNodes ++ Rest, VisitedNext, G);
+traverse(_AccFun, Acc, [], _Visited, _G) ->
+  Acc.
+
+-ifdef(TEST).
```
robertoaloi, line 41: Nit: I would prefer the tests to be in their own module, especially in a case like this where all the required API is exported anyway.

keynslug: Fair point. I've addressed it in 69c3a1d.

keynslug:
> We do something similar in `els_unused_includes_diagnostics`, I wonder if we refactor things to have a single implementation of this.

Ah, thanks for the heads up. I made an attempt in 7017d65, what do you think?

apps/els_lsp/src/els_diagnostics_utils.erl
```diff
@@ -25,6 +26,26 @@ included_uris(Document) ->
   POIs = els_dt_document:pois(Document, [include, include_lib]),
   included_uris([Id || #{id := Id} <- POIs], []).
 
+-spec included_documents(els_dt_document:item()) -> [els_dt_document:item()].
```
robertoaloi, line 29: An alternative could be to reuse the `included_uris/1` function here and to `filtermap` on them with the `els_utils:lookup_document/1`.

robertoaloi (approved):

robertoaloi:
It would be good to squash before the merge.

apps/els_lsp/src/els_fungraph.erl
```diff
@@ -0,0 +1,43 @@
+-module(els_fungraph).
```
robertoaloi, line 1: I wonder if we could avoid rolling out our own traversal by using one of the `digraph_util` helpers or similar.

TheGeorge:
> It would be good to squash before the merge.

You can squash when merging 
Screen Shot 2021-06-04 at 1 17 30 PM

apps/els_lsp/src/els_fungraph.erl
robertoaloi: Let's eventually address this in a follow up PR.

robertoaloi:
> You can squash when merging

I normally avoid that because, from what I understood, no merge commits will be created for the merge, but all commits are simply squashed into one. I like merge commits since I use them to build the changelog. I guess we should also looking at automating that part, too.

apps/els_lsp/src/els_fungraph.erl
keynslug: > I wonder if we could avoid rolling out our own traversal by using one of the `digraph_util` helpers or similar.

From what I was able to conceive so far while working on the codebase, this would probably make perfect sense if we had a complete digraph at hand containing all the relationships (e.g. "includes", "depends on") between different modules and headers. Though this digraph would have to be built and maintained during indexing I believe. The "shared" nature of the digraph internal representation (aka 3 ETS tables) would even provide extra benefit for performance.

apps/els_lsp/src/els_diagnostics_utils.erl
keynslug: Did in fcab36c.

keynslug:
Before doing squashing I wondered if I should also attempt to express [this routine](https://github.com/erlang-ls/erlang_ls/blob/1f6f9edd2089c86d54729abe4074c1d52ea71263/apps/els_lsp/src/els_code_navigation.erl#L139-L151) with fungraph traversal. Or maybe it's followup PRs' material?

apps/els_lsp/src/els_fungraph.erl
robertoaloi: > Before doing squashing I wondered if I should also attempt to express [this routine](https://github.com/erlang-ls/erlang_ls/blob/1f6f9edd2089c86d54729abe4074c1d52ea71263/apps/els_lsp/src/els_code_navigation.erl#L139-L151) with fungraph traversal. Or maybe it's followup PRs' material?

I think that's totally fine not to include here.

robertoaloi:
> I did a bit of refactoring in [5574e19](https://github.com/erlang-ls/erlang_ls/commit/5574e197ef607ef5bedf595fc7e54645e52b64e8), @robertoaloi if this is LGTY I'll squash it all into a single commit at last.

LGTM!

keynslug:
Oops, unrelated commit accidentally sneaked in. Sorry for the noise.

keynslug:
> It would be good to squash before the merge.

Done.

robertoaloi:
Amazing, thanks for the contribution!

PR #1032, opened 2021-06-07
[#1031] Find implementations for callback functions

### LINKED ISSUE #1031: Go to implementation for callback functions
When hovering a callback function such as:

```
-callback to_be_implemented() -> ok.
```

It would be great to be able to retrieve a list of all available implementations and to jump to those.

robertoaloi:
### Description

Using the info from the DB, we can "jump to implementation" for callback functions. This can be especially useful in presence of _optional_ callbacks which only certain modules implement.

For example, the `els_code_lens` behaviour is used in 5 modules, but only 2 of them implement the `precondition/1` callback:

Screenshot 2021-06-07 at 09 53 09

`lsp-find-implementation`

Screenshot 2021-06-07 at 09 53 36

Fixes #1031 .

### DIFF
--- a/apps/els_lsp/priv/code_navigation/src/implementation.erl
+++ b/apps/els_lsp/priv/code_navigation/src/implementation.erl
@@ -1,0 +1,3 @@
+-module(implementation).
+
+-callback to_be_implemented() -> ok.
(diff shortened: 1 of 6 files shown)


TheGeorge (approved):

TheGeorge:
Nice! This will be useful to have.

Issue #992, opened 2021-04-27
Use LSP trace logging facilities


TheGeorge:
**Is your feature request related to a problem? Please describe.**
Currently, the LS logs only it's own log file, which makes issue not surface in the client (unless the LS crashes hard).

**Describe the solution you'd like**
The LSP defines trace log messages, see 
https://microsoft.github.io/language-server-protocol/specifications/specification-current/#logTrace

We should install an additional log-handler that emits these events to the client.

robertoaloi:
@TheGeorge Can we mark this one as done?

TheGeorge:
Yes, I forgot to close the issue.

PR #1033, opened 2021-06-07
[els_core] add logging via client notifications

TheGeorge:
### Description

This adds an additional loghandler, that forwards LSP logging to the client via `window/LogMessage` messages.
(see https://microsoft.github.io/language-server-protocol/specifications//specification-3-17/#window_logMessage)


TheGeorge:
this addresse #992 

Screen Shot 2021-06-07 at 2 28 48 PM

apps/els_lsp/src/els_log_notification.erl
```diff
@@ -0,0 +1,50 @@
...
+    ok = logger:add_handler( ?MODULE
+                           , ?MODULE
+                           , #{ level => LoggingLevel }
+                           ),
+    ignore.
+
+-spec log(logger:log_event(), logger:config_handler()) -> ok.
+log(#{level := Level} = LogEvent, _Config) ->
+    try
+        Msg = logger_formatter:format(LogEvent, #{ template => [ "[", time, "] "
```
robertoaloi, line 20: Something that we should also consider for the other logs is to make these messages a bit more readable. For example, we could start with `[level] msg` and have the remaining parts at the end. Not sure if it's easy, but we could only have the filename instead of the full path. WDYT?

apps/els_lsp/src/els_log_notification.erl
```diff
@@ -0,0 +1,50 @@
+-module(els_log_notification).
+
+-export([start/0]).
+-export([log/2]).
+
+-type lsp_message_type() :: 1 | 2 | 3 | 4.
```
robertoaloi, line 6: It would be good to map the `const`s from the specs to macros and use those instead of the numeric literals. For example:

```
-define(ERROR, 1)
```

And then the following:

```
map_otp_log_level_to_lsp(emergency) -> 1;
```

Would become:

```
map_otp_log_level_to_lsp(emergency) -> ?ERROR;
```

apps/els_lsp/src/els_sup.erl
```diff
@@ -79,6 +79,10 @@ init([]) ->
                , #{ id       => els_server
                   , start    => {els_server, start_link, []}
                   }
+               , #{ id       => els_log_notification
```
robertoaloi, line 82: Do we really need a process for this? Can't we just add the handler at boot?

TheGeorge: My reasoning was that the handler needs to start after els_server. The start function actually returns `ignore` so there is no process.

But you are right, I can just statically add the handler in `erlang_ls.erl`

apps/els_lsp/src/els_log_notification.erl
TheGeorge: agreed, let me play with the format a bit

TheGeorge:
Screen Shot 2021-06-07 at 5 44 14 PM

that's how it looks right now

TheGeorge:
Alright, last update to the format for now, editor and client will display as following:
```
[info] [2021-06-07T18:20:57.534105+01:00] els_background_job:init/1 L101 <0.208.0>
Background job started. [pid=<0.208.0>] [title=<<"Dialyzer (daptoy_fact)">>]
[info] [2021-06-07T18:20:57.534610+01:00] els_background_job:terminate/2 L187 <0.208.0>
Background job completed. [pid=<0.208.0>]
[info] [2021-06-07T18:20:57.534673+01:00] els_background_job:init/1 L101 <0.209.0>
Background job started. [pid=<0.209.0>] [title=<<"Elvis (daptoy_fact)">>]
[info] [2021-06-07T18:20:57.535272+01:00] els_background_job:init/1 L101 <0.210.0>
Background job started. [pid=<0.210.0>] [title=<<"UnusedIncludes (daptoy_fact)">>]
[info] [2021-06-07T18:20:57.535725+01:00] els_background_job:init/1 L101 <0.211.0>
Background job started. [pid=<0.211.0>] [title=<<"UnusedMacros (daptoy_fact)">>]
[info] [2021-06-07T18:20:57.536141+01:00] els_background_job:terminate/2 L187 <0.211.0>
Background job completed. [pid=<0.211.0>]
[info] [2021-06-07T18:20:57.538822+01:00] els_background_job:init/1 L101 <0.212.0>
Background job started. [pid=<0.212.0>] [title=<<"Lenses">>]
```

This will make debugging hopefully easier

robertoaloi (approved):

robertoaloi:
Nice!

PR #1036, opened 2021-06-09
Fix creating macro POIs in function applications

### LINKED ISSUE #1021: Unused macro is actually used
**Describe the bug**
ErlangLS warns about an unused macro. But the macro _is_ actually being used.

**To Reproduce**
```erlang
-module(foo).
-export([bar/0]).

-define(BAZ, baz).  % Unused macro: BAZ (incorrect)
-define(QUUX, quux).
-define(FNORD, fnord).  % Unused macro: FNORD (correct)

bar() ->
    % ...but it _is_ used.
    ?BAZ:call(?QUUX).
```

**Expected behavior**
- BAZ should not be reported as unused, because it is used. This is the bug.
- QUUX should not be reported as unused, because it is used. This is a control.
- FNORD should be reported as unused, because it is not used. This is a control.

**Actual behavior**
See comments in the snippet above.

**Context**
 - `erlang_ls` version (tag/sha): 0.15.0
 - Editor used: VS Code 1.56.2
 - LSP client used: erlang-ls.erlang-ls extension, 0.0.25

gomoripeti:
### Description

If we don't create a POI for a function application (because module or function is not a literal atom) then potentially create POIs from its subtrees.

Fixes #1021  .

### DIFF
--- a/apps/els_lsp/priv/code_navigation/src/diagnostics_unused_macros.erl
+++ b/apps/els_lsp/priv/code_navigation/src/diagnostics_unused_macros.erl
@@ -4,6 +4,8 @@
 
 -define(USED_MACRO, used_macro).
 -define(UNUSED_MACRO, unused_macro).
+-define(MOD, module). %% MOD was incorrectly reported as unused (#1021)
 
 main() ->
+  ?MOD:foo(),
   ?USED_MACRO.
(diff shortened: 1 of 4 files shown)


robertoaloi:
Thanks!

PR #1038, opened 2021-06-13
Add support for hover on record expression.

ztion:
With this change you can hover over a record expression, and get a definition of the record.


robertoaloi:
Thanks. I wonder if we should start making the _hover_ behaviours configurable (it can be done as a follow up).

PR #1040, opened 2021-06-16
Ignore Emacs backup and temporary files

pierre-rouleau:
### Description

Just get Git to ignore Emacs backup and temporary files. That should be an impact for anything else.

Fixes # .


.gitignore
```diff
@@ -24,3 +24,8 @@ current_counterexample.eqc
 *.log
 *.log.*
 doc
+# Emacs Backup files
+*~
+# Emacs temporary files
+.#*
```
robertoaloi, line 30: Hi, thanks for this. Shouldn't the `#` be escaped?

pierre-rouleau:
As far as I know if the first character is not # you don't need to escape it.

.gitignore
pierre-rouleau: As far as I know it needs to be deleted if it is the first character.  
That made me think that one is missing:   \#*

robertoaloi:
Thanks!

PR #1048, opened 2021-06-29
BSP support

al-khanji:
Started adding BSP support, this is an RFC.

### DIFF
--- a/apps/els_lsp/src/els_bsp_client.erl
+++ b/apps/els_lsp/src/els_bsp_client.erl
@@ -28,6 +28,8 @@
         , request/2
         , notification/1
         , notification/2
+        , wait_response/2
+        , check_response/2
         ]).
 
 %%==============================================================================
@@ -40,14 +42,15 @@
 %% Defines
 %%==============================================================================
 -define(SERVER, ?MODULE).
--define(TIMEOUT, 10000).
+-define(BSP_WILDCARD, "*.json").
+-define(BSP_CONF_DIR, ".bsp").
 
 %%==============================================================================
 %% Record Definitions
 %%==============================================================================
--record(state, { request_id = 1 :: request_id()
-               , pending    = [] :: [pending_request()]
-               , port       :: port() | 'undefined'
+-record(state, { request_id = 1    :: request_id()
+               , pending    = []   :: [pending_request()]
+               , port              :: port() | 'undefined'
                , buffer     = <<>> :: binary()
                }).
 
@@ -61,6 +64,78 @@
 -type pending_request() :: [{request_id(), from()}].
 -type from()            :: {pid(), any()}.
 
+
+%%==============================================================================
+%% Compat Stuff
+%% Since the build server can take arbitrary amounts of time to process things
+%% we really would like to use gen_server:send_request et co that are added in
+%% OTP 23, but we also need to support older OTP versions for now - implement
+%% rudimentary scaffolding to fake the new functionality.
+%% Remove whenever only OTP >= 23 is supported.
+%%==============================================================================
+-type server_ref() :: atom() | pid().
+
+-if(?OTP_RELEASE >= 23).
+-spec do_send_request(server_ref(), any()) -> any().
+do_send_request(ServerRef, Request) ->
+  gen_server:send_request(ServerRef, Request).
+
+-spec do_wait_response(any(), timeout()) ->
+        {reply, any()} |
+        timeout |
+        {error, {any(), server_ref()}}.
+do_wait_response(RequestId, Timeout) ->
+  gen_server:wait_response(RequestId, Timeout).
+
+-spec do_check_response(any(), any()) ->
+        {reply, any()} |
+        no_reply |
+        {error, {any(), server_ref()}}.
+do_check_response(Msg, RequestId) ->
+  gen_server:check_response(Msg, RequestId).
+-else.
+-spec do_send_request(server_ref(), any()) -> any().
+do_send_request(ServerRef, Request) ->
+  Self = self(),
+  Ref = erlang:make_ref(),
+  F = fun() ->
+          Result = gen_server:call(ServerRef, Request, infinity),
+          try Self ! {Ref, Result} catch _:_ -> ok end
+      end,
+  {Pid, Mon} = erlang:spawn_monitor(F),
+  {Pid, Mon, Ref, ServerRef}.
+
+-spec do_wait_response(any(), timeout()) ->
+        {reply, any()} |
+        timeout |
+        {error, {any(), server_ref()}}.
+do_wait_response({_Pid, Mon, Ref, ServerRef}, Timeout) ->
+  receive
+    {Ref, Result} ->
+      erlang:demonitor(Mon, [flush]),
+      {reply, Result};
+    {'DOWN', Mon, _Type, _Object, Info} ->
+      {error, {Info, ServerRef}}
+  after Timeout ->
+      timeout
+  end.
+
+-spec do_check_response(any(), any()) ->
+        {reply, any()} |
+        no_reply |
+        {error, {any(), server_ref()}}.
+do_check_response(Msg, {_Pid, Mon, Ref, ServerRef}) ->
+  case Msg of
+    {Ref, Result} ->
+      erlang:demonitor(Mon, [flush]),
+      {reply, Result};
+    {'DOWN', Mon, _Type, _Object, Info} ->
+      {error, {Info, ServerRef}};
+    _ ->
+      no_reply
+  end.
+-endif.
+
 %%==============================================================================
 %% API
 %%==============================================================================
@@ -68,7 +143,7 @@
 start_link() ->
   gen_server:start_link({local, ?SERVER}, ?MODULE, [], []).
 
--spec start_server(uri()) -> ok.
+-spec start_server(uri()) -> {ok, map()} | {error, any()}.
 start_server(RootUri) ->
   gen_server:call(?SERVER, {start_server, RootUri}).
 
@@ -82,7 +157,7 @@ notification(Method) ->
 
 -spec notification(method(), params()) -> any().
 notification(Method, Params) ->
-  gen_server:call(?SERVER, {notification, Method, Params}, ?TIMEOUT).
+  gen_server:cast(?SERVER, {notification, Method, Params}).
 
 -spec request(method()) -> any().
 request(Method) ->
@@ -90,7 +165,17 @@ request(Method) ->
 
 -spec request(method(), params()) -> any().
 request(Method, Params) ->
-  gen_server:call(?SERVER, {request, Method, Params}, ?TIMEOUT).
+  do_send_request(?SERVER, {request, Method, Params}).
+
+-spec wait_response(any(), timeout()) ->
+        {reply, any()} | timeout | {error, {any(), any()}}.
+wait_response(RequestId, Timeout) ->
+  do_wait_response(RequestId, Timeout).
+
+-spec check_response(any(), any()) ->
+        {reply, any()} | no_reply | {error, {any(), any()}}.
+check_response(Msg, RequestId) ->
+  do_check_response(Msg, RequestId).
 
 %%==============================================================================
 %% gen_server Callback Functions
@@ -103,20 +188,21 @@ init([]) ->
 -spec handle_call(any(), any() , state()) ->
         {noreply, state()} | {reply, any(), state()}.
 handle_call({start_server, RootUri}, _From, State) ->
-  RootPath = binary_to_list(els_uri:path(RootUri)),
-  [C|_] = filelib:wildcard(filename:join([RootPath, ".bsp", "*.json"])),
-  {ok, Content} = file:read_file(C),
-  #{ argv := [Cmd|Params]
-   , env := Env0} = jsx:decode(Content, [return_maps, {labels, atom}]),
-  Executable = os:find_executable(binary_to_list(Cmd)),
-  Args = [binary_to_list(P) || P <- Params],
-  Env = [{atom_to_list(K), binary_to_list(V)} || {K, V} <- maps:to_list(Env0)],
-  Opts = [{args, Args}, use_stdio, binary, {env, Env}],
-  ?LOG_INFO( "Start BSP Server [executable=~p] [args=~p] [env=~p]"
-           , [Executable, Args, Env]
-           ),
-  Port = open_port({spawn_executable, Executable}, Opts),
-  {reply, ok, State#state{port = Port}};
+  RootPath = els_uri:path(RootUri),
+  case find_config(RootPath) of
+    undefined ->
+      ?LOG_INFO("Found no BSP configuration. [root=~p]", [RootPath]),
+      {reply, {error, noconfig}, State};
+    #{ argv := [Cmd|Params] } = Config ->
+      Executable = os:find_executable(binary_to_list(Cmd)),
+      Args = [binary_to_list(P) || P <- Params],
+      Opts = [{args, Args}, use_stdio, binary],
+      ?LOG_INFO( "Start BSP Server [executable=~p] [args=~p]"
+               , [Executable, Args]
+               ),
+      Port = open_port({spawn_executable, Executable}, Opts),
+      {reply, {ok, Config}, State#state{port = Port}}
+  end;
 handle_call({request, Method, Params}, From, State) ->
   #state{port = Port, request_id = RequestId, pending = Pending} = State,
   ?LOG_INFO( "Sending BSP Request [id=~p] [method=~p] [params=~p]"
@@ -126,18 +212,16 @@ handle_call({request, Method, Params}, From, State) ->
   port_command(Port, Payload),
   {noreply, State#state{ request_id = RequestId + 1
                        , pending    = [{RequestId, From} | Pending]
-                       }};
-handle_call({notification, Method, Params}, _From, State) ->
+                       }}.
+
+-spec handle_cast(any(), state()) -> {noreply, state()}.
+handle_cast({notification, Method, Params}, State) ->
   ?LOG_INFO( "Sending BSP Notification [method=~p] [params=~p]"
            , [Method, Params]
            ),
   #state{port = Port} = State,
   Payload = els_protocol:notification(Method, Params),
   port_command(Port, Payload),
-  {reply, ok, State}.
-
--spec handle_cast(any(), state()) -> {noreply, state()}.
-handle_cast(_Request, State) ->
   {noreply, State}.
 
 -spec handle_info(any(), state()) -> {noreply, state()}.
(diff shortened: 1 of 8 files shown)


apps/els_lsp/src/els_bsp_provider.erl
```diff
@@ -0,0 +1,147 @@
...
+handle_sources(ConfigKey, SourceFun, Response, State) ->
+  Result = maps:get(result, Response, #{}),
+  Items = maps:get(items, Result, []),
+  Sources = lists:flatten([ maps:get(sources, Item, []) || Item <- Items ]),
+  Uris = lists:flatten([ SourceFun(Source) || Source <- Sources ]),
+  NewPaths = lists:flatten([ maps:get(path, uri_string:parse(Uri), []) || Uri <- Uris ]),
+  OldPaths = els_config:get(ConfigKey),
+  AllPaths = lists:usort([ els_utils:to_list(P) || P <- OldPaths ++ NewPaths]),
+  els_config:set(ConfigKey, AllPaths),
+  els_indexing:start(),
```
robertoaloi, line 123: Won't this get executed multiple times? Once for apps and once for deps. On top of that, the general provider will also trigger it.

apps/els_lsp/src/els_bsp_provider.erl
```diff
@@ -0,0 +1,147 @@
...
+-spec is_enabled() -> true.
+is_enabled() -> true.
+
+-spec handle_request({start, uri()}, state()) -> {ok, state()}.
+handle_request({start, RootUri}, #{ running := false } = State) ->
+  ?LOG_INFO("Starting BSP server in ~p", [RootUri]),
+  case els_bsp_client:start_server(RootUri) of
+    {ok, Config} ->
+      ?LOG_INFO("BSP server started from config ~p", [Config]),
+      enqueue(initialize_bsp),
```
robertoaloi, line 56: Do we need to do this asynchronously? It looks like the code handling the message makes an async request anyway?

robertoaloi (commented):
Thanks for pushing this! It's nice to see the BSP work resumed.

apps/els_lsp/src/els_bsp_client.erl
```diff
@@ -82,15 +85,23 @@ notification(Method) ->
...
 -spec request(method()) -> any().
 request(Method) ->
   request(Method, #{}).
 
 -spec request(method(), params()) -> any().
 request(Method, Params) ->
-  gen_server:call(?SERVER, {request, Method, Params}, ?TIMEOUT).
+  gen_server:send_request(?SERVER, {request, Method, Params}).
+
+-spec wait_response(any(), timeout()) -> {reply, any()} | timeout | {error, {any(), any()}}.
```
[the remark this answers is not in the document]
al-khanji: You're right - I've become too used to otp23 niceties I guess :) I'll refactor it to work on older versions as well.

apps/els_lsp/src/els_bsp_client.erl
```diff
@@ -160,6 +169,36 @@ terminate(_Reason, #state{port = Port} = _State) ->
...
+  choose_config(Candidates).
+
+-spec choose_config([file:filename()]) -> map() | undefined.
+choose_config([]) ->
+  undefined;
+choose_config([F|Fs]) ->
+  try
+    {ok, Content} = file:read_file(F),
+    Config = jsx:decode(Content, [return_maps, {labels, atom}]),
+    Languages = case Config of
```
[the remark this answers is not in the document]
al-khanji: Right - initially I didn't have the try/catch around this bit of code and was trying to guard against json that isn't an object at top level. Will give the entire function a once over.

apps/els_lsp/src/els_general_provider.erl
```diff
@@ -86,21 +86,19 @@ handle_request({initialized, _Params}, State) ->
...
-             , <<"capabilities">> => #{ <<"languageIds">> => [<<"erlang">>] }
-             , <<"data">>         => #{}
-             }
-       );
     false ->
-      ok
+      ok;
+    BspEnabled ->
+      BspStatus = els_bsp_provider:start(RootUri),
+      case {BspEnabled, BspStatus} of
```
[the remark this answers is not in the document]
al-khanji: It is handled - if `false`, we just return `ok` always. Otherwise, try to start the server. If it starts up, great! If not, and `BspEnabled` is `true`, we log an error and kill the LS. If `BspEnabled` is something else (ostensibly `auto`) we log a warning but carry on.

apps/els_lsp/src/els_bsp_provider.erl
al-khanji: Yeah, and it doesn't respect the `initializationOptions` either.

There is some tradeoff here - the BS might take a relatively long time to start up, for example if the project in question has many dependencies to fetch. In that case it might make sense to start indexing right away, as the config file might still contain path entries.

I was hoping the indexing system might be smart enough to detect that a given file hasn't changed since it was last indexed, and that reindexing might be cheap. If not, I think we could make it smarter.

What do you suggest?

apps/els_lsp/src/els_bsp_provider.erl
al-khanji: We could send the request right here, when I was writing it the intent was to separate starting (which is the current request) from initialization (which arguably is a separate step). I can combine the two here if you prefer that.

apps/els_lsp/src/els_general_provider.erl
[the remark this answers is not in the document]
robertoaloi: So maybe document `auto` and make it explicit in the matching?

apps/els_lsp/src/els_bsp_provider.erl
robertoaloi: As a start, I would probably wait for the BS and then follow up with a smarter mechanism if needed. WDYT?

al-khanji:
Any general feedback at this point?

apps/els_lsp/src/els_indexing.erl
```diff
@@ -112,6 +113,17 @@ index_references(#{uri := Uri} = Document, 'deep', false) ->
 index_references(_Document, 'shallow', _) ->
   ok.
 
+-spec maybe_start() -> true | false.
+maybe_start() ->
```
robertoaloi, line 117: It looks like the return value of this function is never used, so it could simply return `ok`.

apps/els_lsp/src/els_indexing.erl
```diff
@@ -112,6 +113,17 @@ index_references(#{uri := Uri} = Document, 'deep', false) ->
...
+  case els_config:get(indexing_enabled) =:= false of
```
robertoaloi, line 118: It may be more readable written as:
```
case els_config:get(indexing_enabled) of
  false ->
    ?LOG_INFO("Skipping Indexing (disabled via InitOptions)");
  X when X =:= true; X =:= auto ->
    start()
  end
```

apps/els_lsp/src/els_formatting_provider.erl
```diff
@@ -92,38 +98,56 @@ handle_request({document_ontypeformatting, Params}, State) ->
...
         end,
   tempdir:mktmp(Fun).
 
--spec format_document_local(binary(), string(), formatting_options()) ->
-        [text_edit()].
-format_document_local(Path, RelativePath,
+-spec format_document_bsp(string(), string(), formatting_options()) ->
+           boolean().
+format_document_bsp(Dir, RelativePath, _Options) ->
+  Method = <<"rebar3/run">>,
```
robertoaloi, line 119: By calling the method `rebar3/run` we are breaking the abstraction a bit, aren't we?

apps/els_lsp/src/els_bsp_provider.erl
```diff
@@ -0,0 +1,262 @@
...
+%%==============================================================================
+%% Includes
+%%==============================================================================
+-include("els_lsp.hrl").
+-include_lib("kernel/include/logger.hrl").
+
+%%==============================================================================
+%% Types
+%%==============================================================================
+-type state() :: #{ running          := boolean()         % BSP server running?
```
robertoaloi, line 29: If I remember correctly, providers are registered by name. So, it may be simpler to check with a `erlang:is_process_alive(whereis(els_bsp_provider))` if the BSP provider is running or not. This would avoid the need of the `running` parameter and the `info` method.

apps/els_lsp/src/els_bsp_provider.erl
```diff
@@ -0,0 +1,262 @@
...
+      erlang:error(badarg, [Item])
+  end.
+
+-spec request(method(), params()) -> request_response().
+request(Method, Params) ->
+  RequestId = send_request(Method, Params),
+  wait_response(RequestId, infinity).
+
+-spec send_request(method(), params()) -> request_id().
+send_request(Method, Params) ->
```
robertoaloi, line 73: My gut feeling is that we should be able to do this with higher-level OTP primitives instead of expicit monitoring and receive, but we can tackle this as an eventual follow up.

robertoaloi (commented):
> Any general feedback at this point?

Considering this PR should not affect any existing feature and that moves the BSP piece of work towards the right(TM) direction, I don't have a problem to integrate it early and refine it incrementally.
My main feedback for the future would be to try and submit smaller changesets, to keep review cycles smaller and faster. Still not sure we should implement BSP support via the `provider` behaviour itself, but we can start as is.
Should we disable the provider by default, though?

apps/els_lsp/src/els_bsp_provider.erl
```diff
@@ -0,0 +1,262 @@
...
+-spec init() -> state().
+init() ->
+  #{ running          => false
+   , root_uri         => undefined
+   , pending_requests => []
+   , pending_sources  => [ apps_paths, deps_paths ]
+   }.
+
+-spec is_enabled() -> true.
+is_enabled() -> true.
```
robertoaloi, line 106: Should we have this disabled by default?

apps/els_lsp/src/els_bsp_client.erl
```diff
@@ -61,14 +64,86 @@
 -type pending_request() :: [{request_id(), from()}].
 -type from()            :: {pid(), any()}.
 
+
+%%==============================================================================
```
TheGeorge, line 68: How may of these do we have in the code base? If it's just this one, I'd say it's ok to have it in the module, otherwise it would make sense to factor these into a separate module:
- makes this usable from other code as well
- makes cleanup easier

apps/els_lsp/src/els_bsp_client.erl
```diff
@@ -103,20 +188,21 @@ init([]) ->
...
-  Opts = [{args, Args}, use_stdio, binary, {env, Env}],
-  ?LOG_INFO( "Start BSP Server [executable=~p] [args=~p] [env=~p]"
-           , [Executable, Args, Env]
-           ),
-  Port = open_port({spawn_executable, Executable}, Opts),
-  {reply, ok, State#state{port = Port}};
+  RootPath = els_uri:path(RootUri),
+  case find_config(RootPath) of
+    undefined ->
+      ?LOG_INFO("Found no BSP configuration. [root=~p]", [RootPath]),
```
TheGeorge, line 194: This should probably be an error, or if there is a fallback a warning
```suggestion
      ?LOG_ERROR("Found no BSP configuration. [root=~p]", [RootPath]),
```

apps/els_lsp/src/els_bsp_client.erl
```diff
@@ -160,6 +244,34 @@ terminate(_Reason, #state{port = Port} = _State) ->
 %%==============================================================================
 %% Internal Functions
 %%==============================================================================
+-spec find_config(els_uri:path()) -> map() | undefined.
+find_config(RootDir) ->
+  Wildcard = filename:join([RootDir, ?BSP_CONF_DIR, ?BSP_WILDCARD]),
+  Candidates = filelib:wildcard(els_utils:to_list(Wildcard)),
```
TheGeorge, line 250: This works with arbitrary strings (e.g. `["/",["*"]]` is valid input)
```suggestion
  Candidates = filelib:wildcard(Wildcard),
```

apps/els_lsp/src/els_bsp_client.erl
```diff
@@ -40,14 +42,15 @@
 %% Defines
 %%==============================================================================
 -define(SERVER, ?MODULE).
--define(TIMEOUT, 10000).
+-define(BSP_WILDCARD, "*.json").
+-define(BSP_CONF_DIR, ".bsp").
```
TheGeorge, line 46: Is this convention? So there is a 

```
.bsp/
```
with some json files which look like
```
{language: ["erlang"], argv = ["rebar3", "...", ...]}
```

TheGeorge (commented):
The main thing missing is some tests for the changes

apps/els_lsp/src/els_bsp_client.erl
robertoaloi: The _Build Server Discovery_ mechanisim and format are documented [here](https://build-server-protocol.github.io/docs/server-discovery.html#the-bsp-connection-details).

apps/els_lsp/src/els_bsp_client.erl
al-khanji: I think it's just the one at the moment - I went ahead and factored it out in the BSP plugin: [rebar3_bsp_gen_server.erl](https://github.com/erlang-ls/rebar3_bsp/blob/d8ed6f020a97ace8f808f7b18eb02224e3e20c08/src/rebar3_bsp_gen_server.erl). It could just be copy/pasted as is. The linked version also has some improvements in the behavior.

apps/els_lsp/src/els_indexing.erl
al-khanji: For my own education - is it preferred in the LS code base to return descriptive values only if they are used?

apps/els_lsp/src/els_formatting_provider.erl
al-khanji: We are, but we were doing so previously as well - there is no guarantee that the format plugin would be installed.

apps/els_lsp/src/els_bsp_provider.erl
al-khanji: The BSP provider does run though even if the BSP server hasn't been started.

apps/els_lsp/src/els_bsp_provider.erl
al-khanji: The provider being enabled doesn't mean that the build server is started.

apps/els_lsp/src/els_bsp_client.erl
al-khanji: I think it's correct the way it is - the RootDir is going to be a binary, hence the output from filename:join is also a binary and must be converted to a list before being passed to filelib:wildcard.

al-khanji:
> Considering this PR should not affect any existing feature and that moves the BSP piece of work towards the right(TM) direction, I don't have a problem to integrate it early and refine it incrementally.

That sounds great!

> My main feedback for the future would be to try and submit smaller changesets, to keep review cycles smaller and faster.

To be honest I struggled quite a while to make the thing work reliably - mostly on the rebar3 side though. I hope that followups will indeed be smaller.

> Still not sure we should implement BSP support via the `provider` behaviour itself, but we can start as is.

I'm also on the fence about that - it did seem like a good idea when I started writing it, but I am kind of leaning towards just making it a gen_server at this point.

> Should we disable the provider by default, though?

The provider being enabled doesn't automatically launch a build server - it's only done if there is a configuration file found. Is there another concern?

apps/els_lsp/src/els_bsp_provider.erl
al-khanji: We should probably turn it into a gen_server and use the same send_request/receive_request etc. functionality as the client already uses.

al-khanji:
I will update the PR with squashed commits tomorrow, I did the BSP plugin side today.

al-khanji:
I can't seem to replicate the failing ct test case locally.

robertoaloi:
> I can't seem to replicate the failing ct test case locally.

The test also fails locally for me, even in isolation. The failure is not deterministic, though, and it seems to occur more often after a clean build.

```
$ rm -rf _build
$ rebar3 ct --suite apps/els_lsp/test/els_references_SUITE.erl --case purge_references
[...]
%%% els_references_SUITE ==> purge_references: FAILED
%%% els_references_SUITE ==>
Failure/Error: ?assertEqual({ok,[#{id => {foo,foo,1},range => #{from => {3,10},to => {3,13}},uri => <<102,105,108,101,58,47,47,47,116,109,112,47,102,111,111,46,101,114,108>>}]}, els_dt_references : find_all ( ))
  expected: {ok,[#{id => {foo,foo,1},
                   range => #{from => {3,10},to => {3,13}},
                   uri => <<"file:///tmp/foo.erl">>}]}
       got: {ok,[#{id => {foo,foo,1},
                   range => #{from => {3,10},to => {3,13}},
                   uri => <<"file:///tmp/foo.erl">>},
                 #{id => "transitive.hrl",
                   range => #{from => {4,1},to => {4,28}},
                   uri =>
                       <<"file:///Users/robert.aloi/git/github/erlang-ls/erlang_ls/_build/test/lib/els_lsp/priv/code_navigation/include/transitive.hrl">>},
                 #{id => "definition.hrl",
                   range => #{from => {5,1},to => {5,28}},
                   uri =>
                       <<"file:///Users/robert.aloi/git/github/erlang-ls/erlang_ls/_build/test/lib/els_lsp/priv/code_navigation/include/transitive.hrl">>}]}
      line: 414
```

I suspect this is linked to the changes to `initOptions` and `indexingEnabled`. That mechanism was introduced to ease testing, so that, in case of tests, only files explicitly mentioned would be indexed **synchronously**. The options seem to be now discarded by the server, so indexing happens asynchronously. In the above case, an extra reference is found.

al-khanji:
> > I can't seem to replicate the failing ct test case locally.
> 
> 
> 
> The test also fails locally for me, even in isolation. The failure is not deterministic, though, and it seems to occur more often after a clean build.
> 
> 
> 
> ```
> 
> $ rm -rf _build
> 
> $ rebar3 ct --suite apps/els_lsp/test/els_references_SUITE.erl --case purge_references
> 
> [...]
> 
> %%% els_references_SUITE ==> purge_references: FAILED
> 
> %%% els_references_SUITE ==>
> 
> Failure/Error: ?assertEqual({ok,[#{id => {foo,foo,1},range => #{from => {3,10},to => {3,13}},uri => <<102,105,108,101,58,47,47,47,116,109,112,47,102,111,111,46,101,114,108>>}]}, els_dt_references : find_all ( ))
> 
>   expected: {ok,[#{id => {foo,foo,1},
> 
>                    range => #{from => {3,10},to => {3,13}},
> 
>                    uri => <<"file:///tmp/foo.erl">>}]}
> 
>        got: {ok,[#{id => {foo,foo,1},
> 
>                    range => #{from => {3,10},to => {3,13}},
> 
>                    uri => <<"file:///tmp/foo.erl">>},
> 
>                  #{id => "transitive.hrl",
> 
>                    range => #{from => {4,1},to => {4,28}},
> 
>                    uri =>
> 
>                        <<"file:///Users/robert.aloi/git/github/erlang-ls/erlang_ls/_build/test/lib/els_lsp/priv/code_navigation/include/transitive.hrl">>},
> 
>                  #{id => "definition.hrl",
> 
>                    range => #{from => {5,1},to => {5,28}},
> 
>                    uri =>
> 
>                        <<"file:///Users/robert.aloi/git/github/erlang-ls/erlang_ls/_build/test/lib/els_lsp/priv/code_navigation/include/transitive.hrl">>}]}
> 
>       line: 414
> 
> ```
> 
> 
> 
> I suspect this is linked to the changes to `initOptions` and `indexingEnabled`. That mechanism was introduced to ease testing, so that, in case of tests, only files explicitly mentioned would be indexed **synchronously**. The options seem to be now discarded by the server, so indexing happens asynchronously. In the above case, an extra reference is found.

Thanks for the explanation. I'll find some time during the next couple of days to debug this.

al-khanji:
@robertoaloi I think it's good to go now.

apps/els_lsp/src/els_indexing.erl
robertoaloi: I guess it's a matter of personal preference and it depends on the specific case. If we foresee a possible usage of the return value it makes sense to expose it, otherwise `ok` is quite common for a function whose purpose is mainly a side effect.

robertoaloi:
Let's get this one in, so we can iterate over it. @al-khanji would you mind adding a section to the docs explaining how to get started with the experimental BSP support? Maybe via an article at https://github.com/erlang-ls/erlang-ls.github.io/tree/mkdocs/docs/articles ?

al-khanji:
> Let's get this one in, so we can iterate over it. @al-khanji would you mind adding a section to the docs explaining how to get started with the experimental BSP support? Maybe via an article at https://github.com/erlang-ls/erlang-ls.github.io/tree/mkdocs/docs/articles ?

Yes, I have been meaning to. Thanks for the nudge!

PR #1062, opened 2021-07-13
Specify category for DAP events

robertoaloi:
### Description

In Emacs, the events were not displayed in the UI otherwise.


TheGeorge:
Can you do a quick test if that changes the behavior from VSCode?

robertoaloi:
> Can you do a quick test if that changes the behavior from VSCode as this changes the category from `console` to `stdout`

Checked and the output still seems to be sent to the "Debug Console", so no changes in VS Code.

robertoaloi:
Reported the issue upstream in `dap-mode` https://github.com/emacs-lsp/dap-mode/issues/491

Issue #1047, opened 2021-06-29
Invalid unused include warning when including eunit.hrl


eproxus:
**Describe the bug**
Consider the following module:
```erlang
-module(my_tests).
-include_lib("eunit/include/eunit.hrl").
the_test() -> ok.
```

A warning will be generated on line 2: `warning Unused file: eunit.hrl ​UnusedIncludes`

**To Reproduce**
Create the module.

**Expected behavior**
No warning is issued since the include file is used, because it exports the `the_test/0` function. If the file is not included, the test doesn't work.

**Actual behavior**
A false warning is issued: `warning Unused file: eunit.hrl ​UnusedIncludes`

**Context**
 - `erlang_ls` version (tag/sha): 0.17.0
 - Editor used: Sublime Text 4 (4109)
 - LSP client used: Sublime LSP v1.6.1

gomoripeti:
thanks Adam, this is indeed a bug. Just cross-linking with the related https://github.com/erlang-ls/erlang_ls/issues/964

robertoaloi:
What we could have (unless we manage to fix all corner cases) is a ignore-list for header files. Candidates could be:

```
eunit/include/eunit.hrl
proper/include/proper.hrl
eqc/include/eqc.hrl
pmod_transform/include/pmod.hrl
stdlib/include/qlc.hrl
stdlib/include/ms_transform.hrl
```

The user could customize the list if required. Similar to the [Exclude OTP Apps list](https://github.com/erlang-ls/erlang_ls/blob/main/apps/els_core/src/els_config.erl#L94)

michalmuskala:
Shouldn't any header file that includes `-parse_transform` attribute be excluded like this? I think this should cover all of the above

robertoaloi:
That should work. I was thinking to cases "other" than parse transforms, but maybe there isn't any.

gomoripeti:
to be very precise it is not a -parse_transform attribute, but a compiler option in a -compile attribute, but probably the presence of any compiler option should make the header be excluded.
for "other" cases Daniel brings up in the other ticket wild attributes that are not known by the compiler.

robertoaloi:
@eproxus This should now be solved in the lastest `HEAD`, let me know how it goes.

Issue #964, opened 2021-04-01
Do not mark includes as unused if they contain module attributes


dszoboszlay:
**Describe the bug**
The `unused_includes` diagnostic reports include files unused even if they contain module attributes, such as `-compile` directives. E.g. `-include_lib("stdlib/include/ms_transform.hrl").` will always be reported as unused, but removing it would disable an important parse transform.

**Context**
 - `erlang_ls` version (tag/sha): 0.13.0
 - Editor used: VS Code

dszoboszlay:
Just to clarify: there are a handful of module attributes that compiler knows about and treats them in a special way (e.g. `-ifdef`, `-define`, `-compile`, `-include` etc.). The diagnostic shall also understand these and know which one are "unused" (e.g. things inside an `-ifdef` section that is going to be removed by the preprocessor; a `-define` that is not used later etc.). It may even go fancy and report redundant things as unused (e.g. repeated `-compile` options).

But the mere presence of most of these special attributes will change the compiled module, hence the include shouldn't be marked as unused. And if there are module attributes that are not understood by the compiler, they must be preserved too because they will turn up in `module_info(attributes)`.

kvelicka:
+1, got somewhat caught out by `erlang_ls` suggesting `ms_transform.hrl` was unused in a module with a few `ets:fun2ms` calls.

PR #1072, opened 2021-08-17
Add config option to exclude unused headers from warnings.

NAR:
### Description

A new option is added to the configuration (exclude_unused_includes).
Header files specified by this configuration option are not checked weather
they are used or not.

Workaround for  #1047  and #964.

### DIFF
--- a/apps/els_lsp/src/els_unused_includes_diagnostics.erl
+++ b/apps/els_lsp/src/els_unused_includes_diagnostics.erl
@@ -64,7 +64,17 @@ find_unused_includes(#{uri := Uri} = Document) ->
     , type_application
     , export_type_entry
     ]),
-  IncludedUris = els_diagnostics_utils:included_uris(Document),
+  IncludedUris0 = els_diagnostics_utils:included_uris(Document),
+  ExcludeUnusedIncludes =
+      lists:filtermap(
+          fun(Include) ->
+              case els_utils:find_header(els_utils:filename_to_atom(Include)) of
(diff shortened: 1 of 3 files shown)


apps/els_lsp/src/els_unused_includes_diagnostics.erl
```diff
@@ -64,7 +64,17 @@ find_unused_includes(#{uri := Uri} = Document) ->
     , type_application
     , export_type_entry
     ]),
-  IncludedUris = els_diagnostics_utils:included_uris(Document),
+  IncludedUris0 = els_diagnostics_utils:included_uris(Document),
+  ExcludeUnusedIncludes =
```
NAR, line 68: Maybe do this file path expansion in the `els_config` module and print a warning there if the header file is not found?

robertoaloi: I would imagine developers reusing their list of `exclude_unused_includes` across multiple projects (which may not include all entries), so that should not be necessary.

robertoaloi (approved):
Thanks for this! I still think we should implement what @michalmuskala suggested in https://github.com/erlang-ls/erlang_ls/issues/1047#issuecomment-872093838, but I believe a configurable list in addition to that may be handy. Would you mind adding the new option to the documentation at https://github.com/erlang-ls/erlang-ls.github.io/blob/mkdocs/docs/configuration.md ?

NAR:
Sure, I'm going to add this to the documentation.

PR #1074, opened 2021-08-17
[#1073] Find non-parametrized macro definitions.

### LINKED ISSUE #1073: erlang_ls can't find definition of ?config macro from ct.hrl
**Describe the bug**

The `?config` macro is used in many `common_test` testcases to fetch input data for testcases. However, `erlang_ls` doesn't seem to be able to find it's definition (and thus reports the `ct.hrl` include file unused). The problem seems to be that when `erlang_ls` parses the source file of the testcase, it sees a `config/2` macro while the actual definition is this:
```
-define(config,test_server:lookup_config).
```
It doesn't even look like a "function" definition for the parser.

**To Reproduce**

Check out e.g. the `els_formatter_SUITE` module, it uses the `?config` macro.

**Expected behavior**

I expected `erlang_ls` to find the `?config` macro.

**Actual behavior**

The macro is not found.

**Context**
 - `erlang_ls` version (tag/sha): c42f6dd05c3802a254e0a67c1f4158f3e63539af
 - Editor used: vim
 - LSP client used: CoC

NAR:
### Description

Some macro definitions that accept parameters do not actually specify
them at declaration, so erlang_ls doesn't find them. One such example is
the ?config macro in the common_test ct.hrl file. So in case a macro
usage looks like it passes some parameters, but erlang_ls can't find a
macro definition that accepts parameters, let's try to find a macro
definition without a parameter list (based on experiments the erlang
compiler seems to do this too).

Fixes #1073  .


NAR (commented):
Needs a testcase!

robertoaloi:
Hi @NAR , thanks for the contribution. It looks like the pipeline is failing on a trivial style check:

```
# apps/els_lsp/src/els_code_navigation.erl [FAIL]
  - line_length
    - Line 85 is too long: goto_definition(Uri, #{ kind := macro, id := {MacroName, _Arity} = Define } = POI) ->.
    ```

NAR:
I fixed the too long line earlier, I hope that's correct.

robertoaloi:
Looks good to me, thanks!

PR #1077, opened 2021-08-25
Simplify protocol implementation

al-khanji:
Use erlang:decode_packet/3 to do the heavy lifting for us. Less code, and might also be faster.

### Description

This pull request drops the remnants of cowlib from erlang_ls and uses built-in OTP functionality instead. This is very similar to the approach I took in rebar3_bsp.


robertoaloi (approved):

robertoaloi:
Thanks!

PR #1078, opened 2021-08-31
Introduce diagnostic codes for the OTP compile

alanz:
Expose the module and error passed to `Module:print_error` as an error code.

This allows us to provide telemetry of the codes, for later analysis, in
situations where this is practical.

It also allows an easy tie-up to any external documentation about how to deal
with the error.

The mechanism is to use the source of the original `format_error/2` clauses to map
onto codes, which have a regular format, e.g. `L1227` where the first letter
indicates the linter, and the rest is just a 4 digit unique number.

This format fits in with with existing error presentation in IDEs, as it is
already used by other compilers such as rust.

### DIFF
--- a/apps/els_lsp/src/els_compiler_diagnostics.erl
+++ b/apps/els_lsp/src/els_compiler_diagnostics.erl
@@ -190,10 +190,12 @@ diagnostic(_Path, MessagePath, Range, Document, Module, Desc0, Severity) ->
 diagnostic(Range, Module, Desc, Severity) ->
   Message0 = lists:flatten(Module:format_error(Desc)),
   Message  = els_utils:to_binary(Message0),
+  Code = make_code(Module, Desc),
   #{ range    => els_protocol:range(Range)
    , message  => Message
    , severity => Severity
    , source   => source()
+   , code     => Code
    }.
 
 %% @doc NOP function for the call to 'Module:format_error/1' in diagnostic/4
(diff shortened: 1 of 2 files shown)


apps/els_lsp/src/els_compiler_diagnostics.erl
```diff
@@ -202,6 +204,436 @@ diagnostic(Range, Module, Desc, Severity) ->
...
+%-----------------------------------------------------------------------
+
+%% @doc Return a unique code for each of the possible errors returned by the
+%% compiler or its subsystems.
+
+-spec make_code(atom(), any()) -> binary().
+
+%% This file
+make_code(els_compiler_diagnostics, _) ->
+ unicode:characters_to_binary("L0000");
```
KennethL, line 217: I really don't understand why you are not just returning the binary directly as a literal like this:
```erlang
make_code(els_compiler_diagnostics, _) ->
 <<"L0000">>;
```

alanz: Good point.

robertoaloi:
I like the idea and I noticed the same concept in Rust, as you say. I am happy to experiment with this a bit, to see what we can build on top of this. In general, think the correct way to proceed would be to raise a proper EEP about this. Would you be interested @KennethL ?

apps/els_lsp/src/els_compiler_diagnostics.erl
```diff
@@ -202,6 +204,436 @@ diagnostic(Range, Module, Desc, Severity) ->
...
+
+%-----------------------------------------------------------------------
+
+%% @doc Return a unique code for each of the possible errors returned by the
+%% compiler or its subsystems.
+
+-spec make_code(atom(), any()) -> binary().
+
+%% This file
+make_code(els_compiler_diagnostics, _) ->
```
robertoaloi, line 216: We can proceed like this, but we may consider refactoring things a bit. For example, we could move the codes to a separate module/header and maybe derive the initial letter from the module itself.

apps/els_lsp/src/els_compiler_diagnostics.erl
```diff
@@ -202,6 +204,436 @@ diagnostic(Range, Module, Desc, Severity) ->
...
+ <<"P1710">>;
+make_code(erl_parse, ["syntax error before: ", _]) ->
+ <<"P1711">>;
+%% Matching 'io_lib:format("bad ~tw declaration", [S])).', must come last
+make_code(erl_parse, "bad " ++ _Str) ->
+ <<"P1798">>;
+make_code(erl_parse, _Other) ->
+ <<"P1799">>;
+
+make_code(Module, Reason) ->
```
robertoaloi, line 631: Maybe we should use `U` as `undefined` prefix...

robertoaloi (approved):

KennethL:
I am positive about introducing diagnostic codes for the compiler, but I
would like to see some examples of use cases where this information is used
in some interesting way.
If RUST already have diagnostic codes, how are they used there?

alanz:
> If RUST already have diagnostic codes, how are they used there?

The most important use is it allows documentation to refer to them in a uniform way, e.g. https://doc.rust-lang.org/error-index.html

And they are presented in the IDE, and in future (possibly) from the compiler, so it becomes a lot easier to search for any posts about dealing with the particular error.

Apart from these, in my day job we are able to log telemetry from our IDE clients, and I want to get a handle on which errors are most prevalent in in day to day usage, and how they cluster, with a view to providing appropriate code actions, etc.

robertoaloi:
I tend to agree with @KennethL . Let's see if we end up building something interesting based on the codes on our side and eventually bring them to OTP *after*.

PR #1079, opened 2021-09-01
Send telemetry notification with compiler diagnostic codes

alanz:
When enabled, send a `telemetry/event` notification containing the URI and error codes for compiler diagnostics in the given file.

The allows analysis of diagnostics type prevalence if the client logs it.

It is disabled by default.

This PR depends on #1078 .

### DIFF
--- a/apps/els_lsp/src/els_compiler_diagnostics.erl
+++ b/apps/els_lsp/src/els_compiler_diagnostics.erl
@@ -22,6 +22,7 @@
 
 -export([ include_options/0
         , macro_options/0
+        , telemetry/2
         ]).
 
 %%==============================================================================
@@ -76,6 +77,7 @@ source() ->
 
 -spec on_complete(uri(), [els_diagnostics:diagnostic()]) -> ok.
 on_complete(Uri, Diagnostics) ->
+  ?MODULE:telemetry(Uri, Diagnostics),
   maybe_compile_and_load(Uri, Diagnostics).
 
 %%==============================================================================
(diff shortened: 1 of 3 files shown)


apps/els_lsp/src/els_compiler_diagnostics.erl
```diff
@@ -76,6 +77,7 @@ source() ->
 
 -spec on_complete(uri(), [els_diagnostics:diagnostic()]) -> ok.
 on_complete(Uri, Diagnostics) ->
+  ?MODULE:telemetry(Uri, Diagnostics),
```
robertoaloi, line 80: Nit: any reason for the fully qualified call

robertoaloi (approved):

robertoaloi:
@alanz Looks good, could you please rebase and solve the conflict? Also:

* It would be great if you could expand on what you have in mind on the client side for the telemetry
* Can you please add the new option to the docs website?

apps/els_lsp/src/els_compiler_diagnostics.erl
alanz: So that we can mock it in the tests.

robertoaloi: In that case it should be enough to export the function. Unless we are mocking both `telemetry/2` and `on_complete/2` (in which case I think meck requires the fully qualified call to be able to work.

alanz: If I call the local version only, it never sees the mocked one.  This is the same reason `?MODULE:publish(...` is called in https://github.com/erlang-ls/erlang_ls/blob/d50da89ebb01e885c06d82b5f338205573f0fc45/apps/els_lsp/src/els_diagnostics_provider.erl#L61

robertoaloi: And I was the one who did that 😅 I guess it's because of the `passthrough` to the module itself...

robertoaloi (approved):

PR #1084, opened 2021-09-14
Fix DAP not finding the source for a module

maxno-kivra:
### Description

This makes the els_dap_rpc try much harder to find the source, by first
asking `int:file/1`, then `code` and `filelib:find_source/1` and
finally falling back to the current behaviour of using `module_info`
and checking the `source` compilation option.

This allows you to build dependencies/OTP in one place, and run it from
another.


robertoaloi (approved):

robertoaloi:
Nice, thanks!

PR #1085, opened 2021-09-19
[#1047] Mark headers with compiler attributes as used

### LINKED ISSUE #964: Do not mark includes as unused if they contain module attributes

robertoaloi:
### Description

Fixes #964 and partially #1047.

### DIFF
--- a/apps/els_lsp/src/els_unused_includes_diagnostics.erl
+++ b/apps/els_lsp/src/els_unused_includes_diagnostics.erl
@@ -65,6 +65,7 @@ find_unused_includes(#{uri := Uri} = Document) ->
     , export_type_entry
     ]),
   IncludedUris0 = els_diagnostics_utils:included_uris(Document),
+  IncludedUris1 = filter_includes_with_compiler_attributes(IncludedUris0),
   ExcludeUnusedIncludes =
       lists:filtermap(
           fun(Include) ->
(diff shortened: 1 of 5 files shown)


apps/els_lsp/src/els_unused_includes_diagnostics.erl
```diff
@@ -120,3 +121,31 @@ inclusion_range(Uri, Document) ->
     _ ->
       #{from => {1, 1}, to => {2, 1}}
   end.
+
+%% @doc Given a list of included uris, filter out the ones containing
+%% compiler attributes. If the Uri cannot be found, keep it in the list.
```
gomoripeti, line 126: to me it looks like if lookup_document fails the Uri is filtered out.

gomoripeti (approved):

apps/els_lsp/src/els_unused_includes_diagnostics.erl
robertoaloi: Nice catch. Fixed (the code, not the comment).

PR #1091, opened 2021-09-27
user background job for hover provider to avoid blocking request queue

TheGeorge:
### Description

When cmd+click on a function (or similar) in a large document, the client might produce the following requests:
- hover request (when having the mouse over the element)
- cancel request (when clicking)
- definitions request

The hover request can take several seconds, and is synchronous currently. All actions that follow the hover request are stalled until it is served. This change uses background jobs for hover instead, which makes them cancellable and provides lower latencies for follow-up requests, e.g. definitions.

### DIFF
--- a/apps/els_lsp/src/els_hover_provider.erl
+++ b/apps/els_lsp/src/els_hover_provider.erl
@@ -83,7 +83,7 @@ run_hover_job(Uri, Line, Character) ->
   {ok, Pid} = els_background_job:new(Config),
   Pid.
 
--spec get_docs({uri(), [poi()]}, undefined) -> map() | null.
+-spec get_docs({uri(), integer(), integer()}, undefined) -> map() | null.
 get_docs({Uri, Line, Character}, _) ->
   {ok, Doc} = els_utils:lookup_document(Uri),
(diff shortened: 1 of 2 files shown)


apps/els_lsp/src/els_hover_provider.erl
```diff
@@ -23,24 +31,72 @@
...
+
+
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+
+-spec run_hover_job(uri(), line(), column()) -> pid().
+run_hover_job(Uri, Line, Character) ->
   {ok, Doc} = els_utils:lookup_document(Uri),
   POIs = els_dt_document:get_element_at_pos(Doc, Line + 1, Character + 1),
```
al-khanji, line 75: I assume you've measured that these operations are fast? They could be moved into the background job as well.

TheGeorge: Hoestly, it's a copy paste artifact. Let's move them into the background job

TheGeorge: I moved it into the background job

robertoaloi:
Thanks!

PR #1096, opened 2021-09-29
[#1094] Add support for Call Hierarchy

### LINKED ISSUE #1094: Introduce support for Call Hierarchy
Implement the following methods according to specification:

* https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_prepareCallHierarchy
* https://microsoft.github.io/language-server-protocol/specifications/specification-current/#callHierarchy_incomingCalls
* https://microsoft.github.io/language-server-protocol/specifications/specification-current/#callHierarchy_outgoingCalls

robertoaloi:
### Description

This PR introduces support for _Call Hierarchy_.

Screenshot 2021-09-29 at 17 38 32

For a nicer overview of what the _Call Hierarchy_ feature provides please watch [this video](https://www.youtube.com/watch?v=r5LA7ivUb2c).

There are still a couple of improvements possible (for example when it comes to recursive functions), which will be addressed as follow-ups.

The feature steps up the game when it comes to finding references (aka _who calls this function_). Erlang LS is already able to quickly identify callers, but that approach forces the developer to keep in mind the tree of _already explored paths_. Now that mental tree is available in the UI.

#### Emacs Support

To use it In Emacs you will require the `lsp-treemacs` package.

To see the incoming calls hierarchy:

```
M-x lsp-treemacs-call-hierarchy
```

To see the outgoing calls hierarchy:

```
C-u M-x lsp-treemacs-call-hierarchy
```

#### VS Code Support

Before this is available in VS Code we need to bump the `languageclient` package to `7.0.0`, which is a breaking change.

Fixes #1094.


apps/els_core/src/els_utils.erl
```diff
@@ -417,3 +420,21 @@ compose_node_name(Name, Type) ->
...
       list_to_atom(NodeName ++ "." ++ Domain)
   end.
+
+%% @doc Given an MFA or a FA, return a printable version of the
+%% function signature, in binary format.
+-spec function_signature( {atom(), atom(), non_neg_integer()} |
+                          {atom(), non_neg_integer()}) ->
+        binary().
+function_signature({M, F, A}) ->
+  els_utils:to_binary(io_lib:format("~p:~ts/~p", [M, F, A]));
```
TheGeorge, line 430:
```suggestion
  els_utils:to_binary(io_lib:format("~p:~ts/~b", [M, F, A]));
```

apps/els_lsp/src/els_call_hierarchy_provider.erl
```diff
@@ -0,0 +1,112 @@
...
+  #{ <<"item">> := #{<<"uri">> := Uri} = Item } = Params,
+  POI = els_call_hierarchy_item:poi(Item),
+  References = els_references_provider:find_references(Uri, POI),
+  Items = [reference_to_item(Reference) || Reference <- References],
+  {incoming_calls(Items), State};
+handle_request({outgoing_calls, Params}, State) ->
+  #{ <<"item">> := Item } = Params,
+  #{ <<"uri">> := Uri } = Item,
+  POI = els_call_hierarchy_item:poi(Item),
+  Applications = applications_in_function_range(Uri, POI),
```
TheGeorge, line 48: this took me a while to understand, applications are function applications, but in Erlang applications usually mean Erlang Applications.

Also I understand applications more as applications of the marked function. Maybe rename to calls_in_function_range or outgoing_calls

apps/els_lsp/src/els_dt_document.erl
```diff
@@ -168,3 +172,27 @@ get_element_at_pos(Item, Line, Column) ->
...
+
+-spec applications_at_pos(item(), non_neg_integer(), non_neg_integer()) ->
+        [poi()].
+applications_at_pos(Item, Line, Column) ->
+  POIs = get_element_at_pos(Item, Line, Column),
+  [POI || #{kind := 'application'} = POI <- POIs].
+
+-spec wrapping_functions(item(), non_neg_integer(), non_neg_integer()) ->
+        [poi()].
+wrapping_functions(Document, Line, Column) ->
```
TheGeorge, line 189: Is this expected to return more than one function?

How is the behavior with funs?

apps/els_lsp/src/els_call_hierarchy_provider.erl
```diff
@@ -0,0 +1,112 @@
...
+  #{id := Id, range := Range} = Function,
+  Name = els_utils:function_signature(Id),
+  Data = #{poi => Function},
+  els_call_hierarchy_item:new(Name, Uri, Range, Range, Data).
+
+-spec reference_to_item(location()) -> els_call_hierarchy_item:item().
+reference_to_item(Reference) ->
+  #{uri := RefUri, range := RefRange} = Reference,
+  {ok, RefDoc} = els_utils:lookup_document(RefUri),
+  [WrappingPOI] = els_dt_document:wrapping_functions(RefDoc, RefRange),
```
TheGeorge, line 85: can this return more than one POI?

apps/els_lsp/src/els_call_hierarchy_provider.erl
robertoaloi: Yeah, it can be confusing. This refers to `application` in [erl_syntax](https://erlang.org/doc/man/erl_syntax.html#application-2) terms. We should probably bring clarity around this, either by renaming or via a comment.

apps/els_lsp/src/els_dt_document.erl
robertoaloi: Good eye :)
It will be always be  a `[]` or a list with exactly one element. Returning a list allows to avoid `case`ing around in the calling code (if the list is empty, `[]` will be sent back to the client).

robertoaloi: This is not taking funs into account at this stage.

robertoaloi:
I saw your comments only after the merge @TheGeorge , sorry about that 🙏 
We can address a couple of them as follow ups.

apps/els_lsp/src/els_call_hierarchy_provider.erl
alanz: In your referenced comment you state zero or one.  When zero, we get a crash.

Issue #1098, opened 2021-09-30
Detect unused record fields


robertoaloi:
Inspired by the [Hank paper](https://arxiv.org/pdf/2107.08699.pdf), implement a new diagnostics backend which can detect unused record fields.

PR #1099, opened 2021-09-30
[#1098] Detect unused record fields

robertoaloi:
### Description

Screenshot 2021-09-30 at 08 16 39

### DIFF
--- a/apps/els_lsp/src/els_unused_record_fields_diagnostics.erl
+++ b/apps/els_lsp/src/els_unused_record_fields_diagnostics.erl
@@ -1,0 +1,74 @@
+%%==============================================================================
+%% Unused Includes diagnostics
+%%==============================================================================
+-module(els_unused_record_fields_diagnostics).
+
+%%==============================================================================
+%% Behaviours
+%%==============================================================================
+-behaviour(els_diagnostics).
+
+%%==============================================================================
+%% Exports
+%%==============================================================================
+-export([ is_default/0
+        , run/1
+        , source/0
+        ]).
+
+%%==============================================================================
+%% Includes
+%%==============================================================================
+-include("els_lsp.hrl").
+
+%%==============================================================================
+%% Callback Functions
+%%==============================================================================
+
+-spec is_default() -> boolean().
+is_default() ->
+  true.
+
+-spec run(uri()) -> [els_diagnostics:diagnostic()].
+run(Uri) ->
+  case filename:extension(Uri) of
+    <<".erl">> ->
+      case els_utils:lookup_document(Uri) of
+        {error, _Error} ->
+          [];
+        {ok, Document} ->
+          UnusedRecordFields = find_unused_record_fields(Document),
+          [make_diagnostic(POI) || POI <- UnusedRecordFields ]
+      end;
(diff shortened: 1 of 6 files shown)


apps/els_lsp/src/els_unused_record_fields_diagnostics.erl
```diff
@@ -0,0 +1,74 @@
...
+  UsagesIds = [Id || #{id := Id} <- Usages],
+  [POI || #{id := Id} = POI <- Definitions, not lists:member(Id, UsagesIds)].
+
+-spec make_diagnostic(poi()) -> els_diagnostics:diagnostic().
+make_diagnostic(#{id := POIId, range := POIRange}) ->
+  Range = els_protocol:range(POIRange),
+  FullName =
+    case POIId of
+      {RecordName, RecordField} ->
+      els_utils:to_binary(
```
gomoripeti, line 67: `els_utils:to_binary(` supports deep lists (`unicode:chardata()`) so no need to flatten. On the other hand should we use `~tp` instead of `~p` to properly print unicode atoms? (I am always confused with unicode, maybe a new `atom_to_label` function in `els_utils` could have it right once and that could be used in the whole code base)

apps/els_lsp/src/els_unused_record_fields_diagnostics.erl
```diff
@@ -0,0 +1,74 @@
...
+  <<"UnusedRecordFields">>.
+
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+-spec find_unused_record_fields(els_dt_document:item()) -> [poi()].
+find_unused_record_fields(Document) ->
+  Definitions = els_dt_document:pois(Document, [record_def_field]),
+  Usages = els_dt_document:pois(Document, [record_field]),
+  UsagesIds = [Id || #{id := Id} <- Usages],
```
gomoripeti, line 58: optimisation: UsagesIds could be an usort-ed list

gomoripeti (approved):

apps/els_lsp/src/els_unused_record_fields_diagnostics.erl
```diff
@@ -0,0 +1,74 @@
...
+
+-spec make_diagnostic(poi()) -> els_diagnostics:diagnostic().
+make_diagnostic(#{id := POIId, range := POIRange}) ->
+  Range = els_protocol:range(POIRange),
+  FullName =
+    case POIId of
+      {RecordName, RecordField} ->
+      els_utils:to_binary(
+        lists:flatten(io_lib:format("~p/~p", [RecordName, RecordField])));
+    Id -> atom_to_binary(Id, utf8)
```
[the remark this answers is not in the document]
robertoaloi: That's copy-paste from the Unused Macro diagnostics backend, cleaning.

apps/els_lsp/src/els_unused_record_fields_diagnostics.erl
robertoaloi: I like that approach. So, I will keep it for now and handle this as a follow up (feel free to contribute on that).

PR #1100, opened 2021-10-01
Eep 48 markdown

garazdawi:
### Description

This PR adds a markdown rendering backend for EEP-48 style documentation.

In VS code the change looks like this. 

Before:
before_els_markdown

After:
after_els_markdown

I'm opening this as a draft to get some feedback if this is something you want. If so, there is more work to be done in order to add tests and also check that is actually renders the markdown nicely in most of the editors.


robertoaloi:
Hi @garazdawi !

I guess improving the way documentation is displayed would be a nice thing to do. I just wonder why we'd need to fork the `shell_docs` module which is currently part of `stdlib`. Is the idea to iterate faster within the `erlang_ls` repo and eventually merge the changes back to OTP?

garazdawi:
> I guess improving the way documentation is displayed would be a nice thing to do.

You don't sound overly convinced?

> I just wonder why we'd need to fork the shell_docs module which is currently part of stdlib. Is the idea to iterate faster within the erlang_ls repo and eventually merge the changes back to OTP?

I choose to make a copy of the Erlang/OTP implementation for rendering in the shell in order to get some of the basic framework in place. I've redone much of the actual rendering logic as markdown and ansi shells are different in what they need.

I started out with the idea of creating a markdown renderer for EEP-48 style documentation in Erlang/OTP. But I think that in order to make the docs look nice, it will have to render different markdown depending on the editor, which leads me to think that this code belongs here instead.

If it turns out that I was wrong and no difference was needed for different editors, I'm not opposed to moving it into Erlang/OTP as I would rather maintain them both in one place than two separate versions.

robertoaloi:
> You don't sound overly convinced?

I am *very* convinced! I was just wondering if there was any reason for *not* wanting this, given you asked :)
So, go ahead! The plan makes sense to me.

garazdawi:
> I am very convinced! I was just wondering if there was any reason for not wanting this, given you asked :)
So, go ahead! The plan makes sense to me.

My main concern was that it is a lot of code and might increase maintenance while not providing enough benefit.

Happy to hear that you would like to have this. I'll spend more time on it and get something that can be merged soon(tm).

garazdawi:
Did some more work.

Added documentation lookup for types:
type_els_markdown

Started using edoc chunks (made possible by https://github.com/erlang/otp/pull/2803 in Erlang/OTP 24) for creating docs from edoc:

Before:
before_edoc_markdown

After:
after_edoc_markdown

I haven't added so that you get documentation on completion of types (yet). Any hints you have of what has to be changed in els_completion_provider to make it work would be most welcome.

I also noticed that Elixir [already has done this work](https://github.com/elixir-lsp/elixir_sense/pull/132), though I think my renderer is a bit more correct/complete. So maybe there is an opportunity to share some code in between the projects.

It would have been really neat if you could refer to other files or allow hovering on things within the returned markdown docs so that we could expand the type documentation within the function documentation if the user want that. But it seems like it is not possible.

robertoaloi:
@garazdawi Let us know when the PR is ready for review!

PR #1109, opened 2021-10-05
[#859] Ignore test data directories.

### LINKED ISSUE #859: indexing should skip test data directories
**Describe the bug**

Files in  `test/xxxx_SUITE_data` directory are indexed.

In the case of https://github.com/WhatsApp/erlfmt/tree/master/test/erlfmt_SUITE_data, this has a `unicode.erl` module which clashes with the system one.

dgud:
Test data directories may contain erlang files with modules names
that clashes with application or system modules, ignore these
directories.

As one example in otp "lib/dialyzer/test/r9c_SUITE_data/src/" contains
the many files that clashes. This causes among other things that
jump to definition almost always jumps to the wrong place when working
inside OTP.

Fixes #859

### DIFF
--- a/apps/els_core/src/els_config.erl
+++ b/apps/els_core/src/els_config.erl
@@ -283,7 +283,15 @@ project_paths(RootPath, Dirs, Recursive) ->
                                    )
             || Dir <- Dirs
           ],
-  lists:append(Paths).
+  case Recursive of
+    false ->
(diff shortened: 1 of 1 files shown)


robertoaloi (approved):

robertoaloi:
Thanks!

PR #1117, opened 2021-10-15
Provide a Gradualizer diagnostic

erszcz:
This implements a basic Gradualizer diagnostic for Erlang LS we, i.e. @garazdawi, Ferenc Pesti and I, built during an Erlang Solutions internal hackathon early this year. Some example screenshots of this integration in action can be seen here - https://github.com/josefs/Gradualizer/pull/344#issuecomment-927850685.

I'm creating this as a draft, since I'm not sure what's the best way to express the dependency on Gradualizer - currently even the newly added test won't pass, since I haven't added Gradualizer to Rebar3 deps. What's the best approach for adding dependencies for optional diagnostics?

Important: Gradualizer, while being continuously improved, is still a work in progress and doesn't successfully self-gradualize yet. It's already useful, but it might also be misleading at times. That being said, I think having the integration in place is worth the effort, as it makes trying out the project and contributing easier.

### DIFF
--- a/apps/els_lsp/src/els_gradualizer_diagnostics.erl
+++ b/apps/els_lsp/src/els_gradualizer_diagnostics.erl
@@ -23,12 +23,6 @@
 -include_lib("kernel/include/logger.hrl").
 
 %%==============================================================================
-%% Type Definitions
-%%==============================================================================
-%% -type macro_config() :: #{string() => string()}.
-%% -type macro_option() :: {atom()} | {atom(), any()}.
-
-%%==============================================================================
 %% Callback Functions
 %%==============================================================================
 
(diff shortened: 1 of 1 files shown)


robertoaloi:
Thanks for this, amazing stuff! I asked in our Slack channel what's the Community view on this. I will also take it with the other core maintainers and come back to you shortly :)

robertoaloi:
Also, I briefly tried the Gradualizer on the Erlang LS project itself and I noticed a couple of things:

* A number of warnings appear from Gradualizer about "undefined functions"
* Some messages seem truncated. For example:

```
{
      "source": "Gradualizer",
      "severity": 2,
      "range": {
        "start": {
          "line": 145,
          "character": 0
        },
        "end": {
          "line": 146,
          "character": 0
        }
      },
      "message": "The pattern #{config := #{entries := Entries, task := Task} = Config,"
    },
```

erszcz:
Thanks for the feedback!

> * A number of warnings appear from Gradualizer about "undefined functions"

I haven't seen this happening, but I'm not saying it's not - could you provide an example?

> * Some messages seem truncated. For example:

Good catch! Fixed now.

I've added the dependency on Gradualizer, but maybe it's worth waiting for https://github.com/josefs/Gradualizer/pull/358 before merging this one.

PR #1118, opened 2021-10-21
Add support for hover docs for types

plux:
### Description

When hovering over `type()` the definition of it will be shown.


gomoripeti:
I'm all for supporting older versions (I use OTP 21) but to me it looks like the other PR also addresses pre OTP 23 https://github.com/erlang-ls/erlang_ls/pull/1100/commits/0d7ab9da8401d0ac41af091d33910f28f64df450#diff-9ef24e9234297083ba850a2d2b6fa6e9fe142810e858ee3a07ab17fa1fca9306R105

hi @garazdawi, could you confirm if https://github.com/erlang-ls/erlang_ls/pull/1100 will support hover docs for types for older OTP version? and what do you think the timeline of that PR is, is it almost ready or it's worth to merge this one to have hover docs for types earlier?

PR #1122, opened 2021-10-27
Do not recurse in els_utils:fold_files

### LINKED ISSUE #859: indexing should skip test data directories

dgud:
Avoids double recursion, recursion is done in els_config.
This fixes #859 which my previous PR failed to do.

From what I can see els_util:fold_files is only used for indexing and input 
data are the directories from els_config which have already been recursed once.

### DIFF
--- a/apps/els_core/src/els_utils.erl
+++ b/apps/els_core/src/els_utils.erl
@@ -181,7 +181,7 @@ macro_string_to_term(Value) ->
       true
   end.
 
-%% @doc Folds over all files in a directory recursively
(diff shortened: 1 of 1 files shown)


robertoaloi:
I don't see any problems with this. @TheGeorge @gomoripeti FYI.

PR #1126, opened 2021-10-30
Remove fragile heuristics from editable_range

gomoripeti:


apps/els_lsp/src/els_parser.erl
```diff
@@ -283,20 +284,18 @@ attribute(Tree) ->
       case spec_function_name(FATree) of
         {F, A} ->
           [FTree, _] = erl_syntax:tuple_elements(FATree),
-          Anno = erl_syntax:get_pos(FTree),
-          %% FIXME this is weird:
-          %% starts at '-', ends at the end of callback function name
-          Start = get_start_location(Tree),
-          CallbackAnno = erl_anno:set_location(Start, Anno),
-          [poi(CallbackAnno, callback, {F, A})];
+          [poi(Pos, callback, {F, A},
```
gomoripeti, line 287: the range of the `callback` POI is now the full form (same way as for `type_definition` and `spec`)

robertoaloi:
Nice cleanup, thanks!

PR #1131, opened 2021-11-01
Remove TCP support

adrianroe:
The commit in
https://github.com/erlang-ls/erlang_ls/commit/d4a38eeb7317887c7860ffd43920219f22c2a969
removed TCP support but left some references to ranch which causes
issues in a nix build environment.


gomoripeti (approved):

robertoaloi:
Thanks!

PR #1132, opened 2021-11-09
Added the '-module' attribute snippet

Daylien:
### Description

I added the `-module(Module)` attribute into the available snippet list.
As for the placeholder I took `Module` because of the [Erlang docs](https://www.erlang.org/doc/reference_manual/modules.html#pre-defined-module-attributes).

Feedback is welcome, thanks!

### DIFF
--- a/apps/els_lsp/src/els_completion_provider.erl
+++ b/apps/els_lsp/src/els_completion_provider.erl
@@ -262,6 +262,7 @@ attributes() ->
   , snippet(attribute_import)
   , snippet(attribute_include)
   , snippet(attribute_include_lib)
+  , snippet(attribute_module)
   , snippet(attribute_on_load)
   , snippet(attribute_opaque)
   , snippet(attribute_record)
@@ -373,6 +374,8 @@ snippet(attribute_include) ->
   snippet(<<"-include().">>, <<"include(${1:}).">>);
 snippet(attribute_include_lib) ->
   snippet(<<"-include_lib().">>, <<"include_lib(${1:}).">>);
+snippet(attribute_module) ->
+  snippet(<<"-module().">>, <<"module(${1:Module}).">>);
 snippet(attribute_type) ->
   snippet(<<"-type name() :: definition.">>,
           <<"type ${1:name}() :: ${2:definition}.">>);
--- a/apps/els_lsp/src/els_completion_provider.erl
+++ b/apps/els_lsp/src/els_completion_provider.erl
@@ -554,7 +554,8 @@ exported_definitions(Module, POIKind, ExportFormat) ->
 
 -spec module_name(els_dt_document:item()) -> [binary()].
 module_name(#{uri := Uri}) ->
-  ModuleName = filename:basename(lists:last(filename:split(Uri)), ".erl"),
+  ModuleName = filename:basename(lists:last(filename:split(Uri)),
+  filename:extension(Uri)),
   PurifiedName = case binary:match(ModuleName, <<"-">>) of
     nomatch -> ModuleName;
     _ -> list_to_binary(
--- a/apps/els_lsp/src/els_completion_provider.erl
+++ b/apps/els_lsp/src/els_completion_provider.erl
@@ -201,6 +201,9 @@ find_completions( Prefix
     %% Check for "-anything"
     [{atom, _, _}, {'-', _}] ->
       attributes();
+    %% Check for "-module("
+    [{'(', _}, {atom, _, module}, {'-', _}] ->
+      [item_kind_file(Path) || Path <- module_name(Document)];
     %% Check for "-export(["
     [{'[', _}, {'(', _}, {atom, _, export}, {'-', _}] ->
       unexported_definitions(Document, function);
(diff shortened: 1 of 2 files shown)


robertoaloi:
Hi @Daylien and thanks for your contribution. The module name is supposed to coincide with the filename, so I wonder if we could default to that, instead. That's the behaviour that I already get in Emacs, for example. In other words, if I have open `roberto.erl`, the module name would default to `roberto`. We just need to be careful adding single quotes whenever the filename requires it (e.g. it contains a dash).

Daylien:
Hey @robertoaloi thanks for the answer! Yeah I though the same, I came up with the idea, to reuse some concepts from the behavior of https://github.com/erlang-ls/erlang_ls/blob/cd050eb506b784b94adfffa3edd20bc4962f678d/apps/els_lsp/src/els_completion_provider.erl#L275-L292 

It's more or less the same for the `-module(Module)` attribute, but only the relative path in which the file is created is relevant for the completion. What's your opinion on that?

I will dig myself more deeply into the code base. So stay tuend!

Daylien:
I added the basic auto-completion functionality. Feedback is really welcome. 

Thanks!

apps/els_lsp/test/els_completion_SUITE.erl
```diff
@@ -133,6 +133,11 @@ attributes(Config) ->
                 , kind => ?COMPLETION_ITEM_KIND_SNIPPET
                 , label => <<"-include().">>
                 }
+             , #{ insertText => <<"module(${1:}).">>
```
robertoaloi, line 136: Should we add a couple of testcases for this? E.g. one which requires quotes and one which doesn't.

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -546,6 +552,18 @@ exported_definitions(Module, POIKind, ExportFormat) ->
       []
   end.
 
+-spec module_name(els_dt_document:item()) -> [binary()].
+module_name(#{uri := Uri}) ->
+  ModuleName = filename:basename(lists:last(filename:split(Uri)),
+  filename:extension(Uri)),
+  PurifiedName = case binary:match(ModuleName, <<"-">>) of
```
robertoaloi, line 559: I believe it won't be enough to match on the `-` here. Quoting the [Erlang docs](https://www.erlang.org/doc/reference_manual/data_types.html#atom):

> An atom is to be enclosed in single quotes (') if it does not begin with a lower-case letter or if it contains other characters than alphanumeric characters, underscore (_), or @.

We could port [the regexp from the Emacs Erlang](https://github.com/erlang/otp/blob/master/lib/tools/emacs/erlang.el#L601) mode to Erlang LS:

```
(defconst erlang-atom-regular-regexp
    (if erlang-regexp-modern-p
        "\\_<[[:lower:]]\\(?:\\sw\\|\\s_\\)*\\_>"
      "\\<[[:lower:]]\\(?:\\sw\\|\\s_\\)*\\>")
    "Regexp describing a regular (non-quoted) atom")
```

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -546,6 +552,18 @@ exported_definitions(Module, POIKind, ExportFormat) ->
       []
   end.
 
+-spec module_name(els_dt_document:item()) -> [binary()].
+module_name(#{uri := Uri}) ->
+  ModuleName = filename:basename(lists:last(filename:split(Uri)),
```
robertoaloi, line 557: The `els_uri` module offers a conversion function from _Uri_ to _Module_. Since in this case we don't need the conversion to atom part, I would recommend modifying that function with something like (untested):

```
module(Uri) ->
  module(Uri, to_atom).

module(Uri, to_atom) ->
  binary_to_atom(module(Uri, to_binary), utf8);
module(Uri, to_binary) ->
  filename:basename(path(Uri), <<".erl">>).
```

robertoaloi (changes requested):
Sorry about the delay in the review, I left some comments. I hope they make sense :)

Daylien:
> Sorry about the delay in the review, I left some comments. I hope they make sense :)

Thanks for the good feedback!

I will work your mentioned changes in after the holidays. Beside that, I wish you a merry Christmas and a happy new year.
Stay tuned for my changes in early 2022.

Daylien:
Hey @robertoaloi!

I continued working on this and I created a little regex match to match against "invalid" chars. All is working fine beside the special characters problem `@!$` and so on (but `-` is working). So far so good.

But here is the big deal. The problem also affects the `-include(...)` and `include_lib(...)` completion trigger. 
To test it, create a file which is called `my@header.hrl`. On triggering the autocompletion provider you will notice that the `@` will be resolved into Latin1 hex so the name is now `my%40header.hrl` and this is cracking my work also up. I will push my code when I know why the `path` is in hex Latin-1 and not in the "normal" Latin-1 format.

Also I discovered the following at https://www.erlang.org/doc/apps/stdlib/unicode_usage.html.
> In Erlang/OTP 20.0, atoms and function can contain Unicode characters. Module names, application names, and node names are still restricted to the ISO Latin-1 range.

Latin-1 is fine for module names but here we talk about filenames which should be treated as UTF-8 AFAIK (please correct me if I'm wrong).

My guess is that the problem lives in `els_utils.erl` or more precise in the `path(Uri)` function.

Maybe you have an idea why this behavior is what is it currently.

PR #1137, opened 2021-11-22
RefactorErl Diagnostics

robertfiko:
### Description

As you (@robertoaloi) asked, I cleaned the code and created this PR with implementation for unused macro and unsecure call detection with RefactorErl backed.


apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
```diff
@@ -0,0 +1,137 @@
...
+run(Uri, RecursionDepth) when RecursionDepth < ?MAX_RECURSION_DEPTH ->
+  case filename:extension(Uri) of
+    <<".erl">> -> 
+      case referl_node() of
+        disabled ->
+          [];
+        Node -> 
+          case rpc:call(Node, ri, add, [binary_to_list(els_uri:path(Uri))], ?TIME_OUT) of
+            {badrpc, _} ->
+              els_server:send_notification(<<"window/showMessage">>, #{ type => ?MESSAGE_TYPE_ERROR, message => <<"Refactor Erl node is down {badrpc}!">> }),
```
gomoripeti, line 49: @robertoaloi would it make sense to add shortcuts to client notifications like `els_utils:show_msg(info | warning | error, Fmt :: string | binary [, Args :: list()])`

apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
```diff
@@ -0,0 +1,137 @@
...
+    Range = els_protocol:range(PoiRange),
+    IssueName = list_to_binary(PoiData), % TODO Robi: the ID and the arity should be encoded to the ID
+    Message = <<"Security Issue: ", IssueName/binary>>,
+    Severity = ?DIAGNOSTIC_WARNING,
+    Source = source(),
+    els_diagnostics:make_diagnostic(Range, Message, Severity, Source).
+
+  
+-spec referl_node() -> atom(). % If called with no args, then try to get it from config
+referl_node() ->
```
gomoripeti, line 102: looks like some common referl functionality could be moved to a library module like `els_referl_util`, eg `referl_node/0/1`, error handling of rpc call to the referl node, processing common referl responses like `[{Option, Tuple, Atom, DataList}]` with a fun etc

gomoripeti (commented):
nice initiative

apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
robertoaloi: Sure, but we can probably take care of that outside of this PR.

apps/els_core/src/els_config.erl
```diff
@@ -55,7 +55,8 @@
                | elvis_config_path
                | indexing_enabled
                | bsp_enabled
-               | compiler_telemetry_enabled.
+               | compiler_telemetry_enabled
+               | refactorerl.
```
robertoaloi, line 59: Considering that someone else is trying a similar approach with Wrangler (at least so I've heard), I wonder if the config should be a bit more generic, as in:

```
external_refactoring_tool:
  name: refactorerl | wrangler
```

Or similar.

apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
robertfiko: @gomoripeti I just started an els_referl module for the "common parts" should be finished by the end of the week. Also cleaned out refactor erl node detectiona and error management.

apps/els_lsp/src/els_diagnostics.erl
```diff
@@ -64,6 +64,8 @@ available_diagnostics() ->
   , <<"elvis">>
   , <<"unused_includes">>
   , <<"unused_macros">>
+  , <<"referl_unused_macros">>
```
robertoaloi, line 67: In my mind, it could make more sense to have `<<"refactorerl">>` as a single diagnostics backend, instead of one per feature. This would also help removing some of the existing duplication between the two newly introduced modules. What do you think?

apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
```diff
@@ -0,0 +1,137 @@
...
+            {badrpc, _} ->
+              els_server:send_notification(<<"window/showMessage">>, #{ type => ?MESSAGE_TYPE_ERROR, message => <<"Refactor Erl node is down {badrpc}!">> }),
+              [];
+            error ->
+              timer:sleep(1000),
+              run(Uri, RecursionDepth + 1);
+            _ ->
+              ModuleName = filename:rootname(filename:basename(binary_to_list(els_uri:path(Uri)))),
+              ReferlResult = rpc:call(Node, refusr_sq, run, [[{positions, linecol}, {output, msg}], [], "mods[name=" ++ ModuleName ++ "].funs.unsecure_calls"]),
+              Pois = convertToPoi(ReferlResult),
```
robertoaloi, line 57: Nit: we are not using camel case for function names.

apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
gomoripeti: awesome

apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
```diff
@@ -0,0 +1,137 @@
...
+    <<".erl">> -> 
+      case referl_node() of
+        disabled ->
+          [];
+        Node -> 
+          case rpc:call(Node, ri, add, [binary_to_list(els_uri:path(Uri))], ?TIME_OUT) of
+            {badrpc, _} ->
+              els_server:send_notification(<<"window/showMessage">>, #{ type => ?MESSAGE_TYPE_ERROR, message => <<"Refactor Erl node is down {badrpc}!">> }),
+              [];
+            error ->
```
robertoaloi, line 51: What's the scenario where a retry can actually help? Also, not sure if we want to really pop a message every second. Maybe use logs instead?

apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
gomoripeti: of course 👍

apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
```diff
@@ -0,0 +1,137 @@
...
+              timer:sleep(1000),
+              run(Uri, RecursionDepth + 1);
+            _ ->
+              ModuleName = filename:rootname(filename:basename(binary_to_list(els_uri:path(Uri)))),
+              ReferlResult = rpc:call(Node, refusr_sq, run, [[{positions, linecol}, {output, msg}], [], "mods[name=" ++ ModuleName ++ "].funs.unsecure_calls"]),
+              Pois = convertToPoi(ReferlResult),
+              [make_diagnostic(Poi) || Poi <- Pois]
```
robertoaloi, line 58: What's the reason to convert the RefactorErl output to a POI first and then to a diagnostic?

apps/els_lsp/src/els_referl_unsecure_calls_diagnostics.erl
robertoaloi: I agree. It would be nice to decouple the diagnostic backend itself from a module which deals with interfacing with RefactorErl. These could be a single `els_refactorerl_diagnostics.erl` (which covers all, configurable, diagnostic types) and `els_refactorerl.erl`

robertoaloi:
Definitely a nice addition @robertfiko , thanks! I left a few comments here and there, but from my side the main thing would be to:

* Use a single `els_refactorerl_diagnostics` backend which covers all diagnostics returned by RefactorErl (which should be configurable)
* Introduce a separate `els_refactorerl` module which contains all the logic to interact with the external RefactorErl node

Before this can be merged, we would also need to:

* Add tests (probably mocking RefactorErl)
* Add an explanation to the user on how to install/configure RefactorErl

The real challenge with RefactorErl, as previously mentioned by @alanz , will be to find a way to provide input to the tool, since this is not really supported by the LSP protocol. We may propose an extension, but that would require custom implementations for all clients. Any suggestion in that sense is more than welcome.

robertfiko:
> Definitely a nice addition @robertfiko , thanks! I left a few comments here and there, but from my side the main thing would be to:
> 
> * Use a single `els_refactorerl_diagnostics` backend which covers all diagnostics returned by RefactorErl (which should be configurable)
> * Introduce a separate `els_refactorerl` module which contains all the logic to interact with the external RefactorErl node
> 
> Before this can be merged, we would also need to:
> 
> * Add tests (probably mocking RefactorErl)
> * Add an explanation to the user on how to install/configure RefactorErl
> 
> The real challenge with RefactorErl, as previously mentioned by @alanz , will be to find a way to provide input to the tool, since this is not really supported by the LSP protocol. We may propose an extension, but that would require custom implementations for all clients. Any suggestion in that sense is more than welcome.

Thank you for your tougths and suggesrtions @robertoaloi!

- I will reorded and repurpose the modules as I wrote in a comment.
- I will write a documentation page for everything that is needed for the setup. As I can see, I can also contribute to the website, so I will write it there. (and will expand the docs when more functionalites will be added)
- I will look into how to add tests to theese functionalities
- A done a quick search on the "input" topic, I found something in the MS LSP Doc: https://microsoft.github.io/language-server-protocol/specification#workspace_executeCommand I do not know if that can work, but for now I will clear out things related to this two features, as this will define the interface. After it I will look deeper into the topic.

Thank you for everything, I will add some new commits to the PR in a few days.

robertfiko:
### Main changes

I hope I covered everything, that we talked about, if not, let me know!

- Documented the code the get more understandable, for you, for me, for everyone
- Merged the logic behind the two diagnostics (unsecured calls and unused macros)
- The RefactorErl node can be configured in the config file, like this:

```
refactorerl:
  node: "name@host"
``` 
 - Also, the diagnostics which the user want to use can be added here, instead of burning it into the source code. Kind of like you are doing currently with diagnostics!
 - Merged the make diagnostic and convert to poi logics into one, so we don't iterate twice.

I hope covered all major changes.

apps/els_lsp/src/els_refactorerl_diagnostics.erl
```diff
@@ -0,0 +1,131 @@
...
+run(Uri) ->
+  run(Uri, 0).
+
+
+-spec run(uri(), number()) -> [els_diagnostics:diagnostic()].
+run(Uri, RecursionDepth) when RecursionDepth < ?MAX_RECURSION_DEPTH ->
+  case filename:extension(Uri) of
+    <<".erl">> ->
+      case els_refactorerl_utils:referl_node() of
+        disabled ->
```
robertoaloi, line 66: The function returns `{ok, _}` or `{error, _}`, so this will never match.

apps/els_lsp/src/els_refactorerl_diagnostics.erl
```diff
@@ -0,0 +1,131 @@
...
+          [];
+        _ ->
```
robertoaloi, line 68: I would be specific on the matches, here.

apps/els_lsp/src/els_refactorerl_diagnostics.erl
```diff
@@ -0,0 +1,131 @@
...
+  ].
+
+  %%@doc
+  %% Adds a module to the RefactorErl node.
+-spec add(any()) -> atom(). %TODO: Add .hrl files
+add(Uri) ->
+  Param = [binary_to_list(els_uri:path(Uri))],
+  case els_refactorerl_utils:referl_node() of
+    {ok, Node} ->
+      rpc:call(Node, ri, add, Param, els_refactorerl_utils:maxtimeout());
```
robertoaloi, line 128: What does the `ri:add` function returns? Do we want to handle errors?

apps/els_lsp/src/els_refactorerl_diagnostics.erl
```diff
@@ -0,0 +1,131 @@
...
+      , "Unused Macros:"
+      , "mods[name="
+      , "].macros[not .references]" }
+  ].
+
+  %%@doc
+  %% Adds a module to the RefactorErl node.
+-spec add(any()) -> atom(). %TODO: Add .hrl files
+add(Uri) ->
+  Param = [binary_to_list(els_uri:path(Uri))],
```
robertoaloi, line 125: Nit: This should probably be `Params`.

apps/els_lsp/src/els_refactorerl_diagnostics.erl
```diff
@@ -0,0 +1,131 @@
...
+  [ {unused_calls, "Security Issue",  "mods[name=", "].funs.unsecure_calls"}
+  ,   {unsecure_macros
+      , "Unused Macros:"
+      , "mods[name="
+      , "].macros[not .references]" }
+  ].
+
+  %%@doc
+  %% Adds a module to the RefactorErl node.
+-spec add(any()) -> atom(). %TODO: Add .hrl files
```
robertoaloi, line 123: `any()` should probably be `uri()`. What does the `atom()` represent?

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,197 @@
...
+      {error, other}
+  end.
+
+%%@doc
+%% Tries to connect to a node
+%% When statues is validate, it reports the success, and failure as well,
+%% when retry, it won't report.
+-spec try_connect_node({validate | retry, atom()}) -> {error, disconnected}
+                                                     | atom().
+try_connect_node({Status, Node}) ->
```
robertoaloi, line 163: A lot of the code here is related to the handling of both transient and permanent failures for the RefactorErl node. As a first iteration, I would suggest we just assume the node is up and running and log eventual RPC failures.
After all, a RPC error is probably due to a misconfiguration and will most likely require human intervention. What do you think? That should significantly simplify things.

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,197 @@
...
+%%@doc
+%% Calls the RefactorErl node, to check if it's alive using ri:ls()
+-spec check_node(atom()) -> {error, timeout}
+                          | {error, badrpc}
+                          | {error, other}
+                          | atom().
+check_node(Node) ->
+  Response = rpc:call(Node, ri, ls, [], 10000),
+  case Response of
+    {{ok, _}, {error, _}} ->
```
robertoaloi, line 145: What kind of errors do we expect here?

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,197 @@
...
+case ReferlResult of
+    [{_, _, _, DataList}] ->
+      convert_to_poi(DataList);
+    [{{_, {FromLine, FromCol}, {ToLine, ToCol}}, Name} | Tail] ->
+      Range = #{ from => {FromLine, FromCol}, to => {ToLine, ToCol} },
+      Id = refactorerl_poi, %"{module(), 'atom()', 'arity()''}",
+      Poi = els_poi:new(Range, application, Id, Name),
+      [ Poi | convert_to_poi(Tail) ];
+    _ ->
+        [] %TODO Robi notify
```
robertoaloi, line 132: Maybe write a log message here?

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,197 @@
...
+      end
+  end.
+
+-spec convert_to_poi(any()) -> poi().
+%%@doc
+%% Convert a RefactorErl result to a POI, for the LS system
+%% The RefactorErl format is how the refusr_sq:run\3 returns
+convert_to_poi(ReferlResult) ->
+case ReferlResult of
+    [{_, _, _, DataList}] ->
```
robertoaloi, line 124: Given that this function takes a generic `any()` as an input (which is fine, since the type is defined outside of the Erlang LS codebase), it's probably wise to name these variables, to indicate what they mean (e.g. `_Message` vs `_`).

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,197 @@
...
+        {badrpc, _} ->
+          disable_node(Node); % Returns disabled
+        error ->
+          [];
+        _ ->
+          Response
+      end
+  end.
+
+-spec convert_to_poi(any()) -> poi().
```
robertoaloi, line 118: Given the type is defined outside of the Erlang LS code base, I would probably define an alias and add a comment about the expected type:

```
%% For the type definition, please refer to [LINK HERE]
%% -type refactorerl_result() :: [{..., ...}] | ...

-type refactorerl_result() :: any().
```

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,197 @@
...
+%% Convert a RefactorErl result to a POI, for the LS system
+%% The RefactorErl format is how the refusr_sq:run\3 returns
+convert_to_poi(ReferlResult) ->
+case ReferlResult of
+    [{_, _, _, DataList}] ->
+      convert_to_poi(DataList);
+    [{{_, {FromLine, FromCol}, {ToLine, ToCol}}, Name} | Tail] ->
+      Range = #{ from => {FromLine, FromCol}, to => {ToLine, ToCol} },
+      Id = refactorerl_poi, %"{module(), 'atom()', 'arity()''}",
+      Poi = els_poi:new(Range, application, Id, Name),
```
robertoaloi, line 129: Any reason why we generate an intermediate POI here? Why not to generate diagnostics directly from RefactorErl results?

robertoaloi (changes requested):
Nice progress!

I have a couple of suggestions so that we can merge this one in and then iterate.

* The extension is somewhat experimental, so it should be off by default
* We need some basic tests for this. Those tests could be mocking the RPC
* The intermediate POI generation seems un-necessary and could be removed
* I would not include error handling as a start and assume the code is up and running when the configuration is present. That should simplify this PR significantly.

apps/els_lsp/src/els_refactorerl_diagnostics.erl
robertfiko: This is changed a bit, as the request to the RefactorErl node was not so stable. I will describe this in a comment, when I merge to this PR.

apps/els_lsp/src/els_refactorerl_utils.erl
robertfiko: Yes, that's a great idea I also taught about something similar. I'll check my possibilities on that one

apps/els_lsp/src/els_refactorerl_utils.erl
robertfiko: That is also changed a bit, as the request to the RefactorErl node was not so stable. I will describe this in a comment when I merge to this PR. But this indication is great idea, I added to the code where I think it needs clarification, if you need it more, please let me know

apps/els_lsp/src/els_refactorerl_utils.erl
robertfiko: Yes I also merged this, it will come soon to this PR ;-)

apps/els_lsp/src/els_refactorerl_utils.erl
robertfiko: This is also changed, yet if you still think it would be nice to write logs somewhere, let me know

apps/els_lsp/src/els_refactorerl_utils.erl
robertfiko: Yes, it will need human intervention, but if everything is fine only the first two cases will match, but I would like to keep the others there, to make it a bit more error "resistant", also it helps me a lot during development. So I would like to leave it there, but I do want to re-think in the future when new features will be added, etc...

robertfiko (commented):
Hi!

Thanks for the great ideas, thoughts, typos and other hints!
Sorry I change the code that much during reviews, but it was needed, let me explain:
- We found out that the connection between ELS and RefactorErl is not so stable, because it becomes unresponsive as the first approach on the communication sent requests to Referl in a way, it manipulated its database, which is let's just say not "concurrent-friendly". Now that is resolved, by calling different module of Referl (which is after the concurrent handling layer, not before) 
- The response will come back as a message, so a few recieves added.
- `convert_to_pot` and `make_diagnostic` were merged and moved to the utils module
- The source() was also moved the utils module, because the make_diagnositc(s) was moved there also, and might more maintainable in the future

I think I coverd all the changes, and requests you made, if not please let me know.

Thanks for everything

robertfiko:
I know I left out the tests. If you think the source code is fine, I will write some!

apps/els_core/src/els_config.erl
robertoaloi: The proposal looks good to me.

apps/els_lsp/src/els_refactorerl_diagnostics.erl
```diff
@@ -0,0 +1,94 @@
...
+    <<".erl">> ->
+      case els_refactorerl_utils:referl_node() of
+        {error, _} ->
+          [];
+        {ok, _} ->
+          case els_refactorerl_utils:add(Uri) of
+            error ->
+              [];
+            ok ->
+              FileName = filename:basename(binary_to_list(els_uri:path(Uri))),
```
robertoaloi, line 50: You should be able to use `Module = els_uri:module(Uri)` here.

apps/els_lsp/src/els_refactorerl_diagnostics.erl
```diff
@@ -0,0 +1,94 @@
...
+              Module = list_to_atom(filename:rootname(FileName)),
+              Diagnostics = refactorerl_diagnostics(),
```
robertoaloi, line 52: Nit: the name is a bit misleading here: it suggests the function is returning a list of diagnostics, but in reality it returns a list of _identifiers_ (which actually looks more like _specs_, of the form `{id(), description(), query()}`).

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,284 @@
...
+      {error, disconnected};
+    {_, Node} ->
+      notification("RefactorErl is connected!", ?MESSAGE_TYPE_INFO),
+      els_config:set(refactorerl, #{"node" => {Node, validated}}),
+      {ok, Node}
+  end.
+
+%%@doc
+%% Gets a request id from the RefactorErl node, and returns it 
+-spec request_id() -> nodedown | {reqid | string()}.
```
robertoaloi, line 265: Nit: For consistency I would probably have all these functions returning either an `ok` or an `error` tuple`. For example, here it would be `{ok, RequestId}` or `{error, Reason}`. I am not sure about the current signature, is the atom `reqid` occasionally returned instead of the string?

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,284 @@
...
+%%@doc
+%% Creates a list of diagnostics form the RefactorErl results and a message
+%% Message can be the same, as this is called per diagnsotic type.
+%% The severity is only warning.
+-spec make_diagnostics(any(), string()) -> [els_diagnostics:diagnostic()].
+make_diagnostics([{{_Path, From, To}, Name} | Tail], DiagMsg) ->
+  Range = #{ from => From, to => To },
+  Id = refactorerl_poi,
+  #{ data := PoiData, range := PoiRange} =
+                        els_poi:new(Range, application, Id, Name),
```
robertoaloi, line 189: I still think we shouldn't create a POI here. An application POI is supposed to be a function application (a call) and the ID is supposed to be the MFA or FA of the function. Here we are only interested in the range, which we already have, but in a slightly different form.

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,284 @@
...
+        {badrpc, _} ->
+          disable_node(Node),
+          error;
+        ok ->
+          receive
+            {ReqID, reply, {ok, _}} -> ok
+          end;
+        deny ->
+          notification("Adding is deined, retry!"), % TODO: Take this out overtime stability proves itself
+          timer:sleep(1000),
```
robertoaloi, line 109: I would really avoid retries. 1 second during editing is an extremely significant delay and the user experience would be seriously degraded. We want to keep the Erlang LS side of the integration as simple as possible (we will have to maintain it, after all), so every other instability issue should be solved on the RefactorErl side in my mind. In case of any error, the correct thing to do in my mind is not to report any diagnostics (`[]`). So, to be able to merge this, I would really encourage you to take away all the error handling, at least until we get the basic integration up and running.

robertoaloi (changes requested):
Nice progress. I still think the error handling should be eventually introduced later (see my comments). Also, we shouldn't need to generate a new POI while generating diagnostics. Other than that and the tests, I think we are very close. 

On a side note, you seem to have some CI issues. I suspect that, if you enable GitHub actions on your fork, you should be able to run CI as well.

apps/els_lsp/src/els_refactorerl_diagnostics.erl
robertfiko: I also made it configurable. Maybe `enabled_diagnositcs()` can work out?

apps/els_lsp/src/els_refactorerl_utils.erl
robertfiko: No, it either returns `'nodown'` or `{reqid | string()}`. If I wrap out the request ID string out from the tuple, the place where I call it, than I need to re-wrap it, as RefactorErl needs it in this tupled form.

A suggestion of mine is to lets go with: `{error, Reason}` or `{ok, {reqid, RequestID}}` or stay with the current implementation. What do you think?

apps/els_lsp/src/els_refactorerl_utils.erl
robertfiko: Okay, removed the POI creation.

robertfiko:
Hopefully, I included everything, you asked for. If not please let me know, I will! ;-) 
I hope, CI will pass, if you give your approval on the code, I will complete the tests, and push those as well.

robertfiko:
Tests & Logic moved to RefactorErl side

I added tests that cover the communication / interace module for RefactorErl
Most of the logic was moved outside ELS to RefactorErl
The remaining logic is well documented in the code
Some thing to note:

The current release of RefactorErl cannot communicate with ELS as, the interface module is not present, but it will be in the next release

apps/els_lsp/test/els_diagnostics_SUITE.erl
```diff
@@ -818,3 +857,29 @@ src_path(Module) ->
...
+      ),
+  meck:expect(els_refactorerl_utils, referl_node, 0, {ok, NodeName}),
+  meck:expect(els_refactorerl_utils, add, 1, ok),
+  meck:expect(els_refactorerl_utils, source_name, 0, <<"RefactorErl">>),
+
+  meck:new(els_refactorerl_diagnostics, [passthrough, no_link, unstick]),
+  meck:expect(els_refactorerl_diagnostics, is_default, 0, true).
+
+
+unmock_refactoerl() ->
```
robertoaloi, line 883: Typo: refactorerl

apps/els_lsp/test/els_diagnostics_SUITE.erl
```diff
@@ -818,3 +857,29 @@ src_path(Module) ->
 
 include_path(Header) ->
   filename:join(["code_navigation", "include", Header]).
+
```
robertoaloi, line 860: These could be moved to a separate module.

apps/els_lsp/src/els_refactorerl_diagnostics.erl
```diff
@@ -0,0 +1,90 @@
...
+%%==============================================================================
+%% Internal Functions
+%%==============================================================================
+% @doc
+% Returns the enabled diagnostics by merging default and configed
+-spec enabled_diagnostics() -> [refactorerl_diagnostic_alias()].
+enabled_diagnostics() ->
+  case els_config:get(refactorerl) of
+    #{"diagnostics" := List} ->
+      [list_to_atom(Element) || Element <- List];
```
robertoaloi, line 74: Nit: It would be good to avoid the `list_to_atom` here.

apps/els_lsp/src/els_refactorerl_utils.erl
```diff
@@ -0,0 +1,160 @@
...
+
+%%@doc
+%% Runs list of diagnostic aliases on refactorerl
+-spec run_diagnostics(list(), atom()) -> list().
+run_diagnostics(DiagnosticAliases, Module) ->
+  case els_refactorerl_utils:referl_node() of
+    {ok, Node} ->
+      %% returns error | ok
+      rpc:call(Node, referl_els, run_diagnostics, [DiagnosticAliases, Module]);
+    _ -> % In this case there was probably error.
```
robertoaloi, line 100: You probably want to log a message here.

robertoaloi (approved):
Left a couple of comments, but nothing blocking. If CI passes, let's get this one in!

robertoaloi:
Thanks!

PR #1141, opened 2021-11-24
[1140] Do not include error description in code

### LINKED ISSUE #1140: Do not assume reason is a string when extracting code
According to the [Erlang/OTP documentation](https://www.erlang.org/doc/man/erl_parse.html#type-error_description), error descriptions are not guaranteed to be strings, but they can be any Erlang term.

The `els_compiler_diagnostics:make_code/2` function erroneously assumes that the description is a string. This is not true, for
example, when it comes to the `parse_trans` application, widely adopted in the Erlang community, which can return arbitrary error descriptions.

Assuming the description is a string (`~s`) causes the compiler diagnostics backend to crash when the description is not a string, so that no compilation diagnostics are displayed at all.

Furthermore, since the description can be arbitrary complex (in the case of `parse_trans` it contains a stack trace`, it shouldn't be included in the code part.

robertoaloi:
### Description

Fixes #1140 .

### DIFF
--- a/apps/els_lsp/priv/code_navigation/src/diagnostics_parse_transform_error.erl
+++ b/apps/els_lsp/priv/code_navigation/src/diagnostics_parse_transform_error.erl
@@ -1,0 +1,8 @@
+-module(diagnostics_parse_transform_error).
+
+-compile({parse_transform, my_parse_transform}).
+
+-export([ function_a/0 ]).
+
+function_a() ->
+  ok.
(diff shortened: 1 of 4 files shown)


michalmuskala:
Hmm... I think this breaks our integration with `erlang_ls` where we rely on this extended information about errors.

Perhaps there's a way to preserve the extra information for "standard" errors, while also not crash for custom ones?

I could think of something like:
```erlang
make_code(Module, Reason) when is_atom(Reason) ->
  unicode:characters_to_binary(io_lib:format("~ts:~ts", [Module, Reason]));
make_code(Module, Reason) when is_atom(element(1, Reason)) ->
  unicode:characters_to_binary(io_lib:format("~ts:~ts", [Module, element(1, Reason)]));
make_code(Module, _Reason) ->
  unicode:characters_to_binary(io_lib:format("~ts", [Module])).
```

What do you think?

robertoaloi:
Hi @michalmuskala I don't have a problem with that. Willing to submit a PR?

Issue #1037, opened 2021-06-09
Support some features in incomplete code


gomoripeti:
**Is your feature request related to a problem? Please describe.**
Incomplete code is currently unparsable, therefor most of the erlang_ls features don't work there. OTOH as Roberto put it "that's where developers spend most of their time on when writing code". We have to identify which features make sense and the most useful when editing code. Also we should investigate which features could be supported and how. This issue is spawned from #1035 , let's discuss in the comments.

**Features and feasibility**
It is els_parser that creates POIs from source code. Many other features work on the POIs. So this part of the task is to create POIs from incomplete code and those features don't have to be modified.

Please suggest what are the other features that need additional support apart from POIs from incomplete code. I suspect some diagnostics are like that.

Btw is there a full list of features of erlang_ls that we can go through?
(collected some from module names in els_lsp)
- [ ] (lens) function references
- [ ] (lens) behaviour usages
- [ ] (lens) suggest spec (N/A)
- [ ] (diagnostics) bound var in pattern
- [ ] (diagnostics) run ct (N/A)
- [ ] (diagnostcs) compiler (N/A)
- [ ] (diagnostics) cross-ref (is this applicable?)
- [ ] (diagnostics) dialyzer (N/A)
- [ ] (diagnostics) elvis(N/A)
- [ ] (diagnostics) unused includes (is this applicable?)
- [ ] (diagnostics) unused macros (is this applicable?)
- [ ] completions (lot of sub features) (already works on tokens, the list of completions and context is gathered from POIs though)
- [ ] goto definition (lot of sub features)
- [ ] hover/docs (sub features)
- [ ] document highlight
- [ ] document symbol
- [ ] folding range
- [ ] formatting (no formatter I know of supports incomplete code, incomplete code indentation is a popular request in the ecosystem)
- [ ] goto implementation
- [ ] references (lot of sub features)
- [ ] rename
- [ ] workspace symbol

**Implementation details and alternatives**
- erlang_ls already looks at the scanned tokens to find certain attributes (export, export_type, spec). This could be extended to figure out as many POIs as possible. OTOH erlang_ls should not reinvent the parser. Basic ones (ongoing): variables, atoms maybe macros. Function calls and records are harder. Fortunately erlang_ls does not care about control structures like case or try-catch which would be the hardest to parse.
- Alternatively it would be cool to have a re-entrant parser with a continuation that is not opaque, but I don't know of any for Erlang.
[...]

PR #1146, opened 2021-11-27
Parse incomplete text

gomoripeti:
In case of unparsable forms, try removing some trailing tokens, this way the
first half of the form could be parsed and POIs extracted.

related to #1037


robertoaloi:
@michalmuskala Would it be possible to review/merge/release https://github.com/WhatsApp/erlfmt/pull/325 ? Then we wouldn't need to pin the fork for `erlfmt`.

apps/els_lsp/src/els_parser.erl
```diff
@@ -49,6 +50,18 @@ forms_to_ast({ok, Forms, _ErrorInfo}) ->
 forms_to_ast({error, _ErrorInfo} = Error) ->
   Error.
 
+-spec parse_incomplete_text(string(), {erl_anno:line(), erl_anno:column()})
```
robertoaloi, line 53: Isn't this unused?

apps/els_lsp/src/els_parser.erl
```diff
@@ -66,46 +79,97 @@ parse_forms(Forms) ->
...
+    {error, {ErrorLoc, erlfmt_parse, _Reason}} ->
+      TrimmedTokens = tokens_until(Tokens, ErrorLoc),
+      parse_incomplete_tokens(TrimmedTokens)
+  end.
+
+%% @doc Drop tokens after given location but keep final dot, to preserve its
+%% location
+-spec tokens_until([erlfmt_scan:token()], erl_anno:location())
+                  -> [erlfmt_scan:token()].
+tokens_until([_Hd, {dot, _} = Dot], _Loc) ->
```
robertoaloi, line 119: Are we sure this cannot be invoked with `[]`?

robertoaloi:
This is amazing! It should drastically improve ergonomics while writing new code.

apps/els_lsp/src/els_parser.erl
gomoripeti: its only used for manual testing (similar to parse_text and parse_file) I found them useful while developing. I should add a comment to clarify.

apps/els_lsp/src/els_parser.erl
gomoripeti: dialyzer was sure :). it's because ensure_dot will only return a non empty list of tokens

apps/els_lsp/src/els_parser.erl
```diff
@@ -66,46 +79,97 @@ parse_forms(Forms) ->
...
+  %% We need to drop at least one token before the dot.
+  %% Otherwise if error location is at the dot, we cannot just drop the dot and
+  %% add a dot again, because it would result in an infinite loop.
+  [Dot];
+tokens_until([Hd | Tail], Loc) ->
+  case erlfmt_scan:get_anno(location, Hd) < Loc of
+    true ->
+      [Hd | tokens_until(Tail, Loc)];
```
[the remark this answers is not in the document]
gomoripeti: as far as I know there is no major perf advantage of tail recursion [first myth](https://www.erlang.org/doc/efficiency_guide/myths.html#myth--tail-recursive-functions-are-much-faster-----than-recursive-functions). This hand rolled version of lists:filter allows to keep the last dot token.

rebar.config
```diff
@@ -12,7 +12,8 @@
        , {docsh,         "0.7.2"}
        , {elvis_core,    "1.1.1"}
        , {rebar3_format, "0.8.2"}
-       , {erlfmt,        "1.0.0"}
+       %%, {erlfmt,        "1.0.0"}
+       , {erlfmt, {git, "https://github.com/gomoripeti/erlfmt.git", {branch, "parser_error_loc"}}} %% Temp until PR is merged
```
gomoripeti, line 16: agree, this PR should only be merged after the erlfmt PR

apps/els_lsp/src/els_parser.erl
gomoripeti: btw I did not test cases which cannot even be scanned, the error handling might be lacking in those cases. will look at that.

gomoripeti: looked at scanning errors:
if text cannot be tokenized then `erlfmt:read_nodes_string/2` returns an error (the error from `erl_scan`) (no forms at all). The new code only handles `raw_string` quasi-forms, which are only returned by `erlfmt` when scanning was successful but parsing was not for some forms. So in this part of the code it is not necessary to handle text which cannot be tokenized.

It would be possible to trim the text at the scanning error position, but that's a different code path and Im not sure how useful it is
- a scanning error because of a not yet closed quote (string or atom): no completion needed within a quote
- other scanning error which is not incomplete code like bad integer `16#g`: compiler diagnostics warns about these, probably best to expect the user to fix these before proceeding (and expecting ELS to work properly) this is more a tree-sitter type of work to fix heuristically.

apps/els_lsp/src/els_parser.erl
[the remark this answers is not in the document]
robertoaloi: Old habits are hard to die I guess :)

apps/els_lsp/src/els_parser.erl
robertoaloi: I see. Fair enough, let's add a comment to clarify.

apps/els_lsp/src/els_parser.erl
robertoaloi: To me the current approach seems good enough for now. Should we look at these possibilities as potential follow-ups?

robertoaloi: One thing we could do is to add a testcase explaining the limitations of the current approach in that specific situation and leave a commented out version of what we would like instead.

apps/els_lsp/src/els_parser.erl
gomoripeti: added a comment in the export section

apps/els_lsp/src/els_parser.erl
gomoripeti: added a testcase for scanning errors and commented future opportunities

rebar.config
robertoaloi: @gomoripeti If this is the only thing left, should we merge this and bump it as a follow up?

gomoripeti: I should add a hash or tag instead of a branch.
Michal provided some feedback on the PR so my branch might need modifications (and might move)

robertoaloi:
Thanks!

PR #1147, opened 2021-11-27
Fix unicode output via stdout

### LINKED ISSUE #1145: Hovering over os:cmd/1 causes a server crash
Simply add a `os:cmd("uname -a")` and hover over it. You will notice a crash in both Emacs and VS Code.

Git bisect shows this as the culprit: a84d93fb87f64765f29e2f564490aa8bc3a03d1f

@garazdawi FYI

garazdawi:
Fix problem with unicode output to stdout.

While at it also improve rendering of definition lists with code blocks in them.

Closes #1145

### DIFF
--- a/apps/els_core/src/els_stdio.erl
+++ b/apps/els_core/src/els_stdio.erl
@@ -23,7 +23,7 @@ start_listener(Cb) ->
 -spec init({function(), atom() | pid()}) -> no_return().
 init({Cb, IoDevice}) ->
   ?LOG_INFO("Starting stdio server..."),
-  ok = io:setopts(IoDevice, [binary]),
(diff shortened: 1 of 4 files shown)


apps/els_core/src/els_stdio.erl
```diff
@@ -23,7 +23,7 @@ start_listener(Cb) ->
 -spec init({function(), atom() | pid()}) -> no_return().
 init({Cb, IoDevice}) ->
   ?LOG_INFO("Starting stdio server..."),
-  ok = io:setopts(IoDevice, [binary]),
```
robertoaloi, line 26: Nice catch!

robertoaloi:
Thanks!

PR #1156, opened 2021-12-11
[#1152] Emit error in case the module name does not coincide with the filename

### LINKED ISSUE #1152: Module file name mismatch compiler error not exposed
**Describe the bug**
A module with a name not matching the file name does issue an error.

**To Reproduce**
Create a module in the file `foo.erl`:
```
-module(bar).
```

**Expected behavior**
An error similar to the Erlang compilation error should be issued:
```
$ erlc foo.erl
/path/to/foo.beam: Module name 'bar' does not match file name 'foo'
```

**Actual behavior**
No error is visible.

**Context**
 - `erlang_ls` version (tag/sha): 0.21.1
 - Editor used: Sublime Text 4 (4122)
 - LSP client used: Sublime LSP v1.14.0

robertoaloi:
### Description

Fixes #1152 .

### DIFF
--- a/apps/els_lsp/src/els_compiler_diagnostics.erl
+++ b/apps/els_lsp/src/els_compiler_diagnostics.erl
@@ -723,7 +723,41 @@ compile_file(Path, Dependencies) ->
   [code:load_binary(Dependency, Filename, Binary)
    || {{Dependency, Binary, Filename}, _} <- Olds],
   Diagnostics = lists:flatten([ Diags || {_, Diags} <- Olds ]),
-  {Res, Diagnostics}.
+  {Res, Diagnostics ++ module_name_check(Path)}.
+
+%% The module_name error is only emitted by the Erlang compiler during
+%% the "save binary" phase. This phase does not occur when code
+%% generation is disabled (e.g. by using the basic_validation or
+%% strong_validation arguments when invoking the compile:file/2
+%% function), which is exactly the case for the Erlang LS compiler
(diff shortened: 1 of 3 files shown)


apps/els_lsp/src/els_compiler_diagnostics.erl
```diff
@@ -723,7 +723,41 @@ compile_file(Path, Dependencies) ->
...
+  case els_dt_document:lookup(Uri) of
+    {ok, [Document]} ->
+      case els_dt_document:pois(Document, [module]) of
+        [#{id := Module, range := Range}] ->
+          case atom_to_list(Module) =:= Basename of
+            true ->
+              [];
+            false ->
+              Message =
+                io_lib:format("Module name '~s' does not match file name '~ts'",
```
gomoripeti, line 748: in case of a unicode module name you probably need the `t` modifier also for `Module` not just `Basename`

robertoaloi: The message is equivalent to the one in:

https://github.com/erlang/otp/blob/656690cb4f9d8b8942b8be0b7aad54f86cf7fee0/lib/compiler/src/compile.erl#L330

According to https://www.erlang.org/doc/apps/stdlib/unicode_usage.html:

> Module names, application names, and node names are still restricted to the ISO Latin-1 range

gomoripeti: oh I see 👍

PR #1158, opened 2021-12-17
[#1050] Add separate way to send log notification to avoid recursion

### LINKED ISSUE #1050: Infinite loop when LSP log level is set to debug
Now that log messages are sent through the LSP tracing facilities, we have an issue.

If the log level includes DEBUG, that causes all LSP calls to be logged, which in turns causes a new log message, which in turns causes a new LSP message, which in turns causes a new log message, which in turns causes a new LSP message, which in turns causes a new log message, which in turns causes a new LSP message, which in turns causes a new log message, which in turns causes a new LSP message, which in turns causes a new log message, which in turns causes a new LSP message, which in turns causes a new log message, which in turns causes a new LSP message, which in turns causes a new log message, which in turns causes a new LSP message, which in turns causes a new log message, which in turns causes a new LSP message, which in turns causes a new log message, which in turns causes a new LSP message, which in turns causes a new log message, which in turns causes a new LSP message...

AminArria:
Fixes #1050

### DIFF
--- a/apps/els_lsp/src/els_server.erl
+++ b/apps/els_lsp/src/els_server.erl
@@ -197,6 +197,11 @@ handle_request(Response, State0) ->
   State0.
 
 -spec do_send_notification(binary(), map(), state()) -> ok.
+%% This notification is specifically filtered out to avoid recursive
+%% calling of log notifications (see issue #1050)
+do_send_notification(<<"window/logMessage">> = Method, Params, State) ->
+  Notification = els_protocol:notification(Method, Params),
+  send(Notification, State);
 do_send_notification(Method, Params, State) ->
   Notification = els_protocol:notification(Method, Params),
   ?LOG_DEBUG( "[SERVER] Sending notification [notification=~s]"


apps/els_lsp/src/els_log_notification.erl
```diff
@@ -19,7 +19,7 @@ log(#{level := Level} = LogEvent, _Config) ->
     try
         Msg = logger_formatter:format( LogEvent
                                      , #{ template => ?LSP_LOG_FORMAT}),
-        els_server:send_notification(<<"window/logMessage">>, #{
+        els_server:send_log(<<"window/logMessage">>, #{
```
robertoaloi, line 22: A _notification_ is a LSP concept (as opposed to a _request_), so I would prefer to maintain the current naming here. Instead, I would add a dedicated clause to the `do_send_notification/3` for the `window/logMessage` method which skips logging. WDYT?

AminArria: Makes sense, but I'll include a comment (sort of like `filtered out to avoid recurvise logging`) so it makes sense why this specific notification is treated differently

robertoaloi: Great idea! Feel free to reference the issue number in the comment.

AminArria: Done

robertoaloi (approved):

robertoaloi:
Sorry for the long wait. The change looks good, waiting for CI.

PR #1182, opened 2022-01-26
Ensure that default node name is valid

plux:
This fixes an issue where initialization of els distribution would fail
due to invalid node name if the project directory contains characters
that are not valid in an erlang node name.


robertoaloi:
You're on a roll!

PR #1183, opened 2022-01-26
Fix issues with rename variable involving specs

### LINKED ISSUE #1180: Rename on a variable renames variables in a following spec
**Describe the bug**
Given this code:
```erl
foo(A) -> A.

-spec bar(A :: integer()) -> integer().
bar(A) -> A.
```
Using the "Rename" command with text "B" on either of the `A` variables in `foo` produces:
```erl
foo(B) -> B.

-spec bar(B :: integer()) -> integer().
bar(A) -> A.
```

**Expected behavior**

It should not touch the spec and produce:
```erl
foo(B) -> B.

-spec bar(A :: integer()) -> integer().
bar(A) -> A.
```

**Context**
 - `erlang_ls` version (tag/sha): Last release (Erlang LS (in erlang_ls), version: 0.21.2, OTP version: 24)
 - Editor used: VS Code

plux:
### Description

Tweak the heuristic for variable renaming inside functions so it doesn't bleed over to following -spec.
Add possibility to rename variables in specs.

Fixes #1180.

### DIFF
--- a/apps/els_lsp/priv/code_navigation/src/rename_variable.erl
+++ b/apps/els_lsp/priv/code_navigation/src/rename_variable.erl
@@ -1,5 +1,5 @@
 -module(rename_variable).
-
+-callback name(Var) -> Var.
 foo(Var) ->
   Var < 0;
 foo(Var) ->
@@ -16,15 +16,16 @@ bar(Var) ->
 baz(Var) ->
   Var.
 
-%% TODO: Add tests
 -record(foo, {a :: Var,
-              b :: Var}).
+              b :: [Var]}).
+%% BUG: `Var' in MACRO(`Var') is not considered a variable POI
+-define(MACRO(Var), Var + Var).
 
-
--define(MACRO(Var), Var).
--callback name(Var) -> Var.
 -type type(Var) :: Var.
 -opaque opaque(Var) :: Var.
+
+foo(Var) ->
+  Var.
 
 -if(Var == Var).
 
--- a/apps/els_lsp/priv/code_navigation/src/rename_variable.erl
+++ b/apps/els_lsp/priv/code_navigation/src/rename_variable.erl
@@ -15,3 +15,17 @@ bar(Var) ->
           when Var :: atom().
 baz(Var) ->
   Var.
+
+%% TODO: Add tests
+-record(foo, {a :: Var,
+              b :: Var}).
+
+
+-define(MACRO(Var), Var).
+-callback name(Var) -> Var.
+-type type(Var) :: Var.
+-opaque opaque(Var) :: Var.
+
+-if(Var == Var).
+
+-endif.
--- a/apps/els_lsp/priv/code_navigation/src/rename_variable.erl
+++ b/apps/els_lsp/priv/code_navigation/src/rename_variable.erl
@@ -10,3 +10,8 @@ foo(_Var) ->
 
 bar(Var) ->
   Var.
+
+-spec baz(Var) -> Var
+          when Var :: atom().
+baz(Var) ->
+  Var.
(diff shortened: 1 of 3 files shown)


apps/els_lsp/src/els_rename_provider.erl
```diff
@@ -222,18 +222,29 @@ new_name(_, NewName) ->
...
-         end,
-  %% Find beginning of first function clause after VarRange
-  To = case [R || #{range := R} <- FunPOIs, els_range:compare(VarRange, R)] of
-        []                 -> {999999999, 999999999}; % End of document
-        [#{from := End}|_] -> End
-       end,
-  #{from => From, to => To}.
+  POIs = els_poi:sort(els_dt_document:pois(Document, [ function_clause
+                                                     , spec
+                                                     ])),
```
michalmuskala, line 227: This should probably consider all possible top-level forms. I haven't checked, but I'd imagine similar issue will happen with things like:
```
foo(A) -> A.
-define(?MACRO(A), A).
```
and other top-level items (like `-callback` and `-type`).

TBH, I'm not sure why the logic tries to look for the beginning point of the next form. Wouldn't it be enough to look for the end point of the current clause?

robertoaloi: If I remember correctly, it's to ensure renaming works even when the current clause is not fully parsable (still being written).

apps/els_lsp/src/els_rename_provider.erl
```diff
@@ -222,18 +222,29 @@ new_name(_, NewName) ->
...
+      SpecRange;
+    [] ->
+      %% Find beginning of first function clause before VarRange
+      From =
+        case [R || #{range := R} <- POIs, els_range:compare(R, VarRange)] of
+          []        -> {0, 0}; % Beginning of document
+          FunRanges -> maps:get(from, lists:last(FunRanges))
+        end,
+      %% Find beginning of first function clause after VarRange
+      To =
```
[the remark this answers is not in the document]
gomoripeti: maybe a better idea is for els_parser to store the full range of a function clause in the `function_clause` POI data

apps/els_lsp/src/els_rename_provider.erl
plux: Good points, I will add support for other top-level forms as well. 
The reason for the currently implemented logic is that function_clause ranges don't contain the beginning and the end of the body of the clause, but rather just the function clause head.

apps/els_lsp/src/els_rename_provider.erl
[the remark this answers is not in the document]
plux: Good point, I will see if I can implement an upper bound using the function wrapping_range.
If `els_parser` could return the full range of the function_clause that would make the implementation of this functionality cleaner, not sure how tricky it is to implement in `els_parser`..

apps/els_lsp/priv/code_navigation/src/rename_variable.erl
```diff
@@ -10,3 +10,23 @@ foo(_Var) ->
...
   Var.
+
+-spec baz(Var) -> Var
+          when Var :: atom().
+baz(Var) ->
+  Var.
+
+-record(foo, {a :: Var,
+              b :: [Var]}).
+%% BUG: `Var' in MACRO(`Var') is not considered a variable POI
```
gomoripeti, line 21: thanks for finding this bug, indeed els_parser currently does not expect sub-POIs in the define name, so it does not recurse into it ([els_parser:attribute_subtrees](https://github.com/erlang-ls/erlang_ls/blob/main/apps/els_lsp/src/els_parser.erl#L972))

robertoaloi: Should we file a ticket about this?

robertoaloi (approved):
To me this looks like a very good step forward. I think we can merge this in and eventually improve it later. Thanks @plux !

PR #1184, opened 2022-01-27
Augment local config onto global config

plux:
### Description
This allows for a a global config which can be overridden by project specific configs.
As suggested by @alanz: https://github.com/erlang-ls/erlang_ls/pull/380#pullrequestreview-341787386


robertoaloi:
Looks good, thanks. Would it be possible (as a follow up) to add one or two test cases for this?

PR #1186, opened 2022-01-31
Fix parsing of define arguments

plux:
### Description

Argument variables POIs weren't parsed in macro definition.
This PR fixes that.

Fixes comment in https://github.com/erlang-ls/erlang_ls/pull/1183#discussion_r795205932


gomoripeti (approved):
🎉

robertoaloi (approved):

PR #1188, opened 2022-02-01
Allow installing into custom directory

HiPhish:
### Description

Allows the user to install the built package into a different directory than `/usr/local` by setting the `PREFIX` environment variable.

```sh
PREFIX=/usr/local/stow/erlang_ls-0.21.2 make -e install
```

I want to keep my manually installed packages easy to upgrade and remove, so I use [GNU Stow](https://www.gnu.org/software/stow/) for that purpose. I install the package to `/usr/local/stow/<name>` (where `<name>` is some arbitrary name I chose), then have Stow create symlinks into `usr/local`. Stow owns these symlinks and if I want to remove or replace a package it knows exactly which symlinks to remove.

This is much better than having a number of files from various packages lying all around the system. erlang_ls is not bad, it has only two files, but even two files here and two files there can quickly add up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,18 @@
 .PHONY: all
 
+PREFIX = '/usr/local'
+
 all:
 	@ echo "Building escript..."
 	@ rebar3 escriptize
 	@ rebar3 as dap escriptize
 
+.PHONY: install
 install: all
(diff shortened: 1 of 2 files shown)


robertoaloi:
Thanks!

PR #1196, opened 2022-02-11
Add support for renaming modules

plux:
### Description
Add support for renaming modules.
Put cursor at modname in `-module(modname)`, and trigger the rename command.
This will rename the .erl file (in the same directory), updated the -module attribute 
and update references to the module such as:
* Function applications 
* Type applications
* Implicit functions
* Imports
* -behaviour attributes

This PR also improves the "find references" functionality for modules, previously it only displayed function references.

There's a small race condition with Emacs `erlang-mode` as it wants to be helpful and suggest that the old filename doesn't match the new module name, you need to answer no when it asks if it should fix the module name.
Personally I have now disabled that check:
`(setq erlang-check-module-name nil)` 
This feature could be handled by erlang_ls with a code action instead as 
we already have a diagnostic to check the module name.
Could be a good beginner task :+1:

### DIFF
--- a/apps/els_lsp/priv/code_navigation/src/rename_module_a.erl
+++ b/apps/els_lsp/priv/code_navigation/src/rename_module_a.erl
@@ -1,0 +1,16 @@
+-module(rename_module_a).
+
+-export([ function_a/0
+        , function_b/0
+        ]).
+-export_type([type_a/0]).
+
+-type type_a() :: any().
+
+-callback function_a() -> type_a().
+
+function_a() ->
+  a.
+
+function_b() ->
+  b.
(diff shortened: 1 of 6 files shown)


robertoaloi:
> Personally I have now disabled that check:
> `(setq erlang-check-module-name nil)`

Good point, we can include that in the provided Emacs config sample:

https://github.com/erlang-ls/erlang_ls/blob/main/misc/dotemacs

robertoaloi:
> This feature could be handled by erlang_ls with a code action instead as
> we already have a diagnostic to check the module name.
> Could be a good beginner task 👍

Yes! I hinted at that in the [release notes](https://github.com/erlang-ls/erlang_ls/releases/tag/0.22.0), so let's craft an issue with that, pointing to the tutorial and let's leave it for new contributors.

robertoaloi:
Nice one, thanks!

PR #1199, opened 2022-02-15
Fix support for renaming modules

robertoaloi:
The textDocument parameter in a TextDocumentEdit must contain a version, even if null (see the OptionalVersionedTextDocumentIdentifier definition from the LSP specification). Without it, some editors (most notably VS Code) would reject the edits.


robertoaloi:
@plux FYI

plux:
Oops missed this, thanks for correction!

PR #1200, opened 2022-02-15
Fix unused variable quickaction 

### LINKED ISSUE #1082: using the add'_' to Variable causes duplication of the clause
**Describe the bug**
```
some(Variable) -> ok.
```
 and using the code action on Variable resutls in 
```
some(_Variable) -> ok.
some(Variable) -> ok.
```
 instead of just 
```
some(_Variable) -> ok.
```

**Context**
 - `erlang_ls` version (tag/sha):  0.18.0
 - Editor used: VSCode

AminArria:
Fixes #1082

The action was created with the same start and end line, which when replaced caused the line to be added instead of replaced

Wanted to add a test to actually verify the change in the file, but couldn't figure out how to get the changes or updated file after doing `els_client:workspace_executecommand/2`

### DIFF
--- a/apps/els_lsp/src/els_code_action_provider.erl
+++ b/apps/els_lsp/src/els_code_action_provider.erl
@@ -77,11 +77,17 @@ unused_variable_action(Uri, Range, Message) ->
   end.
 
 -spec make_unused_variable_action(uri(), range(), binary()) -> [map()].
-make_unused_variable_action(Uri, Range, UnusedVariable) ->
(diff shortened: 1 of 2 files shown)


robertoaloi:
@AminArria This was fixed by @plux as part of https://github.com/erlang-ls/erlang_ls/pull/1212, so I think we can close this one? But thanks anyway, the fix is essentially the same.

AminArria:
Nice, yeah I'll close it then

PR #1212, opened 2022-02-18
Implement new code actions

### LINKED ISSUE #1207: Add Code Action / Quick Fix to rename an incorrectly named module
# Context

One of the core features of Erlang LS is to be able to publish a [diagnostic message](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_publishDiagnostics) whenever an issue is detected in the user code. Diagnostics are often rendered by the IDE with a _squiggle_ of different colour, depending on the severity of the message (e.g. a _warning_ uses a yellow squiggle, while an _error_ uses a red squiggle).

One of these messages is displayed whenever [a module attribute contains a name which does not match the file name](https://www.erlang.org/doc/reference_manual/modules.html#pre-defined-module-attributes). See the picture below.

Screenshot 2022-02-17 at 09 08 11

A [Code Action request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction) is sent from the IDE to the language server to compute the possible [code actions](https://code.visualstudio.com/docs/editor/refactoring) available for a given document and text range. In the case above, the following request is sent from the IDE to the language server as soon as the user hovers the incorrect module name.

```
[Trace - 09:08:14 AM] Sending request 'textDocument/codeAction - (5913)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/robertoaloi/git/github/erlang-ls/erlang_ls/apps/els_lsp/src/els_background_job.erl"
  },
  "range": {
    "start": {
      "line": 3,
      "character": 9
    },
    "end": {
      "line": 3,
      "character": 9
    }
  },
  "context": {
    "diagnostics": [
      {
        "source": "Compiler (via Erlang LS)",
        "severity": 1,
        "range": {
          "start": {
            "line": 3,
            "character": 8
          },
          "end": {
            "line": 3,
            "character": 18
          }
        },
        "message": "Module name 'els_bg_job' does not match file name 'els_background_job'"
      }
    ]
  }
}

[Trace - 09:08:14 AM] Received response 'textDocument/codeAction - (5913)' in 77ms.
Result: []
```

As you can see, the request contains the current location (file path, line number, line column) and an additional _context_ (in this case a list containing one diagnostic, which can in turn be used by the server to provide a list of available code actions. Right now, since this feature is not yet implemented, no code actions (i.e. an empty list or `[]`) is returned to the client.

# What is Required

In case a `textDocument/codeAction` request is received by the language server, the server should check for presence of "Module name mismatch" diagnostics. If one is present, the server should include two code actions (say `rename-module` and `rename-file`) in the response.

Once the possible code action is received by the IDE, the user gets notified. This usually happens via a little bulb icon or similar, which suggests that a code action is available. In this case, the code actions would be of type _quickfix_ and suggest to either rename the module or the file, to ensure they match.

# Code Pointers

* Code action requests are dispatched to the `els_code_action_provider` module, so you probably want to start there

# Testing

Don't forget to include a test for the new functionality! You can look at existing examples in the `els_code_action_SUITE` module and at this [code lenses tutorial](https://erlang-ls.github.io/articles/tutorial-code-lenses) which includes a [testing overview](https://erlang-ls.github.io/articles/tutorial-code-lenses/#adding-tests) which also applies to code actions.

### LINKED ISSUE #1210: Add Code Action / Quick Fix to export an unused function
# Context

One of the core features of Erlang LS is to be able to publish a [diagnostic message](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_publishDiagnostics) whenever an issue is detected in the user code. Diagnostics are often rendered by the IDE with a _squiggle_ of different colour, depending on the severity of the message (e.g. a _warning_ uses a yellow squiggle, while an _error_ uses a red squiggle).

One of these messages is displayed whenever a non existing function is invoked. See the picture below.

Screenshot 2022-02-18 at 15 40 16

A [Code Action request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction) is sent from the IDE to the language server to compute the possible [code actions](https://code.visualstudio.com/docs/editor/refactoring) available for a given document and text range. In the case above, the following request is sent from the IDE to the language server as soon as the user hovers the unused function.

```
[Trace - 03:40:17 PM] Sending request 'textDocument/codeAction - (3375)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/robertoaloi/git/github/erlang-ls/erlang_ls/apps/els_lsp/src/els_server.erl"
  },
  "range": {
    "start": {
      "line": 245,
      "character": 5
    },
    "end": {
      "line": 245,
      "character": 5
    }
  },
  "context": {
    "diagnostics": [
      {
        "source": "Compiler",
        "severity": 2,
        "range": {
          "start": {
            "line": 245,
            "character": 0
          },
          "end": {
            "line": 245,
            "character": 18
          }
        },
        "message": "function my_unused_function/0 is unused",
        "code": "L1230"
      }
    ]
  }
}


[Trace - 03:40:17 PM] Received response 'textDocument/codeAction - (3375)' in 24ms.
Result: []
```

As you can see, the request contains the current location (file path, line number, line column) and an additional _context_ (in this case a list containing one diagnostic, which can in turn be used by the server to provide a list of available code actions. Right now, since this feature is not yet implemented, no code actions (i.e. an empty list or `[]`) is returned to the client.

# What is Required

In case a `textDocument/codeAction` request is received by the language server, the server should check for presence of "Unused function" (L1230) diagnostics. If one is present, the server should include a code action (say `export-function`) in the response.

Once the possible code action is received by the IDE, the user gets notified. This usually happens via a little bulb icon or similar, which suggests that a code action is available. In this case, the code action would be of type _quickfix_ and suggest to export the unused function.

As a start, the function could be exported as its own "export list" and later on we could think about a nicer behaviour (in an Erlang module multiple export lists could be present).

# Code Pointers

* Code action requests are dispatched to the `els_code_action_provider` module, so you probably want to start there

# Testing

Don't forget to include a test for the new functionality! You can look at existing examples in the `els_code_action_SUITE` module and at this [code lenses tutorial](https://erlang-ls.github.io/articles/tutorial-code-lenses) which includes a [testing overview](https://erlang-ls.github.io/articles/tutorial-code-lenses/#adding-tests) which also applies to code actions.

plux:
### Description
Implement a couple of new code actions:

* If function is unused, provide action to export it
* If variable is unbound, provide action to fix spelling
* If module name doesn't match filename, provide action to fix it

This also includes refactoring of `els_code_action_provider` and fixing
the unused variable code action that didn't work properly.

Fixes #1207.
Fixes #1210.

### DIFF
--- a/apps/els_core/src/els_utils.erl
+++ b/apps/els_core/src/els_utils.erl
@@ -20,6 +20,7 @@
         , function_signature/1
         , base64_encode_term/1
         , base64_decode_term/1
+        , levenshtein_distance/2
         ]).
 
 
@@ -449,3 +450,27 @@ base64_encode_term(Term) ->
 -spec base64_decode_term(binary()) -> any().
 base64_decode_term(Base64) ->
   binary_to_term(base64:decode(Base64)).
+
+-spec levenshtein_distance(binary(), binary()) -> integer().
+levenshtein_distance(S, T) ->
+  {Distance, _} = levenshtein_distance(to_list(S), to_list(T), #{}),
+  Distance.
+
+-spec levenshtein_distance(string(), string(), map()) -> {integer(), map()}.
+levenshtein_distance([] = S, T, Cache) ->
+  {length(T), maps:put({S, T}, length(T), Cache)};
+levenshtein_distance(S, [] = T, Cache) ->
+  {length(S), maps:put({S, T}, length(S), Cache)};
+levenshtein_distance([X|S], [X|T], Cache) ->
+  levenshtein_distance(S, T, Cache);
+levenshtein_distance([_SH|ST] = S, [_TH|TT] = T, Cache) ->
+  case maps:find({S, T}, Cache) of
+    {ok, Distance} ->
+      {Distance, Cache};
+    error ->
+      {L1, C1} = levenshtein_distance(S, TT, Cache),
+      {L2, C2} = levenshtein_distance(ST, T, C1),
+      {L3, C3} = levenshtein_distance(ST, TT, C2),
+      L = 1 + lists:min([L1, L2, L3]),
+      {L, maps:put({S, T}, L, C3)}
+  end.
--- a/apps/els_lsp/src/els_code_action_provider.erl
+++ b/apps/els_lsp/src/els_code_action_provider.erl
@@ -29,79 +29,133 @@ handle_request({document_codeaction, Params}, State) ->
 %% Internal Functions
 %%==============================================================================
 
-
 %% @doc Result: `(Command | CodeAction)[] | null'
 -spec code_actions(uri(), range(), code_action_context()) -> [map()].
-code_actions(Uri, _Range, Context) ->
-  #{ <<"diagnostics">> := Diagnostics } = Context,
-  Actions0 = [ make_code_action(Uri, D) || D <- Diagnostics],
-  Actions = lists:flatten(Actions0),
-  Actions.
+code_actions(Uri, _Range, #{<<"diagnostics">> := Diagnostics}) ->
+  lists:flatten([make_code_action(Uri, D) || D <- Diagnostics]).
 
-%% @doc Note: if the start and end line of the range are the same, the line
-%% is simply added.
--spec replace_lines_action(uri(), binary(), binary(), binary(), range())
-                          -> map().
-replace_lines_action(Uri, Title, Kind, Lines, Range) ->
-  #{ <<"start">> := #{ <<"character">> := _StartCol
-                     , <<"line">>      := StartLine }
-   , <<"end">>   := #{ <<"character">> := _EndCol
-                     , <<"line">>      := EndLine }
-   } = Range,
-  #{ title => Title
-   , kind => Kind
-   , command =>
-       els_command:make_command( Title
-                               , <<"replace-lines">>
-                               , [#{ uri   => Uri
-                                   , lines => Lines
-                                   , from  => StartLine
-                                   , to    => EndLine }])
-   }.
+-spec make_code_action(uri(), map()) -> [map()].
+make_code_action(Uri, #{<<"message">> := Message, <<"range">> := Range}) ->
+  make_code_action(
+    [ {"function (.*) is unused", fun action_export_function/3}
+    , {"variable '(.*)' is unused", fun action_ignore_variable/3}
+    , {"variable '(.*)' is unbound", fun action_suggest_variable/3}
+    , {"Module name '(.*)' does not match file name '(.*)'",
+       fun action_fix_module_name/3}
+    ], Uri, Range, Message).
 
--spec make_code_action(uri(), els_diagnostics:diagnostic()) -> [map()].
-make_code_action(Uri, #{ <<"message">> := Message
-                       , <<"range">>   := Range } = _Diagnostic) ->
-  unused_variable_action(Uri, Range, Message).
+-spec make_code_action([{string(), Fun}], uri(), range(), binary()) -> [map()]
+          when Fun :: fun((uri(), range(), [binary()]) -> [map()]).
+make_code_action([], _Uri, _Range, _Message) ->
+  [];
+make_code_action([{RE, Fun}|Rest], Uri, Range, Message) ->
+  Actions = case re:run(Message, RE, [{capture, all_but_first, binary}]) of
+              {match, Matches} ->
+                Fun(Uri, Range, Matches);
+              nomatch ->
+                []
+            end,
+  Actions ++ make_code_action(Rest, Uri, Range, Message).
 
-%%------------------------------------------------------------------------------
-
--spec unused_variable_action(uri(), range(), binary()) -> [map()].
-unused_variable_action(Uri, Range, Message) ->
-  %% Processing messages like "variable 'Foo' is unused"
-  case re:run(Message, "variable '(.*)' is unused"
-             , [{capture, all_but_first, binary}]) of
-      {match, [UnusedVariable]} ->
-          make_unused_variable_action(Uri, Range, UnusedVariable);
-      _ -> []
+-spec action_export_function(uri(), range(), [binary()]) -> [map()].
+action_export_function(Uri, _Range, [UnusedFun]) ->
+  {ok, Document} = els_utils:lookup_document(Uri),
+  case els_poi:sort(els_dt_document:pois(Document, [module, export])) of
+    [] ->
+      [];
+    POIs ->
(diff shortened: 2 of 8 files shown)


apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -29,79 +29,133 @@ handle_request({document_codeaction, Params}, State) ->
...
-                     , <<"line">>      := StartLine }
-   , <<"end">>   := #{ <<"character">> := _EndCol
-                     , <<"line">>      := EndLine }
-   } = Range,
+code_actions(Uri, _Range, #{<<"diagnostics">> := Diagnostics}) ->
+  lists:flatten([make_code_action(Uri, D) || D <- Diagnostics]).
+
+-spec make_code_action(uri(), map()) -> [map()].
+make_code_action(Uri, #{<<"message">> := Message, <<"range">> := Range}) ->
+  make_code_action(
```
robertoaloi, line 39: Nice! This will make it much simpler to contribute new code actions, thanks!

apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -29,79 +29,133 @@ handle_request({document_codeaction, Params}, State) ->
...
+    , {"Module name '(.*)' does not match file name '(.*)'",
+       fun action_fix_module_name/3}
+    ], Uri, Range, Message).
+
+-spec make_code_action([{string(), Fun}], uri(), range(), binary()) -> [map()]
+          when Fun :: fun((uri(), range(), [binary()]) -> [map()]).
+make_code_action([], _Uri, _Range, _Message) ->
+  [];
+make_code_action([{RE, Fun}|Rest], Uri, Range, Message) ->
+  Actions = case re:run(Message, RE, [{capture, all_but_first, binary}]) of
```
robertoaloi, line 52: Later on we may want to use error codes everywhere and avoid regular expressions.

apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -29,79 +29,133 @@ handle_request({document_codeaction, Params}, State) ->
...
+              {match, Matches} ->
+                Fun(Uri, Range, Matches);
+              nomatch ->
+                []
+            end,
+  Actions ++ make_code_action(Rest, Uri, Range, Message).
+
+-spec action_export_function(uri(), range(), [binary()]) -> [map()].
+action_export_function(Uri, _Range, [UnusedFun]) ->
```
robertoaloi, line 61: Matter of taste, but we could move the actual business logic to a separate module (say `els_code_actions`).

apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -29,79 +29,133 @@ handle_request({document_codeaction, Params}, State) ->
...
-replace_lines_action(Uri, Title, Kind, Lines, Range) ->
-  #{ <<"start">> := #{ <<"character">> := _StartCol
-                     , <<"line">>      := StartLine }
-   , <<"end">>   := #{ <<"character">> := _EndCol
-                     , <<"line">>      := EndLine }
-   } = Range,
+code_actions(Uri, _Range, #{<<"diagnostics">> := Diagnostics}) ->
+  lists:flatten([make_code_action(Uri, D) || D <- Diagnostics]).
+
+-spec make_code_action(uri(), map()) -> [map()].
```
robertoaloi, line 37: Nit: we return a list of actions, so maybe rename the function accordingly.

apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -29,79 +29,133 @@ handle_request({document_codeaction, Params}, State) ->
...
+make_code_action(Uri, #{<<"message">> := Message, <<"range">> := Range}) ->
+  make_code_action(
+    [ {"function (.*) is unused", fun action_export_function/3}
```
robertoaloi, line 40: Each diagnostic could return a number of code actions. For example, an unused function warning could result in the ability of exporting or removing a function. So, we could either pass a list of funs here or rename the fun into something like `unused_function_actions` and return there the list of options.

apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -29,79 +29,133 @@ handle_request({document_codeaction, Params}, State) ->
...
+  Actions ++ make_code_action(Rest, Uri, Range, Message).
+
+-spec action_export_function(uri(), range(), [binary()]) -> [map()].
+action_export_function(Uri, _Range, [UnusedFun]) ->
+  {ok, Document} = els_utils:lookup_document(Uri),
+  case els_poi:sort(els_dt_document:pois(Document, [module, export])) of
+    [] ->
+      [];
+    POIs ->
+      #{range := #{to := {Line, _Col}}} = lists:last(POIs),
```
robertoaloi, line 67: Maybe, when multiple export lists are present, we could present them as multiple code actions (i.e. add to first list, add to second list, etc).

robertoaloi (approved):

apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -29,79 +29,133 @@ handle_request({document_codeaction, Params}, State) ->
...
+                      els_range:compare(R, VarRange)],
+      case [{els_utils:levenshtein_distance(V, Var), V} ||
+             V <- VarsInScope,
+             V =/= Var,
+             binary:at(Var, 0) =:= binary:at(V, 0)]
+      of
+        [] ->
+          [];
+        VariableDistances ->
+          {_, SimilarVariable} = lists:min(VariableDistances),
```
robertoaloi, line 112: Maybe we should also set a threshold as well, so that completely different variable names are not presented as an option.

apps/els_core/src/els_utils.erl
```diff
@@ -449,3 +450,27 @@ base64_encode_term(Term) ->
 -spec base64_decode_term(binary()) -> any().
 base64_decode_term(Base64) ->
   binary_to_term(base64:decode(Base64)).
+
+-spec levenshtein_distance(binary(), binary()) -> integer().
+levenshtein_distance(S, T) ->
```
michalmuskala, line 455: In general the "Jaro distance" algorithm is a better metric to use for comparing identifiers - it make it much easier to compare giving a 0..1 score for similarity between two strings, rather than the number of edits.

It also gives somewhat higher weight to matching prefixes which often also helps.

Elixir uses [`String.jaro_distance/2`](https://hexdocs.pm/elixir/String.html#jaro_distance/2) for exactly this purpose - we could probably port the implementation.

PR #1215, opened 2022-02-19
Drop OTP 21

robertoaloi:


gomoripeti:
Should the README and rebar.config also be updated? https://github.com/erlang-ls/erlang_ls/blame/main/README.md#L12 https://github.com/erlang-ls/erlang_ls/blob/main/rebar.config#L33

robertoaloi:
> Should the README and rebar.config also be updated? https://github.com/erlang-ls/erlang_ls/blame/main/README.md#L12 https://github.com/erlang-ls/erlang_ls/blob/main/rebar.config#L33

Yes! https://github.com/erlang-ls/erlang_ls/pull/1217

PR #1223, opened 2022-02-23
Code actions refactoring to fix review comments

plux:
### Description

Fixes review comments in #1212.

### DIFF
--- a/apps/els_lsp/src/els_code_actions.erl
+++ b/apps/els_lsp/src/els_code_actions.erl
@@ -1,0 +1,159 @@
+-module(els_code_actions).
+-export([ create_function/4
+        , export_function/4
+        , fix_module_name/4
+        , ignore_variable/4
+        , remove_macro/4
+        , remove_unused/4
+        , suggest_variable/4
+        ]).
+
+-include("els_lsp.hrl").
+
+-spec create_function(uri(), range(), binary(), [binary()]) -> [map()].
+create_function(Uri, _Range, _Data, [UndefinedFun]) ->
+  {ok, Document} = els_utils:lookup_document(Uri),
+  case els_poi:sort(els_dt_document:pois(Document)) of
+    [] ->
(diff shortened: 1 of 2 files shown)


apps/els_lsp/src/els_code_actions.erl
```diff
@@ -0,0 +1,120 @@
...
+                        , els_protocol:range(Range)) ];
+    error ->
+      []
+  end.
+
+%%==============================================================================
+%% Internal functions
+%%==============================================================================
+
+-spec ensure_range(poi_range(), binary(), [poi()]) -> {ok, poi_range()} | error.
```
robertoaloi, line 97: As per our offline discussion, this function is worth a comment, since it's used for two main reasons:

* No column numbers before OTP 24
* Handling a situation where diagnostics are not up-to-date with a modified buffer

apps/els_lsp/src/els_code_actions.erl
```diff
@@ -0,0 +1,120 @@
...
+ensure_range(#{from := {Line, _}}, SubjectId, POIs) ->
+  SubjectAtom = binary_to_atom(SubjectId, utf8),
```
robertoaloi, line 99: Ids are not necessarily atoms. In the case of includes, for example, they are strings. So this will need to be adapted later on.

robertoaloi (approved):
Nice cleanup. There are a couple of PRs related to code actions from new contributors, so maybe wait with this one to avoid conflicts on their side? Otherwise looks good.

robertoaloi:
@plux It should be fine to rebase this now. Sorry for the trouble and thanks for being kind to new contributors :)

robertoaloi:
@plux Planning to rebase or should I give it a go?

plux:
I plan to do it, but I haven't had time to work on it. Sorry! 
Please give it a try if you want to.

PR #1224, opened 2022-02-23
Better support for goto definition when parsing is incomplete

plux:
### Description

If no POI is matched, try parsing subsets of the text to try to find find a match.  
This should improve the user experience when working with unfinished code.


apps/els_lsp/src/els_definition_provider.erl
```diff
@@ -45,3 +52,59 @@ goto_definition(Uri, [POI|Rest]) ->
     _ ->
       goto_definition(Uri, Rest)
   end.
+
+-spec match_incomplete(binary(), pos()) -> [poi()].
+match_incomplete(Text, Pos) ->
```
gomoripeti, line 57: to me these nice heuristics look useful even outside of definition provider, are they?

Would it make sense to gather them into a common place (either into els_parser module or a new els_incomplete_parser module)?

(there is already a section in els_parser that tries to parse incomplete forms. It tries to remove tokens from the end until the thing becomes valid. I guess that is equivalent to a certain kind of `match_before`. Where that function stops parsing, these functions could continue. Im not sure if these should be standalone exported functions that parse a given snippet of a module or should they be part of the generic full module parsing where resulting POIs are stored in the DB)

apps/els_lsp/src/els_definition_provider.erl
```diff
@@ -45,3 +52,59 @@ goto_definition(Uri, [POI|Rest]) ->
...
+
+-spec match_line(binary(), pos()) -> [poi()].
+match_line(Text, {Line, Character}) ->
+  %% Try to parse only current line
+  LineText0 = string:trim(els_text:line(Text, Line), trailing, ",;"),
+  case els_parser:parse(LineText0) of
+    {ok, []} ->
+      LineStr = els_utils:to_list(LineText0),
+      case lists:reverse(LineStr) of
+        "fo " ++ _ -> %% Kludge to parse "case foo() of"
```
gomoripeti, line 76: clever 👍

els_code_actions.erl
```diff
@@ -0,0 +1,120 @@
+-module(els_code_actions).
```
gomoripeti, line 1: is this module just accidentally part of this commit?

apps/els_lsp/src/els_definition_provider.erl
```diff
@@ -45,3 +52,59 @@ goto_definition(Uri, [POI|Rest]) ->
...
+-spec match_incomplete(binary(), pos()) -> [poi()].
+match_incomplete(Text, Pos) ->
+  %% Try parsing subsets of text to find a matching POI at Pos
+  match_after(Text, Pos) ++ match_line(Text, Pos).
+
+-spec match_after(binary(), pos()) -> [poi()].
+match_after(Text, {Line, Character}) ->
+  %% Try to parse current line and the lines after it
+  {_, AfterText} = els_text:split_at_line(Text, Line),
+  {ok, POIs} = els_parser:parse(AfterText),
```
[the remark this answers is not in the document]
plux: Cool I will try it out!

apps/els_lsp/src/els_definition_provider.erl
plux: I agree, should probably pull these out to a separate module. Most obviously we could use it for reference provider.

els_code_actions.erl
plux: Oops good catch. Belongs to another pr

PR #1225, opened 2022-02-23
Use Jaro distance for the unbound variable code action

plux:
### Description
Fixes PR comment: https://github.com/erlang-ls/erlang_ls/pull/1212#discussion_r810965486

@michalmuskala

### DIFF
--- a/apps/els_lsp/src/els_code_action_provider.erl
+++ b/apps/els_lsp/src/els_code_action_provider.erl
@@ -101,21 +101,15 @@ action_suggest_variable(Uri, Range, [Var]) ->
                       #{range := R, id := Id} <- POIs,
                       els_range:in(R, ScopeRange),
                       els_range:compare(R, VarRange)],
-      case [{els_utils:levenshtein_distance(V, Var), V} ||
-             V <- VarsInScope,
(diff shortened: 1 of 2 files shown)


apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -101,21 +101,15 @@ action_suggest_variable(Uri, Range, [Var]) ->
...
-        VariableDistances ->
-          {_, SimilarVariable} = lists:min(VariableDistances),
-          [ make_edit_action( Uri
-                            , <<"Did you mean '", SimilarVariable/binary, "'?">>
-                            , ?CODE_ACTION_KIND_QUICKFIX
-                            , SimilarVariable
-                            , els_protocol:range(VarRange)) ]
-      end;
+      VariableDistances =
+        [{els_utils:jaro_distance(V, Var), V} || V <- VarsInScope, V =/= Var],
```
robertoaloi, line 105: Do we want to keep the Levenshtain distance as well (I'm happy with that, since it may turn out useful in the future)?

robertoaloi (approved):

PR #1227, opened 2022-02-25
Add Fix for #1202 remove unused include

### LINKED ISSUE #1202: Add Code Action / Quick Fix to remove unused include
# Context

One of the core features of Erlang LS is to be able to publish a [diagnostic message](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_publishDiagnostics) whenever an issue is detected in the user code. Diagnostics are often rendered by the IDE with a _squiggle_ of different colour, depending on the severity of the message (e.g. a _warning_ uses a yellow squiggle, while an _error_ uses a red squiggle).

One of these messages is displayed whenever an `include` or `include_lib` attribute is added to a module, but unused. See the picture below.

Screenshot 2022-02-16 at 15 59 03

A [Code Action request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction) is sent from the IDE to the language server to compute the possible [code actions](https://code.visualstudio.com/docs/editor/refactoring) available for a given document and text range. In the case above, the following request is sent from the IDE to the language server as soon as the user hovers the unused include:

```
[Trace - 03:59:43 PM] Sending request 'textDocument/codeAction - (4981)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/robertoaloi/git/github/erlang-ls/erlang_ls/apps/els_lsp/test/els_rename_SUITE.erl"
  },
  "range": {
    "start": {
      "line": 37,
      "character": 25
    },
    "end": {
      "line": 37,
      "character": 25
    }
  },
  "context": {
    "diagnostics": [
      {
        "source": "UnusedIncludes",
        "severity": 2,
        "range": {
          "start": {
            "line": 37,
            "character": 0
          },
          "end": {
            "line": 37,
            "character": 40
          }
        },
        "message": "Unused file: file.hrl"
      }
    ]
  }
}


[Trace - 03:59:43 PM] Received response 'textDocument/codeAction - (4981)' in 48ms.
Result: []
```

As you can see, the request contains the current location (file path, line number, line column) and an additional _context_ (in this case a list containing one diagnostic, which can in turn be used by the server to provide a list of available code actions. Right now, since this feature is not yet implemented, no code actions (i.e. an empty list or `[]`) is returned to the client.

# What is Required

In case a `textDocument/codeAction` request is received by the language server, the server should check for presence of "Unused include" diagnostics. If one is present, the server should include a code action (say `remove-unused-include`) in the response.

Once the possible code action is received by the IDE, the user gets notified. This usually happens via a little bulb icon or similar, which suggests that a code action is available. In this case, the code action would be of type _quickfix_ and suggest to remove the unused include.

# Code Pointers

* Code action requests are dispatched to the `els_code_action_provider` module, so you probably want to start there

# Testing

Don't forget to include a test for the new functionality! You can look at existing examples in the `els_code_action_SUITE` module and at this [code lenses tutorial](https://erlang-ls.github.io/articles/tutorial-code-lenses) which includes a [testing overview](https://erlang-ls.github.io/articles/tutorial-code-lenses/#adding-tests) which also applies to code actions.

rahulraina7:
### Adds code action to remove unused fixes / include

1. Send additional diagnostics to push data back to code actions.
2. Update tests.

Fixes #1202  .

### DIFF
--- a/apps/els_lsp/src/els_code_action_provider.erl
+++ b/apps/els_lsp/src/els_code_action_provider.erl
@@ -35,31 +35,35 @@ code_actions(Uri, _Range, #{<<"diagnostics">> := Diagnostics}) ->
   lists:flatten([make_code_action(Uri, D) || D <- Diagnostics]).
 
 -spec make_code_action(uri(), map()) -> [map()].
-make_code_action(Uri, #{<<"message">> := Message, <<"range">> := Range}) ->
+make_code_action(Uri,
+  #{<<"message">> := Message, <<"range">> := Range} = D) ->
+    Data = maps:get(<<"data">>, D, <<>>),
   make_code_action(
-    [ {"function (.*) is unused", fun action_export_function/3}
-    , {"variable '(.*)' is unused", fun action_ignore_variable/3}
-    , {"variable '(.*)' is unbound", fun action_suggest_variable/3}
+    [ {"function (.*) is unused", fun action_export_function/4}
+    , {"variable '(.*)' is unused", fun action_ignore_variable/4}
+    , {"variable '(.*)' is unbound", fun action_suggest_variable/4}
     , {"Module name '(.*)' does not match file name '(.*)'",
-       fun action_fix_module_name/3}
-    , {"Unused macro: (.*)", fun action_remove_macro/3}
-    ], Uri, Range, Message).
+       fun action_fix_module_name/4}
+    , {"Unused macro: (.*)", fun action_remove_macro/4}
+    , {"Unused file: (.*)", fun action_remove_unused/4}
+    ], Uri, Range, Data, Message).
 
--spec make_code_action([{string(), Fun}], uri(), range(), binary()) -> [map()]
+-spec make_code_action([{string(), Fun}], uri(), range(), binary(), binary())
+       -> [map()]
           when Fun :: fun((uri(), range(), [binary()]) -> [map()]).
-make_code_action([], _Uri, _Range, _Message) ->
+make_code_action([], _Uri, _Range, _Data, _Message) ->
   [];
-make_code_action([{RE, Fun}|Rest], Uri, Range, Message) ->
+make_code_action([{RE, Fun}|Rest], Uri, Range, Data, Message) ->
   Actions = case re:run(Message, RE, [{capture, all_but_first, binary}]) of
               {match, Matches} ->
-                Fun(Uri, Range, Matches);
+                Fun(Uri, Range, Data, Matches);
               nomatch ->
                 []
             end,
-  Actions ++ make_code_action(Rest, Uri, Range, Message).
+  Actions ++ make_code_action(Rest, Uri, Range, Data, Message).
 
--spec action_export_function(uri(), range(), [binary()]) -> [map()].
-action_export_function(Uri, _Range, [UnusedFun]) ->
+-spec action_export_function(uri(), range(), binary(), [binary()]) -> [map()].
+action_export_function(Uri, _Range, _Data, [UnusedFun]) ->
   {ok, Document} = els_utils:lookup_document(Uri),
   case els_poi:sort(els_dt_document:pois(Document, [module, export])) of
     [] ->
@@ -74,8 +78,8 @@ action_export_function(Uri, _Range, [UnusedFun]) ->
                         , els_protocol:range(#{from => Pos, to => Pos})) ]
   end.
 
--spec action_ignore_variable(uri(), range(), [binary()]) -> [map()].
-action_ignore_variable(Uri, Range, [UnusedVariable]) ->
+-spec action_ignore_variable(uri(), range(), binary(), [binary()]) -> [map()].
+action_ignore_variable(Uri, Range, _Data, [UnusedVariable]) ->
   {ok, Document} = els_utils:lookup_document(Uri),
   POIs = els_poi:sort(els_dt_document:pois(Document, [variable])),
   case ensure_range(els_range:to_poi_range(Range), UnusedVariable, POIs) of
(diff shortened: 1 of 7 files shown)


apps/els_lsp/src/els_unused_includes_diagnostics.erl
```diff
@@ -112,11 +113,8 @@ expand_includes(Document) ->
...
 inclusion_range(Uri, Document) ->
-  Path = binary_to_list(els_uri:path(Uri)),
-  case
-    els_compiler_diagnostics:inclusion_range(Path, Document, include) ++
-    els_compiler_diagnostics:inclusion_range(Path, Document, include_lib) of
-    [Range|_] ->
+  case els_range:inclusion_range(Uri, Document) of
+    {ok, Range} ->
       Range;
     _ ->
```
robertoaloi, line 119: Nit: Since this has been refactored (nice!), this pattern can be restricted to `error`.

apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -35,31 +35,35 @@ code_actions(Uri, _Range, #{<<"diagnostics">> := Diagnostics}) ->
...
 -spec make_code_action(uri(), map()) -> [map()].
-make_code_action(Uri, #{<<"message">> := Message, <<"range">> := Range}) ->
+make_code_action(Uri,
+  #{<<"message">> := Message, <<"range">> := Range} = D) ->
+    Data = maps:get(<<"data">>, D, <<>>),
   make_code_action(
-    [ {"function (.*) is unused", fun action_export_function/3}
-    , {"variable '(.*)' is unused", fun action_ignore_variable/3}
-    , {"variable '(.*)' is unbound", fun action_suggest_variable/3}
+    [ {"function (.*) is unused", fun action_export_function/4}
```
robertoaloi, line 42: Good that we are now passing the optional data field to the specific functions. It could be a nice way to exchange information. Later on (no need to do it as part of this PR), we could always ensure there's data in a diagnostic when constructing one. We may also want to use a map instead of a binary and base64 encode it to ensure it can be passed around between client and server. See the `els_utils:base64_encode_term/1` function and its usages for Call Hierarchy. Again, this is more of a mental note and can be done as a separate task.

apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -147,6 +151,21 @@ action_remove_macro(Uri, Range, [Macro]) ->
       []
   end.
 
+-spec action_remove_unused(uri(), range(), binary(), [binary()]) -> [map()].
+action_remove_unused(Uri, _Range0, Data, [Var]) ->
+  {ok, Document} = els_utils:lookup_document(Uri),
+  case els_range:inclusion_range(Data, Document) of
```
robertoaloi, line 157: I suspect we still need to go through the `ensure_range` function like in the other cases, to avoid situations where the file changed in the meantime, causing text edits not to be applicable any longer. @plux we probably need to figure out a better way to handle this, since this step can easily be missed. Maybe have a synchronization point before applying edits?

robertoaloi (commented):
Hi @rahulraina7! This looks great. I believe the only thing missing is to go through the `ensure_range` function like in the other code actions. Other than that, LGTM!

apps/els_lsp/src/els_code_action_provider.erl
plux: Instead of using POIs we could store the text that the diagnostic refers to and simply check that the document text in the range that the diagnostic points to is the same. Does that make sense?

rahulraina7: @robertoaloi  Isn't inclusion_range essentially doing the same thing instead of using POI's its using text from Diagnostics to find a range. I can't directly reuse `ensure_range` since it wont work for this use-case specifically this line ` Id =:= SubjectAtom` , I can either make another similar function `ensure_range_imports` which will have the same idea as this one, if it makes sense ?

robertoaloi: To me we have two options:
1. Add a clause to the `ensure_range` which doesn't perform the conversion to atoms
2. Probably cleaner, what @plux suggested: store the text that the diagnostic refers to on construction and then check here that this is still a valid "selection".

rahulraina7: Sorry I may be missing something.
`els_range:inclusion_range` looks like this.

```
inclusion_range(Uri, Document) ->
  Path = binary_to_list(els_uri:path(Uri)),
  case
    els_compiler_diagnostics:inclusion_range(Path, Document, include) ++
    els_compiler_diagnostics:inclusion_range(Path, Document, include_lib) of
    [Range|_] ->
      {ok, Range};
    [] ->
      error
  end.
```

`els_compiler_diagnostics:inclusion_range` Looks like this

```
inclusion_range(IncludePath, Document, include) ->
  POIs       = els_dt_document:pois(Document, [include]),
  IncludeId  = els_utils:include_id(IncludePath),
```
This is what we want right ?

robertoaloi: I think you have a point. The major difference (compared to the other code actions) is that we ignore the range returned by the diagnostic and we just fetch it directly from the updated list of POIs. In the other cases we would check whether the range from the updated list of POIs would fit inside the same line of the one reported by the diagnostic (prior versions of Erlang didn't have column numbers in diagnostics). In practice I'm not sure this matters much for unused includes. I am ok to merge this one as-is and to eventually rethink this "checking" strategy, either by taking @plux's approach or by versioning the version of the text.

robertoaloi (approved):

robertoaloi:
Thanks!

PR #1228, opened 2022-02-28
Integrate Wrangler

kdmnk:
Hi Everyone!

Here is our proposition to integrate the Wrangler refactoring tool into ELS with minimal code changes on your side.

Approach: The user enables Wrangler in the config, then the installed Wrangler executables loads into ELS. On document_codeaction requests, the code actions coming from Wrangler will be added to the response. On workspace_executecommand requests, the corresponding commands will be passed to Wrangler.
In the future, we may add a similar ability to add code lenses.

The change have 3 parts:
- this pr
- client side changes: currently only for erlang-ls/vscode: https://github.com/erlang-ls/vscode/pull/125
- pr to wrangler: https://github.com/RefactoringTools/wrangler/pull/96

To try it [install Wrangler](https://github.com/RefactoringTools/wrangler/blob/master/INSTALL) (with my changes) and the following lines to erlang_ls.config file:
```
wrangler:
  enabled: true
  tab_width: 8
  enabled_refactorings:
  - "comment-out-spec"
  - "fold-expression"
  - "generalise-fun"
  - "move-fun"
  - "new-fun"
  - "new-macro"
  - "new-var"
  - "rename-fun"
  - "rename-var"

```

We are happy to discuss once again whether the approach stated is right for everyone and tell me what do you think about the next steps (especially documentation).

See my questions about this pr below.


apps/els_lsp/src/erlang_ls.erl
```diff
@@ -96,7 +108,24 @@ set(log_dir, Dir) ->
 set(log_level, Level) ->
   application:set_env(els_core, log_level, list_to_atom(Level));
 set(port_old, Port) ->
-  application:set_env(els_core, port, Port).
+  application:set_env(els_core, port, Port);
+set(wrangler_enabled, Opt) ->
+  application:set_env(els_core, wrangler_enabled, Opt);
+set(wrangler_dir, "undefined") -> ok;
+set(wrangler_dir, Dir) ->
+  %% FIXME can't log since logging is not configured yet
```
kdmnk, line 116: configure_logging() is called after parse_args(). What do you suggest?

apps/els_lsp/src/els_code_action_provider.erl
```diff
@@ -31,8 +31,12 @@ handle_request({document_codeaction, Params}, State) ->
 
 %% @doc Result: `(Command | CodeAction)[] | null'
 -spec code_actions(uri(), range(), code_action_context()) -> [map()].
-code_actions(Uri, _Range, #{<<"diagnostics">> := Diagnostics}) ->
-  lists:flatten([make_code_action(Uri, D) || D <- Diagnostics]).
+code_actions(Uri, Range, #{<<"diagnostics">> := Diagnostics}) ->
+  Actions = lists:flatten([make_code_action(Uri, D) || D <- Diagnostics]),
+  case application:get_env(els_core, wrangler_enabled) of
```
robertoaloi, line 36: Ideally, this module wouldn't know about Wrangler or other specific "backends".
The assumption here was that code actions are linked to diagnostics, which is not true in your case.
I wonder if we should make this a bit more general and allow external code action providers, not necessarily linked to diagnostics.

apps/els_lsp/src/els_execute_command_provider.erl
```diff
@@ -24,12 +24,17 @@ is_enabled() -> true.
...
-                 , els_command:with_prefix(<<"suggest-spec">>)
-                 , els_command:with_prefix(<<"function-references">>)
-                 ] }.
+  Commands = [ <<"server-info">>
+             , <<"ct-run-test">>
+             , <<"show-behaviour-usages">>
+             , <<"suggest-spec">>
+             , <<"function-references">>
+             ],
+  Commands2 = case application:get_env(els_core, wrangler_enabled) of
```
robertoaloi, line 33: Same concept would apply here.

apps/els_lsp/src/erlang_ls.erl
robertoaloi: In https://github.com/erlang-ls/erlang_ls/pull/1148/files there's an attempt at extending the `els_diagnostics` behaviour with an `init` function. We could use that approach and have a callback module for Wrangler. In the `init`you could then initialize Wrangler.

kdmnk: Since we may want to extend not only code actions but code lenses (and semantic highlights), we need to initialize wrangler in general (and once only, when the extension is loaded). That is different from what you have suggested, right?

apps/els_lsp/src/els_code_action_provider.erl
kdmnk: Yes, you are right. I can attend the meeting on Friday to discuss.

robertoaloi:
So, as discussed in the weekly sync:

* We will use this PR as an opportunity to revisit support for plugins in Erlang LS
* We will provide a plugin "behaviour" which allows a third-party system (eg Wrangler, Gradualizer, RefactorErl) to contribute "features" (eg code lenses, diagnostics, etc)
* The Wrangler integration can be rebased on top of the new architecture

The goal is to reduce the number of dependencies in Erlang LS, as well as provide third-party tools a documented way to integrate.

Each plugin will have its own dedicated section in the config, so plugin-specific options won't need to be passed via the CLI.

We may eventually provide building blocks for interactions with third-party tools (eg via code path, stdio, or Erlang RPC).

robertoaloi:
FYI @TheGeorge @alanz @kdmnk

apps/els_lsp/src/erlang_ls.erl
```diff
@@ -66,6 +66,18 @@ opt_spec_list() ->
     , {string, "stdio"}
     , "DEPRECATED. Only the \"stdio\" transport is currently supported."
     }
+  , { wrangler_enabled
```
robertoaloi, line 69: Should we move these to be using the regular config?

robertoaloi:
@kdmnk Thanks for this! Nice that you introduced support for semantic tokens as well.
Apart from the CI issues, I would move the config to use the regular `erlang_ls.config` file, under a "wrangler" section.
Other than that, it looks good and it will look cleaner once we roll out support for plugin.

kdmnk:
Hi @robertoaloi, I added the initialisation to the config file and I hope the code is nicer with the wrangler_handler module handling the third party calls. What do you think?

robertoaloi:
@kdmnk Looks great, thanks. My only comment would be that, since the semantic token provider is currently only provided by Wrangler, I would return `false` in the server capabilities part if wrangler is not configured and return the options otherwise.

kdmnk:
FYI Documentation for our tool: https://refactoringtools.github.io/docs/wrangler/wls/wrangler_language_server

robertoaloi:
@kdmnk I think formatting is the only part left. I added one question about capabilities.

apps/els_lsp/src/els_general_provider.erl
```diff
@@ -154,7 +154,17 @@ server_capabilities() ->
                 renameProvider =>
                     els_rename_provider:is_enabled(),
                 callHierarchyProvider =>
-                    els_call_hierarchy_provider:is_enabled()
+                    els_call_hierarchy_provider:is_enabled(),
+                semanticTokensProvider =>
```
robertoaloi, line 158: In the case Wrangler is disabled, is this going to be equivalent to a `semanticTokensProvider => false` or do we get extra traffic?

kdmnk: Yes, wls_handler returns false if Wrangler is not enabled in the config, there are no outgoing calls.

robertoaloi: What I meant by "traffic" is communication between the client (editor) and the LS. In other words, is the following (which we would be sending in case Wrangler is disabled):

```
semanticTokensProvider => #{
  legend =>
    #{
       tokenTypes => [],
       tokenModifiers => []
    },
    range => false,
   full => true
}
```                    

Equivalent to:

```
semanticTokensProvider => false
```
 
? Or would client start sending requests to the server that would result in empty responses?

kdmnk: Right, I fixed this, the is_enabled() function had not been used. Now the client does not request semantic tokens, I also tested it.

kdmnk: According to the protocol, the semanticTokensProvider property can not be a boolean value: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokensOptions

robertoaloi (approved):

rebar.config
```diff
@@ -88,8 +88,8 @@
     deprecated_function_calls,
     deprecated_functions
 ]}.
-%% Set xref ignores for functions introduced in OTP 23
-{xref_ignores, [{code, get_doc, 1}, {shell_docs, render, 4}]}.
+%% Set xref ignores for functions introduced in OTP 23 & wrangler_handler
+{xref_ignores, [wls_handler, {code, get_doc, 1}, {shell_docs, render, 4}]}.
```
robertoaloi, line 92: You probably want something like:

%% Set xref ignores for functions introduced in OTP 23 & Wrangler modules:
{xref_ignores, [
    wls_handler,
    {code, get_doc, 1},
    {shell_docs, render, 4},
    {api_wrangler, start, 0},
    wrangler_handler,
    wls_code_lens,
    wls_highlight,
    wls_semantic_tokens,
    wls_code_actions,
    wls_execute_command_provider
]}.

kdmnk: Thanks!

kdmnk: @robertoaloi now?

robertoaloi (approved):

Zabrane:
@robertoaloi @kdmnk can someone explain how to configure Wrangler with Elang-LS please? Found nothing in the doc

kdmnk:
@Zabrane Hi! You can find the documentation here: https://refactoringtools.github.io/docs/wrangler/wls/wrangler_language_server

Zabrane:
@kdmnk amazing job guys. It works perfectly !!!

PR #1233, opened 2022-03-02
Avoid negative line numbers in elvis diagnostics.

ztion:
### Description

I noticed that elvis will set line number to 0 for the atom_naming_convention diagnostic, which results in ErlangLS sending -1 as line number to the LSP client. This makes at least VS code not show any other elvis diagnostics, even other diagnostics with valid line numbers.

This fix will set line number to 1 if we get 0 or below from elvis.

Will try to make a pull request to elvis_core as well to make it send an actual line number, but I think this precaution can still be good.


robertoaloi:
Thanks! Yes, a fix to Elvis would be lovely.

ztion:
The fix for elvis is already merged too, and in fact they even put out a new version on hex.pm! Nice guys.

https://github.com/inaka/elvis_core/pull/223

PR #1240, opened 2022-03-08
Reload project-specific files on every Graudalizer diagnostic run

erszcz:
The check to determine if project files should be imported into Gradualizer comes from the time when the integration relied on Gradualizer being available through ERL_LIBS and would only be started by the diagnostic being called.

Now that Gradualizer is a dependency of ErlangLS, it's running on each startup since ErlangLS itself. This means the check never succeeds, therefore the project specific files necessary for properly setting up the typechecker are never loaded. This leads to false positives being reported such as missing remote types or record definitions.

Reimporting the project files on each run might be a bit redundant, but it seems to work reasonably well in practice. Dogfooding this solution shows it doesn't lead to visible unnecessary load on the system.

### DIFF
--- a/apps/els_lsp/src/els_gradualizer_diagnostics.erl
+++ b/apps/els_lsp/src/els_gradualizer_diagnostics.erl
@@ -32,16 +32,20 @@ is_default() ->
 
 -spec run(uri()) -> [els_diagnostics:diagnostic()].
 run(Uri) ->
-  case start_and_load() of
-    true ->
-      Path = unicode:characters_to_list(els_uri:path(Uri)),
-      Includes = [{i, I} || I <- els_config:get(include_paths)],
-      Opts = [return_errors] ++ Includes,
-      Errors = gradualizer:type_check_files([Path], Opts),
-      lists:flatmap(fun analyzer_error/1, Errors);
-    false ->
-      []
-  end.
+  Files = lists:flatmap(
+            fun(Dir) ->
+                filelib:wildcard(filename:join(Dir, "*.erl"))
+            end,
+            els_config:get(apps_paths) ++ els_config:get(deps_paths)),
+  ?LOG_INFO("Importing files into gradualizer_db: ~p", [Files]),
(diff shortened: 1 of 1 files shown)


robertoaloi:
Hi @erszcz and thanks for this. I am a bit concerned about the performance implications of running the import at every save. While this could be doable for small repository, it can easily create problems at scale.
In https://github.com/erlang-ls/erlang_ls/pull/1148 there's an attempt at introducing an `init` callback for diagnostics. Would it make sense to use it here, so that the import is only executed once?

robertoaloi (changes requested):
It would be great if we could add the `init` callback for diagnostics and only import files once.

erszcz:
Thanks for taking a look, @robertoaloi!

> In https://github.com/erlang-ls/erlang_ls/pull/1148 there's an attempt at introducing an init callback for diagnostics. Would it make sense to use it here, so that the import is only executed once?

I saw the `init` concept mentioned in some other issue and considered using it, but it would introduce another problem. If we only read the files on init, then types added or specs changed as we edit wouldn't be taken into consideration on subsequent runs of the diagnostic.

There's definitely some work on ErlangLS and Gradualizer interop, but just using `init` for the logic which used to be in `start_and_load` doesn't solve the problem - the original idea of `start_and_load` was simply wrong. Reloading the files on each run carries a performance penalty, but a more sophisticated mechanism would probably require hashing files on each diagnostic run to check for changes and reload as necessary.

If it's not a problem, I'd opt for merging this PR as it solves a nasty bug. I think the performance issue is secondary to that, though it certainly has to be revisited at some point.

robertoaloi:
@erszcz Considering that the Gradualizer integration is disabled by default, I don't have a problem with that.

Long term, Erlang LS should have support for `[DidChangeWatchedFiles](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles)` events. Once that is in place, we could revisit the diagnostic API to allow backends to register interest in changes.

Again, thanks for the contribution!

PR #1255, opened 2022-03-24
Skip indexing of generated files

robertoaloi:


apps/els_core/src/els_config_indexing.erl
```diff
@@ -0,0 +1,45 @@
...
+
+-spec default_generated_files_tag() -> string().
+default_generated_files_tag() ->
+  "@generated".
+
+-spec normalize_boolean(boolean() | string()) -> boolean().
+normalize_boolean(Value) when is_list(Value) ->
+  normalize_boolean(list_to_atom(Value));
+normalize_boolean(Value) when is_atom(Value) ->
+  Value.
```
michalmuskala, line 45: `list_to_atom` always makes be a bit uneasy due to DoS - not a risk for erlang_ls, but still 😅

Also, I think a stricter function, in case we get a non-boolean, will blow up earlier and be easier to debug

```suggestion
normalize_boolean("true") -> true;
normalize_boolean("false") -> false;
normalize_boolean(Bool) when is_boolean(Bool) -> Bool.
```

apps/els_lsp/src/els_indexing.erl
```diff
@@ -50,9 +50,33 @@ find_and_index_file(FileName) ->
 
 -spec index_file(binary()) -> {ok, uri()}.
 index_file(Path) ->
-  try_index_file(Path, 'deep'),
+  SkipGeneratedFiles = els_config_indexing:get_skip_generated_files(),
+  GeneratedFilesTag = els_config_indexing:get_generated_files_tag(),
+  try_index_file(Path, 'deep', SkipGeneratedFiles, GeneratedFilesTag),
```
michalmuskala, line 55: The two values are always passed in together - should they be made into just one accessor? Something like:
```
els_config_indexing:skip_generated_config()
```
which would return `disabled | {enabled, Tag}`

It should simplify the code and follow the "make impossible states unrepresentable" approach

robertoaloi: I like this a lot, but let's do this as a follow up. The main reason is that the whole config management needs refactoring, so it would be nice to come up with a common pattern.

apps/els_core/src/els_config_indexing.erl
robertoaloi: Applied the suggestion locally.

michalmuskala (approved):

PR #1260, opened 2022-04-01
On demand indexing

robertoaloi:
Historically, Erlang LS performed a full indexing of the workspace at startup. While this approach works reasonably well for a typical, small, Erlang project, it is problematic for mono-repos, where the initial indexing time can take several minutes. During these minutes, results for operations such as _find references_ or _rename_ may have unexpected, incorrect results.
Furthermore, containing the full indexed information in memory makes the whole server less responsive, especially on machines with limited RAM, and slows down operations.

This PR switches the indexing approach to an incremental one, which is more similar to what modern editors (most notably IntelliJ) tend to do.

During startup, Erlang LS will still performs an initial scan of all files belonging to the workspace but, instead of performing a full indexing for each file, it will only load into memory the minimum amount of information required to satisfy requests from clients.

The _Points of Interest_ (POIs) and other information are calculated on demand, every time they are requested. Once they are calculated, the POIs are stored in memory, so that subsequent requests are faster.

To provide answers to requests that rely on global information (most notably `getReferences`), we adopt the so-called _IDE trick_ (to use the wording from @michalmuskala who introduced the trick to me): first we do a quick scan of the code base to find potential candidates, then index those files and do a more accurate analysis.

This PR reduces both the initial indexing time and the memory footprint of Erlang LS significantly.

The process used to buffer incremental text edits was also rewritten as a `gen_server` and a few race conditions around text synchronization have been identified and solved, which should also lead to a more solid language server.


plux:
Seems interesting. 
I think it would be appropriate to add a config to toggle this behavior.

robertoaloi:
> Seems interesting. I think it would be appropriate to add a config to toggle this behavior.

Fair point. My only concern is that we may end up with effectively two versions of the indexing process, which could make things harder to maintain (and different users may experience subtle bugs when using one or the other). By having *one* way of doing things it would be much simpler. On the other hand, it would be a good way to try this out until it reveals a valid approach.

apps/els_lsp/src/els_dt_document.erl
```diff
@@ -145,18 +145,21 @@ new(Uri, Text) ->
 
 -spec new(uri(), binary(), atom(), kind()) -> item().
 new(Uri, Text, Id, Kind) ->
-  {ok, POIs} = els_parser:parse(Text),
-  MD5        = erlang:md5(Text),
+  MD5 = erlang:md5(Text),
   #{ uri  => Uri
    , id   => Id
    , kind => Kind
    , text => Text
```
gomoripeti, line 152: is it necessary to store the Text if the POIs are not (yet) stored? if Text is still stored it could be used to grep in the in-memory binary. If it's not stored it can spare some memory.

gomoripeti:
Is `Mode = shallow` still necessary if there will be ondemand "mode"? They might be orthogonal

robertoaloi:
> Is `Mode = shallow` still necessary if there will be ondemand "mode"? They might be orthogonal

Probably not. In the latest version I'm working on all indexing is "shallow".

apps/els_lsp/src/els_dt_document.erl
robertoaloi: Switched to an in-memory search (that can definitely be optimized) instead of the grep.

apps/els_lsp/src/els_dt_document.erl
```diff
@@ -201,3 +205,14 @@ wrapping_functions(Document, Line, Column) ->
 wrapping_functions(Document, Range) ->
   #{start := #{character := Character, line := Line}} = Range,
   wrapping_functions(Document, Line, Character).
+
+-spec find_candidates(binary()) -> [uri()].
+find_candidates(Pattern) ->
```
robertoaloi, line 210: Let's say there's room for improvements here...

gomoripeti: great start, though :)

apps/els_lsp/src/els_dt_references.erl
```diff
@@ -96,11 +98,24 @@ find_by_id(Kind, Id) ->
...
+     {ok, Document} = els_utils:lookup_document(Uri),
+     POIs = els_dt_document:pois(Document, [ application
+                                           , behaviour
+                                           , implicit_fun
+                                           , include
+                                           , include_lib
+                                           , type_application
+                                           , import_entry
+                                           ]),
+     [register_reference(Uri, POI) || POI <- POIs]
```
gomoripeti, line 113: do I understand correctly that every time I look up references of an entity, it will call `find_by` and try to insert a lot of references in the ref_db (ets entries might be there already, so some of them might be noop) Why do we store the references in ets then? can't we run the matching on the `POIs` list in place?
Alternatively can we mark Uris, which already registered their references, so we can skip lookup_pois+register_references for them?

apps/els_lsp/src/els_indexing.erl
```diff
@@ -253,11 +199,11 @@ index_dir(Dir, Mode, SkipGeneratedFiles, GeneratedFilesTag) ->
 
 -spec entries_apps() -> [{string(), 'deep' | 'shallow'}].
 entries_apps() ->
-  [{Dir, 'deep'} || Dir <- els_config:get(apps_paths)].
+  [{Dir, 'shallow'} || Dir <- els_config:get(apps_paths)].
```
gomoripeti, line 202: 👍

apps/els_lsp/src/els_dt_references.erl
robertoaloi: In my latest version (soon to be pushed), we only do this when expanding the POIs and enforcing a full indexing.

robertoaloi: My major concern here is that, depending on the pattern, we may end up indexing a lot of files. Need to find a good strategy to reduce the number of candidates significantly.

apps/els_lsp/src/els_dt_document.erl
```diff
@@ -204,3 +205,14 @@ wrapping_functions(Document, Line, Column) ->
...
+  All = ets:tab2list(name()),
+  Fun = fun(#els_dt_document{uri = Uri, text = Text}) ->
```
TheGeorge: Does that mean we read all files of the project into memory? What impact does this have on memory and indexing time?

How does `binary:match` compare to `grep` performance wise? (Seems it should be quicker since the file IO is already done)

robertoaloi: That's unchanged. The loading time is minimal and not a concern (I will try to get some actual numbers).
At this point, the `match` performance is not an issue, the part that takes time is the indexing of a potentially big number of candidates (eg if you search for callers to a function called "get").

robertoaloi: By implementing support for `WorkDoneProgress` we can at least report progress to the user and hopefully show intermediate results and be able to cancel a specific request.

apps/els_lsp/src/els_dt_document.erl
```diff
@@ -201,3 +214,24 @@ wrapping_functions(Document, Line, Column) ->
...
+                         , source = '$2'
+                         , kind = '_'
+                         , text = '$3'
+                         , md5 = '_'
+                         ,pois = '_'}
+        , [{'=/=', '$2', otp}]
+        , [{{'$1', '$3'}}]}],
+  All = ets:select(name(), MS),
+  Fun = fun({Uri, Text}) ->
+            case binary:matches(Text, Pattern) of
```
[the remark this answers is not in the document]
TheGeorge: The main issue I see is that if we are looking for references to something like
```
to() -> paris.
```
this matches to all kind of text and atoms since the there is no tokenisation going on
```
"string containing to"
atom_containing_to
```

I think doing some cheap tokenisation would cut down false positives significantly for these cases.

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -44,8 +44,17 @@ handle_request({completion, Params}, State) ->
...
+  {ok, Document} = els_utils:lookup_document(Uri),
+  #{text := Text, buffer := Buffer} = Document,
+  %% Ensure there are no pending changes. This causes a heavy sync
+  %% point, since for a big module indexing could take a while.  Maybe
+  %% just do a 'soft' flush and avoid indexing here?
+  case Buffer of
+    undefined ->
+      ok;
+    _ ->
+      ok = els_buffer_server:flush(Buffer)
```
TheGeorge: I wonder if we cannot be a bit sloppy here and to make flushing delayed. Looking at the LSP, I see that we can return partial completion results. So what we can do to increase responsiveness is this:
- if no pending changes, just what we do now
- i pending changes, trigger flushing, and return completion results based on current knowledge
- when indexing is complete return full completion results

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -44,8 +44,18 @@ handle_request({completion, Params}, State) ->
                             }
    , <<"textDocument">> := #{<<"uri">> := Uri}
    } = Params,
-  ok = els_index_buffer:flush(Uri),
```
robertoaloi, line 47: A bit of context may help here.

Completion requests typically come at the same time as edits. With the current implementation of the `els_index_buffer` it could happen that, during a completion, edits are still not processed (there's some basic protection against this, but things can still fail).

The completion provider assumes a specific length in the line when reading the text after the flush, but it's not guaranteed that all edits are applied. Fast typers can see the provider (and eventually the whole server) crashing.

With the new approach, the idea is to block callers of the flush for a certain amount of time (right now 200ms), but if edits keep coming the count restarts.

This should also prevent indexing happening multiple times.

apps/els_lsp/src/els_dt_document.erl
```diff
@@ -223,25 +231,42 @@ wrapping_functions(Document, Range) ->
...
+  case erl_scan:string(els_utils:to_list(Text)) of
+    {ok, Tokens, _EndLocation} ->
+      Fun = fun({atom, _Location, Atom}, Words) ->
+                sets:add_element(Atom, Words);
+               ({string, _Location, String}, Words) ->
+                sets:add_element(String, Words);
+               (_, Words) ->
+                Words
+            end,
+      lists:foldl(Fun, sets:new(), Tokens);
```
michalmuskala, line 269: `sets` in general is pretty slow and uses a lot of memory, unless you use `sets:new([{version, 2}])` for the map-based representation.

That said, in general it seems like shipping this data back and forth from ets might be really expensive.

An alternative would be to use a plain map to store the data: `#{atom() | string() => true}`

And query it with a match spec like:
```erl
  MS = [{#els_dt_document{ uri = '$1'
                         , source = '$2'
                         , buffer = '_'
                         , kind = '_'
                         , text = '_'
                         , md5 = '_'
                         , pois = '_'
                         , words = '$3'
                         }
        , [{'=/=', '$2', otp}, {'is_map_key', Search, '$3'}]
        , ['$1']}],
```

This way there's no need to copy the data out of ETS

robertoaloi: Version 2 is only supported from OTP 24, right? Ideally Erlang LS shouldn't care about the project version used by a user, but for the time being we have some assumptions in the code base that force us to be backward compatible. So, will need to add it as an ifdef.

robertoaloi: Thanks for this! I will leave this as a follow up, so that we can actually measure the eventual improvement.

apps/els_lsp/src/els_completion_provider.erl
robertoaloi: Partial results are not yet implemented, I will add this as a follow up.

robertoaloi:
> Seems interesting. I think it would be appropriate to add a config to toggle this behavior.

@plux I attempted at making this configurable, but then realized that a bunch of optimizations have been dropped as part of the change (eg. the deleted_by_uri). By making the whole feature configurable, we would need to reintroduce them for the non-incremenental indexing (they don't really make sense with the new setup), which makes the code quite hairy. I am leaning towards just switching the behaviour and holding from a new release until this proves solid enough. That would allow us to dog-fooding the new code base. And from the first experiments, it looks promising.

apps/els_lsp/src/els_text_search.erl
```diff
@@ -0,0 +1,46 @@
...
+extract_pattern({type, {_M, F, _A}}) ->
+  F;
+extract_pattern({macro, {Name, _Arity}}) ->
+  Name;
+extract_pattern({macro, Name}) ->
+  Name;
+extract_pattern({include, Id}) ->
+  include_id(Id);
+extract_pattern({include_lib, Id}) ->
+  include_id(Id);
```
michalmuskala, line 40: Hmm... It looks like in the `word` splitting we save full strings and then compare for equality. But here we only extract the rootname. 

I think this means that include matching would never actually match since we have indexed strings like `"foo.hrl"`, but we're looking here for `"foo"`. Similar for `include_lib` we'd index `"app/include/foo.hrl"`

michalmuskala (commented):
Some comments from me.

Some of them concern the feature itself, some are to clarify the behaviour of ELS, since I'm not fully familiar with it

apps/els_lsp/priv/code_navigation/src/purge_references.erl
```diff
@@ -0,0 +1,5 @@
+-module(purge_references).
+
+-spec foo(any()) -> ok.
+foo(_X) -> ok.
+bar() -> foo(42).
```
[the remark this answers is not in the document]
robertoaloi: Nice catch! I started refactoring the purge references, before deleting the whole test case and this was a left over. Removed.

apps/els_lsp/src/els_buffer_server.erl
```diff
@@ -0,0 +1,126 @@
...
+        , apply_edits/2
+        , flush/1
+        ]).
+
+-export([ start_link/2 ]).
+
+%%==============================================================================
+%% Callbacks for the gen_server behaviour
+%%==============================================================================
+-behaviour(gen_server).
```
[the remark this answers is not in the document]
robertoaloi: The buffer was introduced as part of [support for incremental text sync](https://github.com/erlang-ls/erlang_ls/pull/1002) and I just ported it to a `gen_server`, so it's not new with this feature.

The main intention was to buffer `didChange` events, so that re-indexing would be retriggered not too often (a didChange event is sent at every keystroke). But I agree that it complicates things and probably introduces some races.

apps/els_lsp/src/els_text_search.erl
robertoaloi: That is correct, need to fix.

robertoaloi: Actually, we don't even need to index strings, with the exceptions of includes, so I may apply the same transformation on indexing.

robertoaloi: Fixed.

apps/els_core/src/els_utils.erl
```diff
@@ -150,13 +150,13 @@ lookup_document(Uri) ->
       {ok, Document};
     {ok, []} ->
       Path = els_uri:path(Uri),
-      {ok, Uri} = els_indexing:index_file(Path),
+      {ok, Uri} = els_indexing:shallow_index(Path, app),
```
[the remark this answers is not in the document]
robertoaloi: The `Source` (`app`) is only used to decide whether to index references or not. In principle, we should be able to detect from the `Uri` if it refers to a `app`, `otp` or `dep`. In practice, we just store references even for `otp` modules, which shouldn't have a significant effect for the user. But it's a fair question, so I will add a comment to eventually follow up.

apps/els_lsp/src/els_dt_signatures.erl
```diff
@@ -72,11 +73,18 @@ insert(Map) when is_map(Map) ->
...
   {ok, [to_item(Item) || Item <- Items]}.
 
--spec delete_by_module(atom()) -> ok.
-delete_by_module(Module) ->
-  Pattern = #els_dt_signatures{mfa = {Module, '_', '_'}, _ = '_'},
-  ok = els_db:match_delete(name(), Pattern).
+-spec delete_by_uri(uri()) -> ok.
+delete_by_uri(Uri) ->
+  case filename:extension(Uri) of
+    <<".erl">> ->
```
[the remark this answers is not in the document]
robertoaloi: Signatures are currently indexed by `MFA` and the respective info is indexed while the module itself is indexed (taking into account the chain of header files). The reason why the check is here is that we also process headers for indexing and that caused an issue since it was deleting all, say, `file` specs when indexing `file.hrl`, which is of course incorrect. Indexing of headers should be reconsidered, but I would keep that out of scope for this PR.

apps/els_lsp/src/els_indexing.erl
```diff
@@ -72,66 +56,84 @@ is_generated_file(Text, Tag) ->
...
-  MD5 = erlang:md5(Text),
-  case els_dt_document:lookup(Uri) of
-    {ok, [#{md5 := MD5}]} ->
-      ok;
-    {ok, LookupResult} ->
-      Document = els_dt_document:new(Uri, Text),
-      do_index(Document, Mode, LookupResult =/= [])
+-spec ensure_deeply_indexed(uri()) -> ok.
+ensure_deeply_indexed(Uri) ->
+  {ok, #{pois := POIs} = Document} = els_utils:lookup_document(Uri),
```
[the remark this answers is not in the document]
robertoaloi: That sounds like a great optimization that can be implemented as a follow up.

apps/els_lsp/src/els_indexing.erl
```diff
@@ -72,66 +56,84 @@ is_generated_file(Text, Tag) ->
...
-  %% Mapping from document id to uri
-  ModuleItem = els_dt_document_index:new(Id, Uri, Kind),
-  ok = els_dt_document_index:insert(ModuleItem),
-  index_signatures(Document),
-  index_references(Document, Mode, Reset).
+-spec deep_index(els_dt_document:item()) -> ok.
+deep_index(Document) ->
+  #{id := Id, uri := Uri, text := Text, source := Source} = Document,
+  {ok, POIs} = els_parser:parse(Text),
+  ok = els_dt_document:insert(Document#{pois => POIs}),
```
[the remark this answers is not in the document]
robertoaloi: Yes, that could happen (even if that would require two `didOpen` for the same file, which is not very likely). Said that, I'm considering to remove the buffer altogether and find an alternative solution for avoiding multiple re-index operations.

robertoaloi:
@michalmuskala For some reason I cannot reply inline to this comment:

https://github.com/erlang-ls/erlang_ls/pull/1260#discussion_r845939961

That should be handled correctly, but it would be great to encode that in a test case.

michalmuskala (approved):
With the clarifying comments and fixes, I think this looks pretty good to go for me

PR #1261, opened 2022-04-04
add completion for clauses

TheGeorge:
### Description

Adds completion support for clauses. The trigger is `;` ~~and it currently only support top-level functions. What's missing is completion for:~~
- ~~funs~~
- ~~receives~~
- ~~try / catch~~
- ~~case~~

## Example

Screenshot 2022-04-04 at 15 54 19

results into 

Screenshot 2022-04-04 at 15 54 26

### DIFF
--- a/apps/els_lsp/src/els_completion_provider.erl
+++ b/apps/els_lsp/src/els_completion_provider.erl
@@ -571,15 +571,14 @@ clause(Document, Line, Column) ->
           {ok, Tokens, _Comments, _Cont} = erlfmt_scan:string_node(ClauseText),
 
           %% return completion
-          I = case clause_completion_type(Tokens) of
-                function ->
-                  completion_item(POI#{kind => function_clause}, #{}, false);
-                'fun' ->
-                  clause_completion_snipppets(fun_clause);
-                other ->
-                  clause_completion_snipppets(other_clause)
-              end,
-          [I];
+          case clause_completion_type(Tokens) of
+            function ->
+              [completion_item(POI#{kind => function_clause}, #{}, false)];
+            'fun' ->
+              clause_completion_snipppets(fun_clause);
+            other ->
+              clause_completion_snipppets(other_clause)
+          end;
         _ ->
           []
       end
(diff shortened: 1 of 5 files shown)


michalmuskala:
I think the behaviour I'd prefer is that on the newline when I start typing the name of the function, it offers to complete, so rather something like:
```
foo(A, B, C) -> ok;
f|
```
complete shows up, and fills:
```
foo(A, B, C) -> ok;
foo(A, B, C) -> |
```

This should work in more cases, including when you're adding a clause in the middle of a function, and not just when first writing out the clauses top-down

TheGeorge:
Interesting angle, having it trigger on the `;` is inspired from how emacs completes clauses. I somewhat disagree with suggestion, because extending from the last `;` (besides emacs) is a workflow that is quite common, e.g. markdown editors, other word processors, or even extending comments. So it's a pattern most people are used to already.

michalmuskala:
My reasoning is that this is in general how completion works - you start typing the name, and it's completed for you. Doing this for clauses seems very natural to me.

Actually, it seems this already works today:

https://user-images.githubusercontent.com/477062/161929103-37780234-e462-4180-99aa-64f1a1bcd96e.mov

Re: adding it in the middle - does this trigger when you just "stand" on a `;` vs when you just typed it in?

TheGeorge:
> Re: adding it in the middle - does this trigger when you just "stand" on a ; vs when you just typed it in?

The LSP has 3 trigger types:
- character (typing `;`)
- invoked (ctrl + SPACE)
- for incomplete completion (when completion list was not complete due to partial result returned)

The one in the middle I triggered with ctrl+SPACE

TheGeorge:
https://user-images.githubusercontent.com/292239/164294104-da5a8aea-86c5-45dc-8fbe-7d015bf42ade.mov

This now supports completion of arbitrary clauses and is context sensitive

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -28,7 +28,7 @@
 %%==============================================================================
 -spec trigger_characters() -> [binary()].
 trigger_characters() ->
-  [<<":">>, <<"#">>, <<"?">>, <<".">>, <<"-">>, <<"\"">>].
+  [<<":">>, <<"#">>, <<"?">>, <<".">>, <<"-">>, <<"\"">>, <<";">>, <<"\n">>].
```
robertoaloi, line 31: Is the `\n` needed?

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -147,15 +147,26 @@ find_completions( Prefix
...
+                  }) ->
+  clause(Document, Line, Column);
 find_completions( Prefix
                , ?COMPLETION_TRIGGER_KIND_INVOKED
                , #{ document := Document
                   , line     := Line
                   , column   := Column
                   }
                ) ->
+  % ?LOG_WARNING("Triggered Completing ~p", [Prefix]),
```
robertoaloi, line 165: Leftover?

apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -538,6 +549,139 @@ exported_definitions(Module, POIKind, ExportFormat) ->
...
   end.
 
+%%==============================================================================
+%% Clauses
+%%==============================================================================
+
+-spec clause(els_dt_document:item(), line(), column()) -> [map()].
+clause(Document, Line, Column) ->
+  try
+  case els_dt_document:pois(Document, [function, function_clause]) of
```
robertoaloi, line 559: We should be able to only keep the try and remove the case?

robertoaloi:
@TheGeorge I tried the feature in both VS Code and Emacs. It looks good in VS Code and I don't get it for some reason in Emacs (which is good, since I prefer the Emacs version). So, no concerns from my side.

apps/els_lsp/src/els_completion_provider.erl
TheGeorge: no, it can be removed

apps/els_lsp/src/els_completion_provider.erl
TheGeorge: Yes, let me clean this up

TheGeorge:
Added tests and Ci should now work after windows CI fix.

robertoaloi:
@TheGeorge Should we try and get this one in? There are some unresolved conflicts.

TheGeorge:
Yeah, just need to rebase after formatting

TheGeorge:
dropped for now

PR #1264, opened 2022-04-12
Single Provider Process

robertoaloi:
Erlang LS historically used one process per provider. This was
problematic, since requests handled by different providers could lead
to race conditions, causing a server crash. For example, a completion
request could not see the latest version of the text in case of
incremental text synchronization.

There was very little gain in parallelizing providers, so we decided to
simplify the architecture and have a single process, instead. There's
still room for refactoring and cleanup, but those will be handled as a
follow up.

As part of the architectural change, we also dropped support for BSP
since it has been stalling for too long time. Instead, we will try to
implement a simpler solution which allows the extraction of paths from
the build system (e.g. via the rebar.conifg).

* Drop BSP Support
* Introduce text synchronization provider
* Run providers as a single process
* Remove buffer server, index in the background instead
* Decouple DAP provider from LSP one
* Brutally kill background jobs
* Set standard_io as default value for the io_device


apps/els_core/src/els_provider.erl
```diff
@@ -43,78 +43,134 @@
...
-handle_call({handle_request, Request}, _From, State) ->
-  #{internal_state := InternalState, provider := Provider} = State,
-  {Reply, NewInternalState} = Provider:handle_request(Request, InternalState),
-  {reply, Reply, State#{internal_state => NewInternalState}}.
+handle_call({handle_request, Provider, Request}, _From, State) ->
+  #{in_progress := InProgress, in_progress_diagnostics := InProgressDiagnostics}
+    = State,
+  case Provider:handle_request(Request, State) of
+    {async, Uri, Job} ->
+      {reply, {async, Job}, State#{in_progress => [{Uri, Job}|InProgress]}};
```
michalmuskala, line 105: Is there some monitoring of the jobs to remove them from the map when they are externally terminated/crash?

apps/els_lsp/src/els_diagnostics_provider.erl
```diff
@@ -2,90 +2,44 @@
...
-  {noresponse, State#{in_progress => [Entry|InProgress]}}.
+  {diagnostics, Uri, Jobs}.
 
 %%==============================================================================
 %% API
 %%==============================================================================
 -spec notify([els_diagnostics:diagnostic()], pid()) -> ok.
 notify(Diagnostics, Job) ->
-  ?SERVER ! {diagnostics, Diagnostics, Job},
+  els_provider ! {diagnostics, Diagnostics, Job},
```
michalmuskala, line 42: I wonder if those notifications should be somehow incorporated into the background job framework - e.g. that it sends as a reply to the process that started the job the value that was returned from the function (kind of similar to Elixir's `Task`)

michalmuskala (approved):
Overall this looks fairly reasonable, just some minor comments.

On the bigger design decisions:
* dropping BSP seems to me like a sensible choice - we don't really have plans to work on that, and could instead integrate with rebar3 more directly
* making the processing less "async" seems like a good choice - main event loop + async processes for specific things seems like a simpler programming model, than every type of request running in parallel

apps/els_core/src/els_provider.erl
robertoaloi: Not that I can see, even if the consequence of leaving an orphan entry in the map shouldn't be critical. I will create a ticket about this. Also, I now realize that there's some logic running in the `terminate` function (to clean up the spinning wheel), so I will roll back the `brutal_kill` strategy bit.

robertoaloi: #1271

apps/els_lsp/src/els_diagnostics_provider.erl
robertoaloi: I'm not familiar with Elixir's `Task` but it sounds like a good idea. Created #1272

PR #1265, opened 2022-04-12
Document versioning

robertoaloi:
This PR introduce _versioning_ for documents, signatures and references. Versioning is available as part of the LSP specification. Versioning is currently supported by all major LSP clients (Emacs, VS Code, etc).

Whenever a module is indexed for the first time, the version is originally set to `null`. This allows the server to function even with LSP clients not specifying versions on change (but they will be exposed to all sorts of race conditions).

LSP clients, whenever sending a _change_ event for a document (`didChange`, `didChangeWatchedFile`), they normally include a version, which is guaranteed to be strictly incrementing (even if there can be gaps between two subsequent versions).

Whenever a change is detected, the language server immediately stores the new version of the text in the DB and triggers a new indexing process in the background. Subsequent changes will cause existing jobs to abort, to avoid un-necessary indexing of files which already changed and triggered a new process.

On completion, the asynchronous indexing process will verify if the current version has changed in the meantime and discard its own results if that's the case. Notice that the background process will confirm its write to DB in three cases:

* If the current version in the DB is `null` (the document has never changed)
* If the current version in the DB is older
* If the version is the same (text is already stored, but the current process has been parsing POIs, references and signatures for exactly that version)

Words (used to reduce the number of candidates for finding references) are also recomputed when storing the changed information.

Since ETS does not support transactions, writes to the DB are serialized via a single process (the `els_db_server` _gen_server_). The _lookup-and-insert_ is performed within the same callback function to avoid a clear race condition.
While the current implementation is not ideal and will most likely be improved over time, it still represents an improvement compared to the current situation, where race conditions are extremely common during completion and on text change, especially for _fast typers_.

The `MD5` property of the document has been removed since not used any longer.

### DIFF
--- a/apps/els_lsp/src/els_db_server.erl
+++ b/apps/els_lsp/src/els_db_server.erl
@@ -2,7 +2,6 @@
 %%% @doc The db gen_server.
 %%% @end
 %%%=============================================================================
-
 -module(els_db_server).
 
 %%==============================================================================
@@ -13,8 +12,15 @@
         , delete/2
         , delete_object/2
         , match_delete/2
+        , select_delete/2
         , write/2
+        , conditional_write/4
         ]).
+
+%%==============================================================================
+%% Includes
+%%==============================================================================
+-include_lib("kernel/include/logger.hrl").
 
 %%==============================================================================
 %% Callbacks for the gen_server behaviour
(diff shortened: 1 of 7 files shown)


apps/els_lsp/src/els_db_server.erl
```diff
@@ -83,7 +93,24 @@ handle_call({match_delete, Table, Pattern}, _From, State) ->
...
+        false ->
+          ?LOG_DEBUG("Skip insertion due to invalid condition "
+                     "[table=~p] [key=~p]",
+                     [Table, Key]),
+          {reply, {error, condition_not_satisfied}, State}
+      end;
+    [] ->
+      true = ets:insert(Table, Object),
+      {reply, ok, State}
+  end.
```
michalmuskala, line 113: I wonder, if this could use ETS's `select_replace` instead for atomic writes directly in ETS. This still has a small chance of race conditions between when you look up the value and write it - the underlying database could have changed.

robertoaloi: Considering all writes to the ETS are serialized via a gen_server, this should not be a problem in practice.

robertoaloi:
> TBH I'm not sure this fixes all the possible cases. In particular - what are the situation when the version is bumped? What operations use the unversioned bump and what operations on the `Document` modify it synchronously vs asynchronously.
> 
> I think having this documented somewhere, would make it easier to understand, if the new behaviour will be actually correct

Considering this PR solves a very common crash which happens during completions, I am leaning towards merging this and then following up on documentation and addressing the other possible race conditions.

Issue #1209, opened 2022-02-17
Add folding support for records


robertoaloi:
### Context

[Folding](https://code.visualstudio.com/docs/editor/codebasics#_folding) is the ability to show/hide (collapse/uncollapse) regions of code. Many editors offer to the user the ability to fold and unfold regions at a specific level and/or recursively, using either a mouse or a keyboard shortcut. Folding can simplify reading complex code by allowing the user to focus on the significant parts only.

Today Erlang LS offers only some primitive folding capabilities, limited to functions, as illustrated below:

Before folding:
<img width="600" alt="Screenshot 2022-02-17 at 10 22 52" src="https://user-images.githubusercontent.com/91769/154445194-3dbeb33f-f338-4bec-af34-a11921e5e9ec.png">

After folding:
<img width="583" alt="Screenshot 2022-02-17 at 10 23 09" src="https://user-images.githubusercontent.com/91769/154445201-9c3b2e26-b846-495e-b8bf-1a83c3eb9041.png">

Whenever a folding  action is triggered in the IDE, a [textDocument/FoldingRange](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#textDocument_foldingRange) request is sent to the language server:

```
[Trace - 10:21:55 AM] Sending request 'textDocument/foldingRange - (6060)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/robertoaloi/git/github/erlang-ls/erlang_ls/apps/els_lsp/src/els_hover_provider.erl"
  }
}
```

Notice that the request only contains the _URI_ of the current file and no additional information. The language server, using the information stored in its database of _points of interest_ returns a list of **all** available folding ranges for the requested document (ignore the `999999`s for the purpose of this task):

```
[Trace - 10:21:55 AM] Received response 'textDocument/foldingRange - (6060)' in 1ms.
Result: [
  {
    "startLine": 30,
    "startCharacter": 9999999,
    "endLine": 31,
    "endCharacter": 9999999
  },
  {
    "startLine": 34,
    "startCharacter": 9999999,
    "endLine": 35,
    "endCharacter": 9999999
  },
  [...]
]
```

[...]

PR #1266, opened 2022-04-12
Folding for records

AminArria:
Issue #1209 

This PR is meant to improve folding of code by removing the need of `folding_range` PIOs and instead having it as a field in `data` for the others PIOs.

Later in `els_folding_range_provider` we can lookup the specific kind of PIOs we want to enable folding for, for now `function` and `record`


robertoaloi:
Hi @AminArria ! I didn't look at this until now since it was marked as a draft and had a red CI, but I now realize that this is 100% equivalent to #1268, which I just merged. Will add a comment to that PR to ensure credit is shared.

PR #1268, opened 2022-04-14
Refactor folding_ranges onto POIs, Add support for records.

tks2103:
Moving folding_range from functions to a general field on each poi. Added support for records. Refactored support for functions. Added foldingrange_provider test

### DIFF
--- a/apps/els_lsp/priv/code_navigation/src/folding_ranges.erl
+++ b/apps/els_lsp/priv/code_navigation/src/folding_ranges.erl
@@ -1,0 +1,11 @@
+-module(folding_ranges).
+
+function_foldable() ->
+  ?assertEqual(2.0, 4/2).
+
+-record(unfoldable_record, { field_a }).
+
+-record(foldable_record, { field_a
+                         , field_b
+                         , field_c
+                        }).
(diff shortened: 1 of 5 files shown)


robertoaloi (approved):

robertoaloi:
Thanks a lot for this @tks2103 ! Great generalization!

CI is failing, but I believe it's because many tests depend on the content of the `code_navigation` test module. We could either adjust the expectations or have a specific test file for folding ranges and point to that in the test suite.

apps/els_lsp/priv/code_navigation/src/folding_ranges.erl
```diff
@@ -0,0 +1,11 @@
+-module(code_navigation).
```
tks2103, line 1: oops, sec

tks2103:
@robertoaloi is this passing CI now? Can't figure out if it ran another set of tests...

I don't have commit access, feel free to merge if it's good.

robertoaloi:
> @robertoaloi is this passing CI now? Can't figure out if it ran another set of tests...

Since this is your first contribution to the project, GitHub requires me to manually approve the CI run. Triggered it now, so let's see. You can also run this manually on your machine, following the steps at:

https://github.com/erlang-ls/erlang_ls/blob/main/.github/workflows/build.yml

So, essentially, you can:

    rebar3 do compile, ct, lint, proper, dialyzer, xref, edoc

robertoaloi:
Thanks!

robertoaloi:
And credit should also go to @AminArria who provided the original implementation in #1266.

PR #1277, opened 2022-04-22
Make edoc an optional dependency

filmor:
Some distributions may not include it, and as it's not installed via hex,
startup will fail.

### DIFF
--- a/apps/els_lsp/src/els_lsp.app.src
+++ b/apps/els_lsp/src/els_lsp.app.src
@@ -18,6 +18,9 @@
         els_core,
         gradualizer
     ]},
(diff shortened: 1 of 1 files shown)


robertoaloi:
Hi @filmor and thanks. Won't this causes some runtime errors if the application is not available? If we make the application optional, we should probably ensure those use cases degrade gracefully (eg not to show docs).

filmor:
It does seem to degrade half-way gracefully, it doesn't take down all of erlang-ls, just the `edoc` calls are crashing. Definitely much better than the current state of things for me :)

PR #1282, opened 2022-04-27
Allow user to provide custom PREFIX

sidkshatriya:
Allow user to provide custom PREFIX

Example:

```bash
$ PREFIX=~/.local make install
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 .PHONY: all
 
-PREFIX = '/usr/local'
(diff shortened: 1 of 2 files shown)


robertoaloi:
Hi @sidkshatriya  and thanks. Do you mind updating the README which suggests to run `make` with the `-e` option? That shouldn't be necessary any longer after your change.

sidkshatriya:
Done!

PR #1287, opened 2022-05-05
Solidify provider process

robertoaloi:
Since the main supervision strategy is rest_for_one, move the els_provider
process after the els_server one (restarting the els_server process currently
results in a node crash anyway). There is no need to restart the els_server
in case of els_provider failure, especially considering that the provider
process is the most likely to fail.

Also, give the provider a chance to cleanup on restart, by cancelling
pending jobs.

Finally, since it can now happen that, after a restart, the provider
receives messages (results or diagnostics) by background jobs, ensure
that such responses are handled correctly: results are propagated to
the server (no change) while diagnostics are ignored.

### DIFF
--- a/apps/els_core/src/els_provider.erl
+++ b/apps/els_core/src/els_provider.erl
@@ -13,6 +13,7 @@
         , handle_call/3
         , handle_cast/2
         , handle_info/2
+        , terminate/2
         ]).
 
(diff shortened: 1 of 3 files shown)


michalmuskala (approved):
Makes sense

PR #1297, opened 2022-05-12
Apply erlfmt to the entire codebase

robertoaloi:
* Format the code base using [whatsapp/erlfmt](https://github.com/WhatsApp/erlfmt)
* Add a CI step to ensure code formatting is respected
* Do not format files included in `priv` or `_SUITE_test_data` to avoid un-necessary tweaking of test cases
* Increase max line length in Elvis to 100, to avoid conflict with formatter

Later on, we could improve CI to not just check, but actually format and amend PRs on merge.

### DIFF
--- a/rebar.config
+++ b/rebar.config
@@ -1,89 +1,100 @@
-{erl_opts, [ debug_info
-           , warnings_as_errors
-           , warn_export_vars
-           , warn_unused_import
-           , warn_missing_spec_all
-           ]
-}.
+{erl_opts, [
+    debug_info,
+    warnings_as_errors,
+    warn_export_vars,
+    warn_unused_import,
+    warn_missing_spec_all
+]}.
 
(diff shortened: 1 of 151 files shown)


rebar.config
```diff
@@ -87,3 +89,8 @@
 %% Disable warning_as_errors for redbug to avoid deprecation warnings.
 {overrides, [ {del, redbug, [{erl_opts, [warnings_as_errors]}]}
             ]}.
+
+{erlfmt, [
+    write,
+    {files, "apps/*/{src,include,test}/*.{erl,hrl,app.src}"}
```
michalmuskala, line 95:
```suggestion
    {files, ["apps/*/{src,include,test}/*.{erl,hrl,app.src}", "rebar.config", "elvis.config"]},
```

robertoaloi: Right, I planned to do it but then forgot. Thanks.

PR #1301, opened 2022-05-15
Improve create undefined function quick action

### LINKED ISSUE #1280: The "create function" code action can produce code in the incorrect place.
The code action takes the end location of the _last_ POI. This is not guaranteed the one with the widest range (e.g. it could be an atom within the last function). Instead, we could consider the folding range of the last "function" POI.

plux:
### Description

* Create function right after current function
* Create function with correct number of arguments
* Don't leave trailing whitespaces
* Don't create -spec

Fixes #1280

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,7 @@ end_of_line = LF
 
 [*]
 indent_style = space
-indent_size = 2
+indent_size = 4
 trim_trailing_whitespace = true
 insert_final_newline = true
 max_line_length = 80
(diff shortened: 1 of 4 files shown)


robertoaloi:
Nice, thanks. Why not to leave the specs (e.g. as `any()`)?

plux:
None of the codebases I've worked with (except erlang_ls..) enforces -specs on non-exported functions, so adding the -spec is just causing clutter in my opinion.
Adding an incomplete -spec seems pointless to me, and we already have the suggest spec feature which can add a specs
I guess we could add a code action to add missing spec if there is such a diagnostic.

I could add it back if you want to, but my preference is to not add it.

michalmuskala:
Adding a spec with just `any()` would be actually worse than no spec at all. AFAIK Dialyzer reports worse errors in cases like that since it interferes with its inference logic.

robertoaloi:
Fair enough, let's get this one as-is. Yes, the quick action would actually be amazing.

PR #1303, opened 2022-05-16
Cleanup Points of Interest

robertoaloi:
This PR is the first of a series that aims at cleaning up the code around Points of Interest in Erlang LS.
The main problem I'm trying to solve is that the knowledge about the structure of each POI (eg a function)
is spread across the code base. A specification for the POIs does not really exist and bugs can occur in many places due to undocumented assumptions.

The goal of the cleanup is to:

* Create a formal specification (with type definitions) for the POIs (eg around ranges)
* Introduce a POI behaviour that can be implemented by individual POIs
* Create getters to extract information about POIs in a more structured manner.

In this episode:

* Move `poi/0` definition from header to module
* Introduce `els_poi` behaviour, implement `label/1` function for some of the pois
* Move symbol transformation functions to `els_poi`
* Move `folding_range` to `els_poi`
* Remove remains from `folding_range` poi_kind


gomoripeti:
nice initiative!

how will formal specification via type definitions for the POIs look like? eg will there be a `<els_poi_module>:poi()` type that defines exactly what fields that specific poi can have?
(if poi would be a record, than callback modules could define more specific types for them eg
```
els_poi.erl

-record(poi, {id :: atom() | string | ...}
-type poi() :: #poi{}.
%% OR
-type poi() :: els_poi_function:poi() | els_poi_define:poi() | ...
-export_type([poi/0]).
```
```
els_poi_function.erl

-type poi() :: #poi{id = id()}.
-export_type([poi/0]).
```

robertoaloi:
@gomoripeti That's exactly the idea. We could even use `-opaque` to ensure other parts of the code do not make assumptions on the internal structure of a `poi()`.

PR #1307, opened 2022-05-19
Implement textDocument/signatureHelp

### LINKED ISSUE #1300: implement textDocument/signatureHelp

Signature help is a very cool feature from the LSP spec. It provides documentation as you write the arguments of a function call and highlights the active parameter as you type.

It looks like this would need a new provider module `els_signature_help_provider` which could re-use some of the machinery from `els_completion_provider` to determine docs. I think we can determine the active signature by scanning backwards in the `els_text:tokens/1` potentially across multiple `els_text:line/3`s (function calls may spread across lines unlike completions).

A good first iteration for this would be to add signature help for functions but it could be extended to support macros and types as well.

Snippets could be seen as competing with signature help, but IMO the two features complement each other since signature help keeps the documentation for the function call open. My editor doesn't support snippets yet and I think it might be easier to implement signature help here 😅

**Additional context**

`textDocument/signatureHelp` in the spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp

An example with `rust-analyzer` from [helix-editor/helix#1755](https://github.com/helix-editor/helix/pull/1755):

command

With ElixirLS:

[asciicast](https://asciinema.org/a/494420)

the-mikedavis:
### Description

Implements [signature help](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_signatureHelp) from the LSP spec.

For this PR it's just remote and local functions - BIFs, macros, or types aren't included yet.

The approach for finding the signature (if it exists) when a trigger is emitted is to scan backwards through the tokens that preceed the trigger and look for an unmatched opening parenthesis. Also in this recursive scanning function, we determine the active parameter number by counting the number of commas. It's not that straightforward though: you might have expressions in your calls that themselves contain commas like `maps:put(key, value, #{a => b, c => d})`. To count the active parameter number, we use a stack of counters (implemented as `list(non_neg_integer())`) of how many times we see a comma. We push a new counter when we see a closing token (`>>`, `]`, `)`, `}`) and pop when we see an opening token (`<<`, `[`, `(`, `{`), so commas within expressions are counted but discarded when we scan past the data structure (remember that the tokens are in reverse order).

This works pretty well but there are edge cases like that the signature help pane will close (language server returns `null`) if you type a comma within a string, for example: `"foo, `. I haven't encountered any false positives where the server shows signature help when it shouldn't yet.

An asciicast of the signature help working in helix: https://asciinema.org/a/495869

Fixes #1300.

### DIFF
--- a/apps/els_core/include/els_core.hrl
+++ b/apps/els_core/include/els_core.hrl
@@ -540,17 +540,17 @@
 %%------------------------------------------------------------------------------
 -type parameter_information() :: #{
     label := binary(),
-    documentation => binary()
+    documentation => markup_content()
 }.
 -type signature_information() :: #{
     label := binary(),
-    documentation => binary(),
+    documentation => markup_content(),
     parameters => [parameter_information()]
 }.
 -type signature_help() :: #{
     signatures := [signature_information()],
-    active_signature => number(),
-    active_parameters => number()
+    activeSignature => non_neg_integer(),
+    activeParameter => non_neg_integer()
 }.
 
 %%------------------------------------------------------------------------------
--- a/apps/els_core/src/els_client.erl
+++ b/apps/els_core/src/els_client.erl
@@ -26,6 +26,7 @@
     '$_unexpectedrequest'/0,
     completion/5,
     completionitem_resolve/1,
+    signature_help/3,
     definition/3,
     did_open/4,
     did_save/1,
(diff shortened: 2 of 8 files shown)


the-mikedavis:
There are a few more improvements other than expanding the kinds of code this work on (types, macros):

* Signature help can be triggered manually (i.e. by some keybind in the editor), so ideally `find_signature/4` should not return `null` when the trigger character is not `(` or `,`. I think it would be a small change to remove that assumption so I'll convert this to a draft while I work on that.
* The client may pass the list of signatures back to the server when re-triggering signature-help. We could use this to be a little more efficient and deduplicate the work of generating documentation between requests for the same signature. I think this could be left to a future PR - I haven't seen any performance problems with this feature yet, it seems to be fairly instantaneous.

robertoaloi:
This is an amazing contribution, thanks! I wonder if we could bring this in as "disabled" by default in its basic form (with the possibility to enable it via the config) and iterate on it. We could  then try it out and iterate until stable.

the-mikedavis:
I think that's a good idea, I'll add a key to the config 👍

robertoaloi:
Thanks for this! Getting it in as disabled, so it can be refined.

PR #1314, opened 2022-05-25
Add completion support for features

plux:
### Description

Add completion support for [feature](https://www.erlang.org/doc/reference_manual/features.html) related macros and attributes.


robertoaloi:
You're on a roll!

PR #1315, opened 2022-05-25
Add atom typo diagnostics

plux:
### Description
* Add diagnostics to catch misspellings of common atoms.
* Add code action to fix the typo

### Example usage
atom_typo


robertoaloi:
Nice reuse of the Jaro distance, thanks! I appreciate it's false by default, so we can ensure it's not too resource consuming before rolling it out to all users.

PR #1322, opened 2022-06-09
Optimize unused includes detection by avoiding unnecessary work

robertoaloi:
While identifying unused includes, the current implementation performs
a go-to-definition operation for all POIs within a document, even if
all the .hrl candidates have already been excluded.

This change stops the iteration as soon as no candidates are present.

The whole detection algorithm could probably be revisited (it would be
good to check if this problem has been tackled in literature), but for
now this optimization will lower the execution time for "unused
includes" diagnostics and reduce the stress on the language server,
especially for big modules.

As an example, this optimization reduces the computing time for
"unused includes" diagnostics for the `els_parser.erl` module from ~1s
to ~200ms.


michalmuskala (approved):
This looks reasonable

PR #1327, opened 2022-06-11
Consider argument of ifdef, ifndef, undef attributes to be a macro

### LINKED ISSUE #1323: Wrong navigation for ifdef
**Describe the bug**
Given the following snippet where `~` represents cursor position:
```erl
-define(FOO, 1).

-ifdef(F~OO).
-endif.

-undef(FOO).
```

issuing a go to definion command lands us at:
```erl
-undef(FOO).
%%.    ^^^
```

**Expected behavior**
This should navigate to:
```erl
-define(FOO, 1).
%%.     ^^^
```

**Context**
 - `erlang_ls` version (tag/sha): 0.36
 - Editor used: VS Code

plux:
### Description

This should fix the navigation issues described in the issue.

Fixes #1323 .

### DIFF
--- a/apps/els_lsp/src/els_parser.erl
+++ b/apps/els_lsp/src/els_parser.erl
@@ -429,6 +429,14 @@ attribute(Tree) ->
                 undefined ->
                     [poi(Pos, spec, undefined)]
             end;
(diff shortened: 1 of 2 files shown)


robertoaloi:
Thanks!

PR #1328, opened 2022-06-13
Add support for finding callback implementation for dynamic calls

plux:
### Description

Add support for finding implementation when pointing to a dynamic call that is a behaviour callback.


plux:
Again, test failing in CI, but not on my machine -_-

robertoaloi:
@plux We need to find the root cause. Filed this one: https://github.com/erlang-ls/erlang_ls/issues/1335

plux:
This is ready to be merged

PR #1329, opened 2022-06-13
Merge server and provider processes.

robertoaloi:
The els_provide process is a remain from the times where providers had
their own dedicated processes and it does not currently provide
additional value. It actually makes provider errors not bubble up to
the client. In fact, in case of a provider error the provider process
would crash, but the server would indefinitely wait for a
response. The error would not be returned via LSP.

This change simplifies the architecture of Erlang LS by merging the
els_server and els_provider processes:

* Get rid of els_provider process and internal_state
* Remove obsolete available_providers function
* Do not pass server state to providers

As a follow up, it should be possible to:

* Pass requests as-they-are (as opposed to only the parameters) to
providers, avoiding an un-necessary translation layer
* Generalize the code in els_method

### DIFF
--- a/apps/els_core/src/els_provider.erl
+++ b/apps/els_core/src/els_provider.erl
@@ -2,254 +2,33 @@
 
 %% API
 -export([
-    handle_request/2,
-    start_link/0,
-    available_providers/0,
-    cancel_request/1,
-    cancel_request_by_uri/1
-]).
-
--behaviour(gen_server).
--export([
-    init/1,
-    handle_call/3,
-    handle_cast/2,
-    handle_info/2,
-    terminate/2
+    handle_request/2
 ]).
 
 %%==============================================================================
 %% Includes
 %%==============================================================================
--include_lib("kernel/include/logger.hrl").
+-include("els_core.hrl").
 
--callback handle_request(request(), any()) ->
+-callback handle_request(provider_request()) -> provider_result().
+
+-type provider() :: atom().
+-type provider_request() :: {atom() | binary(), map()}.
+-type provider_result() ::
     {async, uri(), pid()}
     | {response, any()}
     | {diagnostics, uri(), [pid()]}
     | noresponse.
--callback handle_info(any(), any()) -> any().
--optional_callbacks([handle_info/2]).
 
--type config() :: any().
--type provider() ::
-    els_completion_provider
-    | els_definition_provider
-    | els_document_symbol_provider
-    | els_hover_provider
-    | els_references_provider
-    | els_formatting_provider
-    | els_document_highlight_provider
-    | els_workspace_symbol_provider
-    | els_folding_range_provider
-    | els_implementation_provider
-    | els_code_action_provider
-    | els_general_provider
-    | els_code_lens_provider
-    | els_execute_command_provider
-    | els_rename_provider
-    | els_text_synchronization_provider
-    | els_signature_help_provider.
--type request() :: {atom() | binary(), map()}.
--type state() :: #{
-    in_progress := [progress_entry()],
-    in_progress_diagnostics := [diagnostic_entry()],
-    open_buffers := sets:set(buffer())
-}.
--type buffer() :: uri().
--type progress_entry() :: {uri(), job()}.
--type diagnostic_entry() :: #{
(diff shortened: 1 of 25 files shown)


apps/els_core/src/els_provider.erl
```diff
@@ -2,254 +2,33 @@
...
 %%==============================================================================
 %% Includes
 %%==============================================================================
--include_lib("kernel/include/logger.hrl").
+-include("els_core.hrl").
+
+-callback handle_request(provider_request()) -> provider_result().
 
--callback handle_request(request(), any()) ->
+-type provider() :: atom().
```
michalmuskala, line 15: It's the same for dialyzer, but better as a documentation

```suggestion
-type provider() :: module().
```

apps/els_lsp/src/els_code_lens_provider.erl
```diff
@@ -47,7 +47,7 @@ run_lenses_job(Uri) ->
         title => <<"Lenses">>,
         on_complete =>
             fun(Lenses) ->
-                els_provider ! {result, Lenses, self()},
+                els_server ! {result, Lenses, self()},
```
michalmuskala, line 50: I wonder, if it would make sense to have some mechanism and abstraction around this - raw sends always make me nervous

apps/els_lsp/src/els_methods.erl
```diff
@@ -184,54 +193,56 @@ shutdown(Params, State) ->
 %% exit
 %%==============================================================================
 
--spec exit(params(), state()) -> no_return().
+-spec exit(params(), els_server:state()) -> no_return().
 exit(_Params, State) ->
     Provider = els_general_provider,
-    Request = {exit, #{status => maps:get(status, State, undefined)}},
+    Request = {exit, #{status => maps:get(status, State)}},
     {response, _Response} = els_provider:handle_request(Provider, Request),
```
michalmuskala, line 200: I wonder - does it make sense to have the `els_provider:handle_request` abstraction? Wouldn't it be better to directly call the handler function? I feel like less indirection will make the code easier to follow

apps/els_core/src/els_provider.erl
```diff
@@ -2,254 +2,33 @@
...
+-type provider_request() :: {atom() | binary(), map()}.
```
michalmuskala, line 16: When is the request type passed in as a `binary()`?

michalmuskala (commented):
Overall this does what it says on the tin, and looks pretty good.

I think, though, this can go a step further and remove even more abstraction. For example, rather than having code like:
```erl
Provider = els_call_hierarchy_provider,
Request = {incoming_calls, Params},
els_provider:handle_request(Provider, Request)
```

I think it would be much clearer to see:
```erl
els_call_hierarchy_provider:incoming_calls(Params)
```
And I don't think it would loose any generality compared to the current implementation.

michalmuskala (approved):

robertoaloi:
> Overall this does what it says on the tin, and looks pretty good.
> 
> I think, though, this can go a step further and remove even more abstraction. For example, rather than having code like:
> 
> ```erlang
> Provider = els_call_hierarchy_provider,
> Request = {incoming_calls, Params},
> els_provider:handle_request(Provider, Request)
> ```
> 
> I think it would be much clearer to see:
> 
> ```erlang
> els_call_hierarchy_provider:incoming_calls(Params)
> ```
> 
> And I don't think it would loose any generality compared to the current implementation.

100% agree and this is very similar to what I had in mind in the follow-ups section of this PR. Will address this separately, though.

apps/els_lsp/src/els_code_lens_provider.erl
robertoaloi: Will follow up.

apps/els_core/src/els_provider.erl
robertoaloi: Probably an artifact of the past, I can fix.

apps/els_lsp/src/els_code_lens_provider.erl
robertoaloi: https://github.com/erlang-ls/erlang_ls/issues/1334

Issue #1205, opened 2022-02-16
Jump To Definition for Common Test TestCases


robertoaloi:
## Context

One of the core features provided by Erlang LS is _code navigation_, which allows the user to jump from the current location to the corresponding definition. This mechanism works for functions, macros, records and more. The way code navigation works is by means of a database of _points of interest_ which Erlang LS keeps in memory for each project. When accessing a project for the first time, Erlang LS scans every file in the project and extracts all _point of interests_ (e.g. a function definition, or a function application), together with their location (file path, starting line number, starting column number, ending line number, ending column number, etc).

Whenever a [Go To Definition Request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_definition) is sent from the IDE to the language server, Erlang LS looks up in the database the corresponding entry and returns the appropriate result.

Here's a sample request from the IDE:

```
[Trace - 05:25:03 PM] Sending request 'textDocument/definition - (5794)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/robertoaloi/git/github/erlang-ls/erlang_ls/apps/els_lsp/src/els_code_navigation.erl"
  },
  "position": {
    "line": 51,
    "character": 9
  }
}
```

The user, while visiting the `els_code_navigation.erl` file, requested a "go to definition" while in position `51:9`. In VS Code this is usually done by pressing `Cmd+click`.

If we peek at the actual code in `els_code_navigation.erl`, we can see that in this case the user was hovering a function application (i.e. a function call):

<img width="717" alt="Screenshot 2022-02-16 at 17 31 54" src="https://user-images.githubusercontent.com/91769/154311527-cdfb6af4-be4b-4cea-9082-89ff5669b65f.png">

Notice how in the protocol, compared to the IDE, line numbers are offset by 1, so line 52 in the editor corresponds to line 51 in the request.

Once received the request, the server interrogates its own DB and realizes that the location points to the local function call to `find/3`. It then performs an additional query, looking for the definition of the `find/3` function. The function is located on the same file (or `Uri`) on line `129`, so the following response can be sent.

```
[Trace - 05:25:03 PM] Received response 'textDocument/definition - (5794)' in 1ms.
Result: {
  "uri": "file:///Users/robertoaloi/git/github/erlang-ls/erlang_ls/apps/els_lsp/src/els_code_navigation.erl",
  "range": {
    "start": {
      "line": 128,
      "character": 0
    },
    "end": {
      "line": 128,
[...]

PR #1330, opened 2022-06-13
[#1205] Implement goto_definition for testcases.

RobinMorisset:
Summary:
https://github.com/erlang-ls/erlang_ls/issues/1205 suggested supporting the goto-definition feature for testcases.
Testcases are atoms (function definitions) and currently whenever doing goto-definition on an atom we treat it as a module.
I simply modified goto_definition to first look for a function of arity 1 in the current module when it is unclear whether the user is looking for a module or a testcase.

Test Plan:
I added the "testcase" test-case in els_definition_SUITE, following the structure of all other test-cases in that file.
This new testcase corresponds to doing "goto definition" on the atom "one" which appears in "all() -> [ one ]" in sample_SUITE.erl.
It verifies that we are sent to the definition of one at the bottom of sample_SUITE.erl.
I then checked that this new test-case passed, and that I did not break any other test in that suite. In particular, there were already tests for goto-definition on module names, and they still pass.

### DIFF
--- a/apps/els_lsp/src/els_code_navigation.erl
+++ b/apps/els_lsp/src/els_code_navigation.erl
@@ -71,10 +71,20 @@ goto_definition(
             Result
     end;
 goto_definition(
+    Uri,
+    #{kind := atom, id := Id} = POI
+) ->
+    %% Two interesting cases for atoms: testcases and modules.
+    %% Testcases are functions, so we first look for a function with the same name in the local scope
+    %% If we can't find it, we hope that the atom refers to a module.
+    case find(Uri, function, {Id, 1}) of
+        {error, _Error} -> goto_definition(Uri, POI#{kind := module});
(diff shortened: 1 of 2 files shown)


robertoaloi:
Hi @RobinMorisset and thanks for this! Love the approach and its simplicity. Considering that according to the protocol [it is possible to return multiple definitions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_definition) for a single request, I wonder we may want to take this a step further and give the user the ability to decide whether to go to a function or a module. We wouldn't even need to restrict to a specific arity and present all available options.

Said that, I would consider that out of scope for this task, and it can be done as a follow up.

robertoaloi (approved):

robertoaloi:
Thanks!

PR #1338, opened 2022-06-15
GotoDef returning multiple definitions for atoms

### LINKED ISSUE #1331: Return multiple definitions (modules, functions) to the user
## Context

One of the core features provided by Erlang LS is _code navigation_, which allows the user to jump from the current location to the corresponding definition. This mechanism works for functions, macros, records and more. The way code navigation works is by means of a database of _points of interest_ which Erlang LS keeps in memory for each project. When accessing a project for the first time, Erlang LS scans every file in the project and extracts all _point of interests_ (e.g. a function definition, or a function application), together with their location (file path, starting line number, starting column number, ending line number, ending column number, etc).

Whenever a [Go To Definition Request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_definition) is sent from the IDE to the language server, Erlang LS looks up in the database the corresponding entry and returns the appropriate result.

Here's a sample request from the IDE:

```
[Trace - 05:25:03 PM] Sending request 'textDocument/definition - (5794)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/robertoaloi/git/github/erlang-ls/erlang_ls/apps/els_lsp/src/els_code_navigation.erl"
  },
  "position": {
    "line": 51,
    "character": 9
  }
}
```

The user, while visiting the `els_code_navigation.erl` file, requested a "go to definition" while in position `51:9`. In VS Code this is usually done by pressing `Cmd+click`.

If we peek at the actual code in `els_code_navigation.erl`, we can see that in this case the user was hovering a function application (i.e. a function call):

Screenshot 2022-02-16 at 17 31 54

Notice how in the protocol, compared to the IDE, line numbers are offset by 1, so line 52 in the editor corresponds to line 51 in the request.

Once received the request, the server interrogates its own DB and realizes that the location points to the local function call to `find/3`. It then performs an additional query, looking for the definition of the `find/3` function. The function is located on the same file (or `Uri`) on line `129`, so the following response can be sent.

```
[Trace - 05:25:03 PM] Received response 'textDocument/definition - (5794)' in 1ms.
Result: {
  "uri": "file:///Users/robertoaloi/git/github/erlang-ls/erlang_ls/apps/els_lsp/src/els_code_navigation.erl",
  "range": {
    "start": {
      "line": 128,
      "character": 0
    },
    "end": {
      "line": 128,
      "character": 4
    }
  }
}
```

### What is required

When a `textDocument/definition` is performed on an _atom_, it is handled in a special way. Atoms in Erlang are constant literals, so it does not make too much sense to look for an atom definition per se. But atoms can also be used in special contexts. For example, Erlang module names are also atoms. Therefore, whenever the user tries to use the jump-to-definition functionality on an atom, we decided to jump to the respective module definition if that's available. That's not always desirable, though.
In [this PR](https://github.com/erlang-ls/erlang_ls/pull/1330) we extended the behaviour to jump to local functions first and, only if one is not defined, to a module. Since the LSP protocol allows returning multiple definitions to the user, why not to let the user decide?

* Generalize the `els_code_navigation:goto_definition/2` to return a list of definitions, instead of one.
* For atoms, include in the list of returned elements both local functions (of any arity) and the respective module, if they exist

### Code Pointers

* Go To Definition requests are handled in the `els_code_navigation` module, so you probably want to start there
* You probably want to specialize the clause of the `els_code_navigation:goto_definition/2` function which handles atoms

### Testing

Don't forget to include a test for the new functionality! You can look at existing examples in the `els_definition_SUITE` module and at this [code lenses tutorial](https://erlang-ls.github.io/articles/tutorial-code-lenses) which includes a [testing overview](https://erlang-ls.github.io/articles/tutorial-code-lenses/#adding-tests) for Erlang LS.

### Problems following the instructions?

Gabisampaio:
Summary: Allowing for more flexiblity when dealing with Erlang atoms. Hence, whenever the user tries to navigate to a definition, they will be able to choose which definition to navigate to.

Test Plan: Added the `multiple_atom_instances` test to the `els_definition_SUITE.erl`.
Ran the full test suite:
```
http_proxy='http://fwdproxy:8080' https_proxy='http://fwdproxy:8080' rebar3 ct --suite apps/els_lsp/test/els_definition_SUITE.erl
```
All tests passed.

Reviewers: @robertoaloi

Subscribers: @schaitoff

Tasks: T123303743

Tags:

### Description

I changed the `goto_definition` function of the `els_code_navigation` module to return a list of definitions instead of a single definition. Hence, when handling atoms, the developer is able to choose which definition to navigate to. I also adapted existing tests and added a new test in `els_definition_SUITE.erl`.

Fixes #1331.


apps/els_lsp/priv/code_navigation/src/code_navigation.erl
```diff
@@ -122,3 +122,9 @@ macro_b(_X, _Y) ->
 
 function_mb() ->
   ?MACRO_B(m, b).
+
+code_navigation() -> code_navigation.
```
robertoaloi, line 126: It looks like there's a case which doesn't work as expected. Try to tweak this test code by using an atom referring to a different module name. Eg:

```
code_navigation_atoms() -> code_navigation_atoms.

code_navigation_atoms(X) -> X.

multiple_instances() -> {code_navigation_atoms, [simple_list], "abc"}.
```
Only the functions will be found.

robertoaloi (changes requested):
Added a comment in the `code_navigation.erl` test module. In the more generic case where the atom does not coincide with the current module name, only functions are found.

robertoaloi:
To auto-format the code (so that CI does not complain), you can simply run `rebar3 fmt` locally and commit the changes.

apps/els_lsp/src/els_call_hierarchy_provider.erl
```diff
@@ -90,7 +90,8 @@ application_to_item(Uri, Application) ->
     #{id := Id} = Application,
     Name = els_utils:function_signature(Id),
     case els_code_navigation:goto_definition(Uri, Application) of
-        {ok, DefUri, DefPOI} ->
+        {ok, Defs} ->
+            {DefUri, DefPOI} = hd(Defs),
```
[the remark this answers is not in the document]
plux: I would suggest matching like this instead:
```erlang
{ok, [{DefUri, DefPOI} | _ ]} ->
```

apps/els_lsp/priv/code_navigation/src/code_navigation.erl
Gabisampaio: Many thanks for spotting this. I believe the issue has now been fixed.

Gabisampaio:
@robertoaloi @plux Many thanks for the detailed feedback. I think that the code is looking much better now.

robertoaloi (approved):

apps/els_lsp/src/els_definition_provider.erl
```diff
@@ -45,8 +45,14 @@ goto_definition(_Uri, []) ->
     null;
 goto_definition(Uri, [POI | Rest]) ->
     case els_code_navigation:goto_definition(Uri, POI) of
-        {ok, DefUri, #{range := Range}} ->
-            #{uri => DefUri, range => els_protocol:range(Range)};
+        {ok, Definitions} ->
```
robertoaloi, line 48: This now returns a list of maps, so the signature needs to be updated.

robertoaloi: You can run `rebar3 dialyzer` locally to run this type of checks without waiting for CI.

robertoaloi:
Thanks!

PR #1345, opened 2022-07-02
Do not send the 'body' field for DAP Initialized Event.

the-mikedavis:
### Description

The body field is required on most DAP Events and optional on Terminated but
is actually not included in Initialized at all. Clients parsing the Events
strictly will think an Initialized Event with a body is malformed, so this
change drops the field entirely for Initialized.

https://microsoft.github.io/debug-adapter-protocol/specification#Events_Initialized


the-mikedavis:
As a meta note: the contributing guide mentions that there should always be an issue for a PR - is it ok to send small fix PRs like this without creating an issue first?

PR #1346, opened 2022-07-03
DAP: allow breakpoint conditions to be `null`.

the-mikedavis:
### Description

A client may pass these breakpoint conditions as `null` rather than omitting the field from the message.

Relevant part of the spec: https://microsoft.github.io/debug-adapter-protocol/specification#Types_SourceBreakpoint

Notationally I think `?` allows a field to be passed as `null`, right? I can't seem to find a snippet in the spec that clarifies that point.


the-mikedavis:
Ah ok: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#languageServerProtocol

> The document uses TypeScript interfaces in strict mode to describe these. This means for example that a `null` value has to be explicitly listed and that a mandatory property must be listed even if a falsify value might exist.

And https://github.com/microsoft/debug-adapter-protocol/issues/130#issuecomment-669830521

> `null` is only a valid value if explicitly specified in a type definition.

---

Yep, Helix is sending `null` for these at the moment but I can fix that behavior within Helix 👍

PR #1347, opened 2022-07-03
DAP: add `expensive` field to scopes type.

the-mikedavis:
### Description

There's a required `expensive` field on the Scope Type used
to hint to the client whether the variables in the given scope are
expensive to retrieve.

https://microsoft.github.io/debug-adapter-protocol/specification#Types_Scope


the-mikedavis:
For a quick fix for clients that need this field we can just hard-code to `false`. I haven't read through the scopes part of the code yet (😅) - is it ever expensive on the server to fetch locals? Should we have some hueristic for setting this to `true`?

robertoaloi:
Nice catch. @TheGeorge do you have any opinions on this?

PR #1357, opened 2022-08-05
Ensure to pass an absolute path to the `els_uri:uri/1` function

robertoaloi:
The `compile:file/1` function can return a relative path depending on
the directory the emulator is started on. This corner case was
resulting in occasional crashes for the users which prevented
diagnostics to appear if an included file contained errors.

### DIFF
--- a/apps/els_lsp/src/els_compiler_diagnostics.erl
+++ b/apps/els_lsp/src/els_compiler_diagnostics.erl
@@ -680,17 +680,22 @@ inclusion_range(IncludePath, Document, include) ->
     [Range || #{id := Id, range := Range} <- POIs, Id =:= IncludeId];
 inclusion_range(IncludePath, Document, include_lib) ->
     POIs = els_dt_document:pois(Document, [include_lib]),
-    IncludeId = els_utils:include_lib_id(IncludePath),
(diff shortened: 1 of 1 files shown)


apps/els_lsp/src/els_compiler_diagnostics.erl
```diff
@@ -680,18 +680,23 @@ inclusion_range(IncludePath, Document, include) ->
     [Range || #{id := Id, range := Range} <- POIs, Id =:= IncludeId];
 inclusion_range(IncludePath, Document, include_lib) ->
     POIs = els_dt_document:pois(Document, [include_lib]),
-    IncludeId = els_utils:include_lib_id(IncludePath),
+    IncludeId = els_utils:include_lib_id(absolute_path(IncludePath)),
```
alanz, line 683: Do we know that IncludePath is always relative?

robertoaloi: It can also be an absolute one, but the join function takes care of that, if that's the case (maybe I should have called it `ensure_absolute_path`):

```
{ok,"/Users/robertoaloi/git/github/erlang-ls/erlang_ls"}
2> filename:join(Cwd, "/tmp/foo/bar").
"/tmp/foo/bar"
```

Issue #1365, opened 2022-08-21
No end progress sent when request is cancelled


ztion:
**Describe the bug**
For some requests, ErlangLS sends a progress notification. ("Lenses 0/1"). But no end progress is sent when request is cancelled. This is most easily seen for long running code lens, when switching to a different file in VScode. The request is cancelled, but the progress lives for ever. I work in a big code base, and I see this often when switching between files.

This can be seen in the output

[Trace - 10:49:57] Received notification '$/progress'.
Params: {
    "token": "1e985bd0-a71e-4a80-9a7a-227c121a3256",
    "value": {
        "cancellable": false,
        "kind": "begin",
        "message": "0 / 1",
        "percentage": 0,
        "title": "Lenses"
    }
}

[Trace - 10:49:57] Received notification 'window/logMessage'.
Params: {
    "message": "[2022-08-20T10:49:57.754785+02:00] [info] [SERVER] Suspending response [background_job=<0.859.0> id=10] [els_server:handle_request/2 L284] <0.148.0>\n",
    "type": 4
}

[2022-08-20T10:49:57.754785+02:00] [info] [SERVER] Suspending response [background_job=<0.859.0> id=10] [els_server:handle_request/2 L284] <0.148.0>

[Trace - 10:49:58] Sending notification '$/cancelRequest'.
Params: {
    "id": 10
}
[2022-08-20T10:50:03.832562+02:00] [error] Supervisor: {local,els_background_job_sup}. Context: shutdown_error. Reason: killed. Offender: id=els_background_job,pid=<0.859.0>. [supervisor:do_terminate/2 L905] <0.141.0>

**To Reproduce**
Easiest to reproduce with a large file. In VScode, open the large file, when you see the "Lenses 0/1" in the progress, switch to another file. The request will be cancelled by VScode, but the progress will never be ended. The erlfmt_parse.erl file is one easy file to test with.

**Expected behavior**
The progress should be ended.

**Actual behavior**
The progress is never stopped and lives "for ever".

**Context**
 - `erlang_ls` version (tag/sha): 0.41.2
 - Editor used: VScode
 - LSP client used:

PR #1366, opened 2022-08-21
Send end progress when request is cancelled.

ztion:
### Description
If task is long running, the background job gen_server will not process messages meanwhile. So when stopping background job, it seems supervisor will timeout and kill the process instead of graceful shutdown. So terminate function is not called.
This PR changes to run the task in a subprocess.

Also send an error response for cancelled tasks. As I understand it, even cancelled requests must send a response.

Fixes https://github.com/erlang-ls/erlang_ls/issues/1365 .


robertoaloi:
@ztion It makes sense to me, thanks for catching this and for providing a fix!

PR #1374, opened 2022-09-15
[dap] support non-verified breakpoints 

### LINKED ISSUE #1373: DAP crashes when a breakpoint is set on an unavailable module
**Describe the bug**
DAP crashes when a breakpoint is set on a module that is not availabe in the node that is debugged
**To Reproduce**
open a module that is not available in the node, and set a breakpoint

**Expected behavior**
Error message or something

**Actual behavior**
DAP crashes silently

TheGeorge:
### Description
support non-verified breakpoints in case the module isn't available in the debugged node

Before this would crash, now create a non-verified breakpoint, with some description

Fixes #1373

### DIFF
--- a/apps/els_dap/src/els_dap_breakpoints.erl
+++ b/apps/els_dap/src/els_dap_breakpoints.erl
@@ -139,7 +139,12 @@ do_line_breakpoints(Node, Module, LineBreakPoints, Breaks, Set) ->
 -spec do_function_breaks(node(), module(), [function_break()], breakpoints(), boolean()) ->
     breakpoints().
 do_function_breaks(Node, Module, FBreaks, Breaks, Set) ->
-    [[els_dap_rpc:break_in(Node, Module, Func, Arity) || {Func, Arity} <- FBreaks] || Set],
+    case Set of
+        true ->
+            [els_dap_rpc:break_in(Node, Module, Func, Arity) || {Func, Arity} <- FBreaks];
+        false ->
+            ok
+    end,
     case Breaks of
         #{Module := ModBreaks} ->
             Breaks#{Module => ModBreaks#{function => FBreaks}};
--- a/apps/els_dap/src/els_dap_breakpoints.erl
+++ b/apps/els_dap/src/els_dap_breakpoints.erl
@@ -3,8 +3,8 @@
     build_source_breakpoints/1,
     get_function_breaks/2,
     get_line_breaks/2,
-    do_line_breakpoints/4,
-    do_function_breaks/4,
+    do_line_breakpoints/5,
+    do_function_breaks/5,
     type/3
 ]).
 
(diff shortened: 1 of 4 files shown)


apps/els_dap/src/els_dap_general_provider.erl
```diff
@@ -1084,3 +1091,25 @@ distribution_error(Error) ->
             io_lib:format("Could not start Erlang distribution. ~p", [Error])
         )
     ).
+
+-spec maybe_interprete_and_clear_module(node(), module()) -> {boolean(), binary()}.
```
robertoaloi, line 1095: Nit: interprete -> interpret

apps/els_dap/src/els_dap_general_provider.erl
```diff
@@ -1084,3 +1091,25 @@ distribution_error(Error) ->
...
+maybe_interprete_and_clear_module(ProjectNode, Module) ->
+    case els_dap_rpc:interpretable(ProjectNode, Module) of
+        true ->
+            {module, Module} = els_dap_rpc:i(ProjectNode, Module),
+
+            %% purge all breakpoints from the module
```
robertoaloi, line 1101: Why do we want to purge them?

apps/els_dap/src/els_dap_breakpoints.erl
```diff
@@ -108,32 +108,38 @@ get_function_breaks(Module, Breaks) ->
...
+    boolean()
 ) ->
     breakpoints().
-do_line_breakpoints(Node, Module, LineBreakPoints, Breaks) ->
-    maps:map(
-        fun(Line, _) -> els_dap_rpc:break(Node, Module, Line) end,
-        LineBreakPoints
-    ),
+do_line_breakpoints(Node, Module, LineBreakPoints, Breaks, Set) ->
+    case Set of
```
robertoaloi, line 123: In which case do we get `Set=false`?

apps/els_dap/src/els_dap_breakpoints.erl
```diff
@@ -108,32 +108,38 @@ get_function_breaks(Module, Breaks) ->
...
             Breaks#{Module => #{line => LineBreakPoints, function => []}}
     end.
 
--spec do_function_breaks(node(), module(), [function_break()], breakpoints()) ->
+-spec do_function_breaks(node(), module(), [function_break()], breakpoints(), boolean()) ->
     breakpoints().
-do_function_breaks(Node, Module, FBreaks, Breaks) ->
-    [els_dap_rpc:break_in(Node, Module, Func, Arity) || {Func, Arity} <- FBreaks],
+do_function_breaks(Node, Module, FBreaks, Breaks, Set) ->
+    [[els_dap_rpc:break_in(Node, Module, Func, Arity) || {Func, Arity} <- FBreaks] || Set],
```
robertoaloi, line 142: Can we please use a regular statement (or a function) instead of the double list comprehension? :)

apps/els_dap/src/els_dap_general_provider.erl
```diff
@@ -192,20 +192,20 @@ handle_request(
...
             LineBreaks,
-            Breakpoints0
+            Breakpoints0,
+            IsModuleAvailable
         ),
+
     BreakpointsRsps = [
-        #{<<"verified">> => true, <<"line">> => Line}
-     || {{_, Line}, _} <- els_dap_rpc:all_breaks(ProjectNode, Module)
+        #{<<"verified">> => IsModuleAvailable, <<"line">> => Line, message => Message}
```
robertoaloi, line 207: Should message be a binary?

robertoaloi (approved):
LGTM, just a few nits.

apps/els_dap/src/els_dap_general_provider.erl
TheGeorge: It's just easier to code this way. The DAP protocol gives the following information on changes to breakpoints:
- complete set of line breakpoints per module
- complete set of function breakpoints in total

So we delete all breakpoints and set them again.

We should probably calculate a delta and then use `delete_break` and `del_break_in`. This PR doesn't change the general way we set breakpoints, just makes sure we don't crash on invalid ones.

apps/els_dap/src/els_dap_breakpoints.erl
TheGeorge: When the module does not exist, e.g. the debugged node cannot load the module.

apps/els_dap/src/els_dap_general_provider.erl
TheGeorge: Yeah, message should be a binary.

PR #1384, opened 2022-09-26
[#1353] Add support for prepareRename request

### LINKED ISSUE #1353: Add support for Prepare Rename Request
[Prepare Rename Request](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_prepareRename) was introduced in LSP ver. 3.12.0 and editors such as neovim have started to use it for project-wide renaming with LSP. It would be nice to support it.

As far as I can tell, the solution would be to add a new provider module `els_prepare_rename_provider.erl` which would largely mimic `els_rename_provider` except that instead of returning `{response, WorkspaceEdits}`, it should return the following result type (taken from the LSP spec):

```
Range | { range: Range, placeholder: string } | { defaultBehavior: boolean } | null
```

We could potentially force editors to use `rename` instead of `prepareRename` locally but that seems like an ugly hack instead of aiming to be at feature parity with the LSP spec.

**Additional context**

image

plux:
### Description

Add support for prepareRename request

Fixes #1353.

### DIFF
--- a/apps/els_core/src/els_client.erl
+++ b/apps/els_core/src/els_client.erl
@@ -46,7 +46,8 @@
     document_formatting/3,
     document_rangeformatting/3,
     document_ontypeformatting/4,
-    document_rename/4,
+    rename/4,
+    prepare_rename/3,
     folding_range/1,
     shutdown/0,
     start_link/1,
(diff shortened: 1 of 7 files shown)


plux:
I've only tested this in Emacs with lsp-mode. Would be good if someone with VS Code / Neovim could try it out as well.

PR #1386, opened 2022-10-03
Improve context awareness of completion

### LINKED ISSUE #1252: context sensitive completion for functions
currently we always materialize function arguments. If the courser is next to an `(` we shouldn't do this:

```erlang
string:|("some string"),
       ^
```

plux:
### Description
Improve context awareness of completion

- Only relevant type completions should be suggested when in type context (such as `-spec`, `-type`, and type definition inside `-record`).
- Don't include args in completion when pointer is right before `(`

Fixes #1252.


robertoaloi (approved):

robertoaloi:
Amazing, thanks! 🙏

PR #1395, opened 2022-10-28
Check snippet support in record field completion

the-mikedavis:
### Description

Completion for fields in records (see #1381) currently sends only snippet syntax. This change checks if the client supports snippets and falls back to only providing the field name when the client does not support snippets.


plux:
Looks good, I forgot about this.

robertoaloi (approved):

robertoaloi:
Thanks!

PR #1397, opened 2022-10-31
Show type docs when hovering type definitions

the-mikedavis:
### Description

This is a similar change as #1385: hovering definitions for types and opaques now shows the same documentation as hovering references. This is useful for checking how the hover will look to consumers of a type when writing docs.


robertoaloi:
Thanks!

PR #1398, opened 2022-11-02
Fix performance issue with unused include diagnostics

plux:
### Description

Unused include diagnostic can hog the CPU if there's unreachable include files in combination with huge search path and large source files.

The issue boils down to calling `file:path_open/3` on the unreachable files for every POI in the source file.
To improve the performance memoization on error was added in `els_indexing:find_and_deeply_index_file/1` to avoid repeatedly calling `file:path_open/3` on the same filename.


robertoaloi:
Thanks!

PR #1399, opened 2022-11-02
Add code action to add behaviour callbacks

plux:
### Description
Code action to add missing -behaviour callbacks.

add_callbacks

### DIFF
--- a/apps/els_lsp/src/els_execute_command_provider.erl
+++ b/apps/els_lsp/src/els_execute_command_provider.erl
@@ -23,7 +23,8 @@ options() ->
         <<"ct-run-test">>,
         <<"show-behaviour-usages">>,
         <<"suggest-spec">>,
-        <<"function-references">>
+        <<"function-references">>,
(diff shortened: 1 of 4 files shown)


apps/els_lsp/src/els_execute_command_provider.erl
```diff
@@ -105,6 +106,85 @@ execute_command(<<"suggest-spec">>, [
...
+        <<"uri">> := Uri,
+        <<"behaviour">> := Behaviour
+    }
+]) ->
+    {ok, Document} = els_utils:lookup_document(Uri),
+    case els_utils:find_module(binary_to_atom(Behaviour, utf8)) of
+        {error, _} ->
+            [];
+        {ok, BeUri} ->
+            %% Put exported callback functions after -behaviour() or -export()
```
robertoaloi, line 120: As a follow up, we could verify whether some callback functions are already exported and add them to the existing list.

robertoaloi:
Nice one!

PR #1402, opened 2022-11-08
Make test suite work in OTP 25

fabjan:
Problem
======

The hover and completion provider test suites have some test cases for showing docs.

Some tests are testing the Erlang LS internal fallback docs provided if EEP-48 docs can not be extracted. These test cases don't work with newer OTP versions (because their expected values depend on the EEP-48 docs not being available).

Some of the tests are testing edoc extraction itself, and have conditional expectations depending on if EEP-48 docs are available or not. These tests assume that OTP >= 24 will always have EEP-48 docs for `file:write` if `code:get_docs(file)` is not empty, but this is not the case (see e.g. https://github.com/erlang-ls/erlang_ls/pull/1402#discussion_r1020932452). The test suites and the code in els_docs are not in agreement about when the docs are available.

Proposed Solution
==============

* Mock the els_docs eep48 calls to say eep48 is unavailable for the tests that test that fallback
* Skip the OTP module specific tests if OTP was not built with doc chunks (as far as I can tell, those test cases don't test any Erlang LS code that other test cases don't cover)

If the doc chunks are expected to be there in OTP for anyone running the test suite, perhaps we can document how to resolve that (as it is now, I don't know how to make the test suite pass locally).

Bonus Content
===========

* Add OTP 25 to CI.
* Drop OTP 22 from CI.

### DIFF
--- a/apps/els_lsp/test/els_completion_SUITE.erl
+++ b/apps/els_lsp/test/els_completion_SUITE.erl
@@ -1919,12 +1919,8 @@ resolve_type_application_remote_otp(Config) ->
         Result
     ),
     #{documentation := #{value := ActualValue}} = Result,
-    case {has_eep48_edoc(), has_eep48(file)} of
-        {true, false} ->
-            %% If this should fail the test we should document requirements for
-            %% the development environment so that OTP has the EEP-48 docs.
-            {skip, "This OTP build does not have EEP-48 docs for the 'file' module"};
-        {true, true} ->
+    case has_eep48(file) of
+        true ->
             ExpectValue = <<
                 "```erlang\n-type name_all() ::\n     string() |"
                 " atom() | deep_list() | (RawFilename :: binary()).\n"
@@ -1939,7 +1935,7 @@ resolve_type_application_remote_otp(Config) ->
                 "\\(not even at the end\\)\\.\n"
             >>,
             ?assertEqual(ExpectValue, ActualValue);
-        {false, _} ->
+        false ->
             ExpectValue = <<
                 "```erlang\n-type name_all()  :: "
                 "string() | atom() | deep_list() | "
--- a/apps/els_lsp/test/els_completion_SUITE.erl
+++ b/apps/els_lsp/test/els_completion_SUITE.erl
@@ -1971,7 +1971,7 @@ has_eep48(Module) ->
         {ok, {docs_v1, _, erlang, _, _, _, Docs}} ->
             lists:any(
                 fun
-                    ({_, _, _, #{}, _}) -> true;
+                    ({_, _, _, Doc, _}) when is_map(Doc) -> true;
                     ({_, _, _, _, _}) -> false
                 end,
                 Docs
--- a/apps/els_lsp/test/els_completion_SUITE.erl
+++ b/apps/els_lsp/test/els_completion_SUITE.erl
@@ -1675,44 +1675,49 @@ resolve_application_remote_otp(Config) ->
         <<"write/2">>
     ),
     #{result := Result} = els_client:completionitem_resolve(Selected),
-    Value =
-        case has_eep48(file) of
-            true ->
-                <<
-                    "```erlang\nwrite(IoDevice, Bytes) -> ok | {error, "
-                    "Reason}\nwhen\n  IoDevice :: io_device() | atom(),\n  Bytes ::"
-                    " iodata(),\n  Reason :: posix() | badarg | terminated.\n```\n\n"
-                    "---\n\nWrites `Bytes` to the file referenced by `IoDevice`\\. "
-                    "This function is the only way to write to a file opened in `raw`"
-                    " mode \\(although it works for normally opened files too\\)\\. "
-                    "Returns `ok` if successful, and `{error, Reason}` otherwise\\."
-                    "\n\nIf the file is opened with `encoding` set to something else "
-                    "than `latin1`, each byte written can result in many bytes being "
-                    "written to the file, as the byte range 0\\.\\.255 can represent "
-                    "anything between one and four bytes depending on value and UTF "
-                    "encoding type\\.\n\nTypical error reasons:\n\n* **`ebadf`**  \n"
-                    "  The file is not opened for writing\\.\n\n* **`enospc`**  \n"
-                    "  No space is left on the device\\.\n"
-                >>;
-            false ->
-                <<
-                    "## file:write/2\n\n---\n\n```erlang\n\n  write(File, "
-                    "Bytes) when is_pid(File) orelse is_atom(File)\n\n  write(#file_"
-                    "descriptor{module = Module} = Handle, Bytes) \n\n  write(_, _) "
-                    "\n\n```\n\n```erlang\n-spec write(IoDevice, Bytes) -> ok | "
-                    "{error, Reason} when\n      IoDevice :: io_device() | atom(),"
-                    "\n      Bytes :: iodata(),\n      Reason :: posix() | "
-                    "badarg | terminated.\n```"
-                >>
-        end,
-    Expected = Selected#{
-        documentation =>
-            #{
-                kind => <<"markdown">>,
-                value => Value
-            }
-    },
-    ?assertEqual(Expected, Result).
+    ?assertMatch(
+        #{
+            data := #{module := <<"file">>, function := <<"write">>, arity := 2},
+            documentation := #{kind := <<"markdown">>, value := _}
+        },
+        Result
+    ),
+    #{documentation := #{value := ActualValue}} = Result,
+    case {has_eep48_edoc(), has_eep48(file)} of
+        {true, false} ->
+            %% If this should fail the test we should document requirements for
+            %% the development environment so that OTP has the EEP-48 docs.
+            {skip, "This OTP build does not have EEP-48 docs for the 'file' module"};
+        {true, true} ->
+            ExpectValue = <<
+                "```erlang\nwrite(IoDevice, Bytes) -> ok | {error, "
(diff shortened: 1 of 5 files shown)


apps/els_lsp/test/els_hover_SUITE.erl
```diff
@@ -490,13 +508,20 @@ nonexisting_type(Config) ->
     #{result := Result} = els_client:hover(Uri, 22, 15),
     %% The spec for `j' is shown instead of the type docs.
     Value =
-        case has_eep48_edoc() of
-            true ->
+        case list_to_integer(erlang:system_info(otp_release)) of
+            %% WIP: I think this might not be the way to go,
+            %% just putting it here for the PR.
```
fabjan, line 513: The results were different for me locally with OTP 24 and OTP 25.

apps/els_lsp/test/els_completion_SUITE.erl
```diff
@@ -1675,44 +1675,49 @@ resolve_application_remote_otp(Config) ->
...
-    },
-    ?assertEqual(Expected, Result).
+    ?assertMatch(
+        #{
+            data := #{module := <<"file">>, function := <<"write">>, arity := 2},
+            documentation := #{kind := <<"markdown">>, value := _}
+        },
+        Result
+    ),
+    #{documentation := #{value := ActualValue}} = Result,
```
fabjan, line 1685: This refactor was made to make case the last part of the expression, so it can return `{skip, _}`.

(The `data := #{module := <<"file">>, function := <<"write">>, arity := 2}` check is new though.)

apps/els_lsp/test/els_completion_SUITE.erl
```diff
@@ -1675,44 +1675,49 @@ resolve_application_remote_otp(Config) ->
...
+    case {has_eep48_edoc(), has_eep48(file)} of
+        {true, false} ->
+            %% If this should fail the test we should document requirements for
+            %% the development environment so that OTP has the EEP-48 docs.
+            {skip, "This OTP build does not have EEP-48 docs for the 'file' module"};
```
fabjan, line 1690: Maybe you want this to fail instead of skipping. I think the docs should be updated to ensure developers have the right stuff locally to run the test suite.

I have OTP 24 and 25 installed via asdf and `code:get_doc(file)` yields a map with docs, but no functions have docs:

```
1> {ok, {docs_v1, 0, erlang, <<"application/erlang+html">>, none, _, ModDocs}} = code:get_doc(file).
2> [{F, A, Doc} || {{function, F, A}, _, _, Doc, _} <- ModDocs].
[{native_name_encoding,0,none},
 {format_error,1,none},
 {pid2name,1,none},
 {get_cwd,0,none},
 {get_cwd,1,none},
 {set_cwd,1,none},
 {delete,1,none},
 {delete,2,none},
 {rename,2,none},
 {make_dir,1,none},
 {del_dir,1,none},
 {del_dir_r,1,none},
 {read_file_info,1,none},
 {read_file_info,2,none},
 {altname,1,none},
 {read_link_info,1,none},
 {read_link_info,2,none},
 {read_link,1,none},
 {read_link_all,1,none},
 {write_file_info,2,none},
 {write_file_info,3,none},
 {list_dir,1,none},
 {list_dir_all,1,none},
 {read_file,1,none},
 {make_link,2,none},
 {make_symlink,2,none},
 {write_file,2,...},
 {write_file,...},
 {...}|...]
3> [{F, A, Doc} || {{function, F, A}, _, _, Doc, _} <- ModDocs, Doc =/= none].
[]
```

apps/els_lsp/test/els_hover_SUITE.erl
```diff
@@ -524,10 +549,32 @@ nonexisting_module(Config) ->
...
-        {ok, _} -> true;
-        _ -> false
+        {ok, {docs_v1, _, erlang, _, _, _, Docs}} ->
+            lists:any(
+                fun
+                    ({_, _, _, #{}, _}) -> true;
+                    ({_, _, _, _, _}) -> false
+                end,
+                Docs
+            );
```
fabjan, line 577: This and `has_eep48/1` in the other suite were only updated to know when the cases can be skipped. If skipping is not wanted this can go away (or be used to improve test debugging).

fabjan:
Feel free to close this PR if there is too much else going on. I prefer to not be the one closing it since it is not targeting my repository.

Maybe main has moved too much anyway by now (didn't look) for this to make sense.

robertoaloi:
Hi @fabjan, sorry for not being responsive to Pull Requests in the past couple of months. The contribution looks good to me, but the CI jobs seem to have expired and I cannot find a way to restart them. If you could rebase/force-push this branch we can merge this one, assuming CI passes.

fabjan:
No worries at all. Sure I'll give it a try!

robertoaloi:
@fabjan It looks like a test suite is still failing on this one.

fabjan:
Cheers, Time to look at this one now after the rebases and other PR was merged 👍

fabjan:
Alright let's see now how it plays. I got all green checks running locally on OTP 23 so my "fix" is a guess, but I think it is sound: https://github.com/erlang-ls/erlang_ls/pull/1402/commits/de4f89dd4279b64009a35a18f791d6c7c73c8107

fabjan:
Can't get OTP 22 to work on my M1 Mac, so my attempts at making the test suites work for all releases are being made a little bit in the dark.

fabjan:
@robertoaloi I think I have to give up.

I don't know enough about how the test suite or the docs rendering feature was built to make a call on how to fix the test suite.

* I can't install OTP 22 locally to test.
* `-if(?OTP_RELEASE >= 23).` is used to conditionally compile/include/export some of the code used.
* `-ifdef(NATIVE_FORMAT).` is used to conditionally compile some other code.
* `list_to_integer(erlang:system_info(otp_release)) >= 24.` is used to select some test code.
* The test code in addition to the above depends on exactly how the local OTP you use was compiled (does the stdlib have doc chunks).

This PR was a much bigger task than I saw before me. It seems to make it work the test code needs a bigger refactor.

robertoaloi:
Hi @fabjan I think it's ok to drop support for OTP 22. They just announced the RC1 for OTP 26. So, feel free to remove it from the pipeline.

fabjan:
I bumped the Windows CI job to OTP 23 as well (it was at 22). Interestingly, it was green last time even though it seems like it tested nothing: https://github.com/erlang-ls/erlang_ls/actions/runs/4198452678/jobs/7282095992

Maybe now with OTP 23 it will run the tests and go red..

apps/els_lsp/test/els_hover_SUITE.erl
fabjan: looking into this again, maybe it was something temporary

fabjan: I tried using the has_eep48_edoc() detector instead, but it is giving `true` on OTP 24 and OTP 25, but they get different results. This might be something we should maybe dive into and see why, and to fix it instead of futzing with the tests.

fabjan: "resolving" this thread to open one with my findings in the newest commit instead.

apps/els_lsp/test/els_hover_SUITE.erl
```diff
@@ -526,13 +540,15 @@ nonexisting_type(Config) ->
     #{result := Result} = els_client:hover(Uri, 22, 15),
     %% The spec for `j' is shown instead of the type docs.
     Value =
-        case has_eep48_edoc() of
-            true ->
+        case list_to_integer(erlang:system_info(otp_release)) of
+            25 ->
                 <<
-                    "## j/1\n\n---\n\n```erlang\n\n  j(_) \n\n```\n\n"
-                    "```erlang\n-spec j(doesnt:exist()) -> ok.\n```"
```
fabjan, line 533: I think there is a bug in `els_eep48_docs:render` when running in OTP versions under 25, and that is why this current test code looks a bit weird (it checks if has_eep48_edoc() is true or false but uses the same test data regardless)

On OTP 24 locally on my machine, `els_eep48_docs:render` returns `{error, function_missing}` for this `j` function that is hovered. On OTP 25 it returns some kind of doc chunks instead even of the function is not documented. Maybe OTP started automatically adding doc chunks for specs?

Anywho, I didn't want to make changes to the production code in this PR adding OTP 25 to CI, so not diving into if `els_eep48_docs:render` should be improved (especially since it seems to me that it's under OTP 24 it is behaving badly).

apps/els_lsp/test/els_completion_SUITE.erl
fabjan: Ah, this is no longer useful after the latest few commits in this branch.

fabjan:
Converted to draft until the useless diffs are cleaned up.

But CI is all green now!

apps/els_lsp/test/els_hover_SUITE.erl
```diff
@@ -78,12 +78,21 @@ init_per_testcase(TestCase, Config) ->
 
 -spec end_per_testcase(atom(), config()) -> ok.
 end_per_testcase(TestCase, Config) ->
+    lists:foreach(
```
robertoaloi, line 81: The fact that meck registers a module with the `_meck` suffix is an implementation detail that we should not rely on. We should only unload `els_docs` for those testcases where the module is mocked.

TheGeorge: The correct way to do "unmock if mocked" is to use the `meck` API instead. `meck:mocked` returns a list of mocked modules.

fabjan: Ah nice to know 👍

robertoaloi:
Thanks!

PR #1404, opened 2022-11-10
Find references for atoms

the-mikedavis:
### Description

This change indexes atoms in the references table and adds support for finding atoms in the references provider.


plux:
Thanks for your contribution @the-mikedavis !

Issue #1060, opened 2021-07-10
Regression in DAP server


robertoaloi:
The DAP server does not bootstrap correctly with the latest `HEAD`. `git bisect` showed the regression has been introduced by commit `67827d4dd95cfb4ea950d15018bc45f3ff6c352c`.

That commit introduced a server-initiated notification in case of a missing configuration. The `els_dap_server` seems to reuse the `els_config:initialize/3` function which sends the notification via the `els_server`, not available via the DAP escript.

As a temporary and quick solution, we can add a flag for the configuration check and use it only in the LSP server. A more appropriate solution is to remove this dependency from the DAP server onto the `els_config` module, which is only used for a couple of options (e.g. capabilities and runtime).

PR #1406, opened 2022-11-14
[#1401] Present configuration read errors for the user

### LINKED ISSUE #1401: erlang_ls should be more explicit when it finds an `erlang_ls.config` that isn't valid

Right now, if `erlang_ls.config` is not valid, `erlang_ls` will log it and move on, but it is not always abundantly clear that this has happened. For instance

```
[2022-11-07T11:22:18.612874-05:00] [debug] Could not read config file: path="/home/nick/Documents/code/my_project/erlang_ls.config" class=throw error={yamerl_exception,[{yamerl_parsing_error,error,"Unexpected \"yamerl_scalar\" token following a \"yamerl_scalar\" token",2,1,unexpected_token,{yamerl_scalar,2,1,{yamerl_tag,2,1,{non_specific,"?"}},flow,plain,"plt_path \"/home/nick/Documents/code/my_project/_build/default/rebar3_21.3.8.4_plt\""},[]}]} [els_config:consult_config/1 L350] <0.148.0>
```

On my system, where I have no global `erlang_ls.config`, this manifests itself in VSCode with an alert saying there's no `erlang_ls.config` file.

I think there's a few ways this could be tackled

1. At minimum, the above log could be bumped up from a DEBUG to a WARN or ERROR.
2. Editors like VSCode could give a more explicit error that a config file was invalid, much like there is when there is no config file.
3. `erlang_ls` should maybe stop execution if this happens. I can think of very few cases where someone might want to continue operation after a configuration is determined to be invalid; falling back to the next available config ironically masks the problem, since it might not be obvious that this occurred.

See previous section

**Additional context**
[This thread](https://erlanger.slack.com/archives/CAZR2QDB3/p1667836049607529) in the Erlang slack is where this originally came up

fabjan:
Description
========

If the the configuration file(s) found cannot be parsed, show a notice for the user instead of silently moving on to the next config alternative.

Fixes  #1401

### DIFF
--- a/apps/els_core/src/els_config.erl
+++ b/apps/els_core/src/els_config.erl
@@ -359,10 +359,12 @@ consult_config(Path) ->
     Options = [{map_node_format, map}],
     try yamerl:decode_file(Path, Options) of
         [] ->
-            ?LOG_WARNING("Using empty configuration from: ~p", [Path]),
+            ?LOG_WARNING("Using empty configuration from ~s", [Path]),
             {ok, #{}};
-        [Config] ->
-            {ok, Config}
+        [Config] when is_map(Config) ->
+            {ok, Config};
+        _ ->
+            {error, {syntax_error, Path}}
     catch
         Class:Error ->
             {error, {Class, Error}}
--- a/apps/els_core/src/els_config.erl
+++ b/apps/els_core/src/els_config.erl
@@ -92,27 +92,37 @@
 
 -spec initialize(uri(), map(), map()) -> ok.
 initialize(RootUri, Capabilities, InitOptions) ->
-    initialize(RootUri, Capabilities, InitOptions, _ReportMissingConfig = false).
+    initialize(RootUri, Capabilities, InitOptions, _ErrorReporting = no_els_server).
 
--spec initialize(uri(), map(), map(), boolean()) -> ok.
-initialize(RootUri, Capabilities, InitOptions, ReportMissingConfig) ->
+-spec initialize(uri(), map(), map(), use_els_server | no_els_server) -> ok.
+initialize(RootUri, Capabilities, InitOptions, ErrorReporting) ->
     RootPath = els_utils:to_list(els_uri:path(RootUri)),
     ConfigPaths = config_paths(RootPath, InitOptions),
-    {GlobalConfigPath, GlobalConfig} = consult_config(
-        global_config_paths(),
-        false
-    ),
-    {LocalConfigPath, LocalConfig} = consult_config(
-        ConfigPaths,
-        ReportMissingConfig
-    ),
+    {GlobalConfigPath, MaybeGlobalConfig} = find_config(global_config_paths()),
+    {LocalConfigPath, MaybeLocalConfig} = find_config(ConfigPaths),
     ConfigPath =
         case LocalConfigPath of
-            undefined -> GlobalConfigPath;
-            _ -> LocalConfigPath
+            undefined ->
+                report_missing_config(ErrorReporting),
+                GlobalConfigPath;
+            _ ->
+                LocalConfigPath
         end,
-    %% Augment Config onto GlobalConfig
-    Config = maps:merge(GlobalConfig, LocalConfig),
+    Config =
+        case {MaybeGlobalConfig, MaybeLocalConfig} of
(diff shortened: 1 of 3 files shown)


fabjan:
The notice in VS Code:

Skärmavbild 2022-11-14 kl  20 18 02

Example log message (I don't know why it shows up twice, that happens for me with any version of the language server):

Skärmavbild 2022-11-14 kl  20 18 15

apps/els_core/src/els_config.erl
```diff
@@ -92,27 +92,37 @@
 
 -spec initialize(uri(), map(), map()) -> ok.
 initialize(RootUri, Capabilities, InitOptions) ->
-    initialize(RootUri, Capabilities, InitOptions, _ReportMissingConfig = false).
+    initialize(RootUri, Capabilities, InitOptions, _ErrorReporting = no_els_server).
 
--spec initialize(uri(), map(), map(), boolean()) -> ok.
-initialize(RootUri, Capabilities, InitOptions, ReportMissingConfig) ->
+-spec initialize(uri(), map(), map(), use_els_server | no_els_server) -> ok.
```
fabjan, line 97: My reasoning for changing this parameter was that until I happened upon https://github.com/erlang-ls/erlang_ls/issues/1060 I thought the flag was only about verbosity or the config file missing. But after seeing that issue, it appears the purpose of the flag was to prevent `els_server:send_notification` from being called when the DAP script calls `els_config:initialize/3`.

(Perhaps `els_config:initialize/3` can be removed completely and `els_dap_general_provider.erl` updated to call /4 with `no_els_server`).

Other names for the flag could maybe be `no_notice | show_notice`.

fabjan:
Feel free to close this PR if there is too much else going on. I prefer to not be the one closing it since it is not targeting my repository.

robertoaloi:
LGTM, but again the CI jobs seem to be expired and could not find a way to re-run them. If you can force-push, we can merge this one.

fabjan:
I rebased on master since I was going to force push and get another CI run anyway.

When doing a last minute test locally I found that a bad YAML yields a crash when merging the configs so I added a little `syntax_error` reason for the broken config reporting.

Skärmavbild 2023-02-11 kl  19 50 00

apps/els_core/src/els_config.erl
```diff
@@ -86,33 +86,46 @@
     providers => map()
 }.
 
+-type error_reporting() :: lsp_notification | log.
```
fabjan, line 89: As I read through my changes again after some time away from it I came to the conclusion that `no_els_server` and `use_els_server` were kind of bad names for the variants. I think this is more clear.

apps/els_core/src/els_config.erl
```diff
@@ -86,33 +86,46 @@
...
+
 %%==============================================================================
 %% Exported functions
 %%==============================================================================
 
 -spec initialize(uri(), map(), map()) -> ok.
 initialize(RootUri, Capabilities, InitOptions) ->
-    initialize(RootUri, Capabilities, InitOptions, _ReportMissingConfig = false).
+    %% https://github.com/erlang-ls/erlang_ls/issues/1060
+    initialize(RootUri, Capabilities, InitOptions, _ErrorReporting = log).
```
fabjan, line 98: `initialize/3` could be removed completely unless it is kept for some backwards compatibility I can't see.

fabjan:
@robertoaloi I really didn't mean to spam when I commented the PRs can be closed. I just wanted to make sure they were not going to linger around making a mess!

I will touch them one by one now, so the other two will be "on hold" until this one is merged or closed.

Sorry for scope creeping a bit here after you already reviewed it, but I just changed the names of the error reporting selector variants and a little error handling for configs that are valid YAML but not maps (and thus can't be used at all) which I discovered when picking this back up, rebasing and testing locally.

apps/els_core/src/els_config.erl
robertoaloi: I don't see a reason to keep it, either.

robertoaloi:
> @robertoaloi I really didn't mean to spam when I commented the PRs can be closed.

No problem at all, thanks for the call to action and for the contribution!

robertoaloi:
Thanks for the contribution!

fabjan:
Looks like the Windows CI failed because chocolatey could not install Erlang 22.3 :(

Issue #655, opened 2020-04-05
Extra autocompletions when defining functions


michalmuskala:
**Is your feature request related to a problem? Please describe.**
After using the extension for couple days in my day-to-day work, I noticed some missing autocompletions, that I seem rather annoying that they are missing.

**Describe the solution you'd like**
* autocomplete the function head when typing in another head of the same (possibly incomplete) function. For example given the snippet below, when cursor is at `|` there should be a completion for `foo(Arg)`. Furthermore, I would say that that should be the only suggestion offered - other functions shouldn't be displayed since they would be invalid anyway.
```erlang
foo(1) -> ok;
|
```
* autocomplete missing functions. When cursor is at the very beginning of the line and we start typing in a name of a function, there should be autocompletions for functions that are missing in the file (a use of the function was already written down, but the function itself not yet). Today we get a list of all the functions defined in the file - this isn't very useful when defining a new function.

**Describe alternatives you've considered**
I'm not sure there are alternatives other than not supporting this at all.

robertoaloi:
Hi @michalmuskala , they look like nice improvements to me. Are you planning to contribute them yourself?

michalmuskala:
I don't have plans to work on this soon-ish, I might if it becomes frustrating enough, though 😉 If I pick it up I'll ping this issue.

PR #1407, opened 2022-11-16
Add completion for function clause

ztion:
### Description

With this change you will get an autocompletion for a new function clause after you type the semicolon ";" in a function clause body.

So when you type
```
test_fun(Test1, Test2) ->
    test;
```

The editor will popup the option to autocomplete function clause, and ErlangLS will insert the following, where you can tab between the fields:

```
test_fun(Arg1, Arg2) ->
   Body
```

The autocompletion should not be offered in function guard sequences or in case/receive/etc clauses. Maybe it would be nice to have something like this for other kind of clauses as well, but since incomplete statements seems not to be parsed currently, it is hard to tell them apart.

Something similar was mentioned in https://github.com/erlang-ls/erlang_ls/issues/655.

### DIFF
--- a/apps/els_lsp/src/els_completion_provider.erl
+++ b/apps/els_lsp/src/els_completion_provider.erl
@@ -43,7 +43,8 @@ trigger_characters() ->
         <<"-">>,
         <<"\"">>,
         <<"{">>,
-        <<" ">>
+        <<" ">>,
+        <<";">>
     ].
 
 -spec handle_request(els_provider:provider_request()) -> {response, any()}.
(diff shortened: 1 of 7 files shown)


apps/els_lsp/src/els_completion_provider.erl
```diff
@@ -418,6 +427,77 @@ complete_type_definition(Document, Name, ItemFormat) ->
...
+                            [atom_to_list(Name), ArgumentText, Arity + 1]
+                        ),
+
+                    [
+                        #{
+                            label => <<"Add function clause">>,
+                            kind => ?COMPLETION_ITEM_KIND_SNIPPET,
+                            insertText => list_to_binary(Snippet),
+                            insertTextFormat => ?INSERT_TEXT_FORMAT_SNIPPET,
+                            insertTextMode => ?INSERT_TEXT_MODE_AS_IS
```
ztion, line 490: Not 100% happy about the "As Is" insert mode, since I think it might not be perfect for people with other indentation styles..

Could it possibly be done using "textEdit" property of the completion item? But can that one add text on the line below?

TheGeorge:
I attempted something similar in #1261 

There was some discussion if this is something we ultimately wanted to include in the LS.

PR #1413, opened 2022-11-28
Allow `code_reload.node` to be a list of nodes

mkuratczyk:
Hi. When working on clustered applications (RabbitMQ in my case), it's useful to reload the code on all nodes in the cluster. Otherwise I can't really use reloading since I still need to restart the other nodes to pick up the changes. This is a very rudimentary change that gets the job done for me (I've been happily using it for a few days). Would you accept something like that?

I decided to "overload" the `code_reload.node` setting and allow it to be a list. Alternatively, we could have a separate field (`code_reload.nodes`) but then we would need to handle both being set for example. I'm happy to make such changes if you prefer.

Thanks,

### DIFF
--- a/apps/els_lsp/src/els_code_reload.erl
+++ b/apps/els_lsp/src/els_code_reload.erl
@@ -12,15 +12,30 @@ maybe_compile_and_load(Uri) ->
     Ext = filename:extension(Uri),
     case els_config:get(code_reload) of
         #{"node" := NodeStr} when Ext == <<".erl">> ->
-            Node = els_utils:compose_node_name(
-                NodeStr,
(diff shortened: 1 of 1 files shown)


plux:
Thanks for your contribution @mkuratczyk !
This seems like a fine addition.
Seems like there is a conflict hat needs to be resolved in order to merge this.
Please rebase on main and resolve the conflict so we can merge it.

mkuratczyk:
Done. Thank you!

plux:
Thanks for your contribution @mkuratczyk !

CI seems to be a failing on coveralls for some reason, I will merge anyway.

PR #1417, opened 2023-01-27
Add docs memoization

### LINKED ISSUE #1208: Memoize `textDocument/hover` requests to produce Edoc
### Context

One of the core features provided by Erlang LS is to provide inline documentation on _hover_. The way inline documentation works is by means of a database of points of interest which Erlang LS keeps in memory for each project. When accessing a project for the first time, Erlang LS scans every file in the project and extracts all point of interests (e.g. a function definition, or a function application), together with their location (file path, starting line number, starting column number, ending line number, ending column number, etc). When the user hovers a function call (known as a function _application_ in the code base), the IDE sends a [textDocument/hover] request to the server:

```
[Trace - 09:22:34 AM] Sending request 'textDocument/hover - (5958)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/robertoaloi/git/github/erlang-ls/erlang_ls/apps/els_lsp/src/els_docs.erl"
  },
  "position": {
    "line": 91,
    "character": 16
  }
}
```

In this case the user was apparently hovering the location `91:6` in the `els_docs.erl` file. If we peek at the code, we notice that corresponds to the call to the local `eep48_docs/4` function. Notice how the line and column numbers in the request are offset by 1, compared to the ones in the editor (line 91 in the request corresponds to line 92 in the editor).

Screenshot 2022-02-17 at 09 24 33

Upon receiving the request, the language server interrogates its own internal database and realizes that the hovered function application is `eep48_docs/4` from the `els_docs` module. Using that information, the server calculates the documentation for the function and returns it to the IDE:

```
[Trace - 09:22:35 AM] Received response 'textDocument/hover - (5958)' in 57ms.
Result: {
  "contents": {
    "value": "```erlang\neep48_docs(Type :: function | type,\n      M :: atom(),\n      F :: atom(),\n      A :: non_neg_integer()) ->\n       {ok, string()} | {error, not_available}.\n```\n\n---\n\nFetch EEP\\-48 style Docs\n\nOn OTP 23\\+ systems, Erlang has support for fetching documentation from the code or from a \\.chunk file\\. This function tries to fetch and render the documentation using EEP\\-48 docs if available\\.\n\nIf it is not available it tries to create the EEP\\-48 style docs using edoc\\.\n",
    "kind": "markdown"
  }
}
```

To produce that documentation, Erlang LS first checks whether a [doc chunk](https://www.erlang.org/doc/apps/kernel/eep48_chapter.html) is available. If not, it uses the [edoc](https://www.erlang.org/doc/apps/edoc/chapter.html) application to build the documentation. This last case can be very expensive, especially for large modules, and it is repeated for every hover request.

One important consideration to make is that, as soon as the user moves away from the function, a [$/CancelRequest](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#cancelRequest) is sent to the server and the process producing the documentation is killed.

### What is Required

Erlang LS could [memoize](https://en.wikipedia.org/wiki/Memoization#:~:text=In%20computing%2C%20memoization%20or%20memoisation,the%20same%20inputs%20occur%20again.) documentation requests and responses. Using a `gen_server` or an `ETS table` as a cache, the language server could store results from previous documentation requests and return them from the cache instead of recalculating docs every time.

The caching should still happen, even in case of a cancellation. Instead of killing the process, the server would let it finish and store the results in the cache.

Of course it will be important to invalidate entries from the cache whenever the file containing the function documentation is modified.

It should possible to enable/disable memoization via a configuration setting.

### Code Pointers

* Hover requests are dispatched to the `els_hover_provider` module
* The logic to produce documentation is contained in the `els_docs` module
* Configuration settings are handled via the `els_config` module

# Testing

Don't forget to include a test for the new functionality! You can look at existing examples in the `els_definition_SUITE` module and at this [code lenses tutorial](https://erlang-ls.github.io/articles/tutorial-code-lenses) which includes a [testing overview](https://erlang-ls.github.io/articles/tutorial-code-lenses/#adding-tests) for Erlang LS.

jdamanalo:
Closes #1208, and a little more. Since the memoization is done in `els_docs`, the change should also help `els_signature_help_provider` and `els_completion_provider`. Should tests be written here for the integration with the 2?

Also, there is a slight difference between this PR and #1208 in that there is a small amount of time when the memoization can be cancelled. I wasn't able to think of a solution without changing `els_background_job`.

### DIFF
--- a/apps/els_core/src/els_config.erl
+++ b/apps/els_core/src/els_config.erl
@@ -167,6 +167,7 @@ do_initialize(RootUri, Capabilities, InitOptions, {ConfigPath, Config}) ->
     RefactorErl = maps:get("refactorerl", Config, notconfigured),
     Providers = maps:get("providers", Config, #{}),
     EdocParseEnabled = maps:get("edoc_parse_enabled", Config, true),
+    DocsMemo = maps:get("docs_memo", Config, false),
 
     %% Initialize and start Wrangler
     case maps:get("wrangler", Config, notconfigured) of
@@ -220,6 +221,7 @@ do_initialize(RootUri, Capabilities, InitOptions, {ConfigPath, Config}) ->
     ok = set(plt_path, DialyzerPltPath),
     ok = set(code_reload, CodeReload),
     ok = set(providers, Providers),
+    ok = set(docs_memo, DocsMemo),
     ?LOG_INFO("Config=~p", [Config]),
     ok = set(
         runtime,
--- a/apps/els_lsp/priv/code_navigation/src/docs_memo.erl
+++ b/apps/els_lsp/priv/code_navigation/src/docs_memo.erl
@@ -1,0 +1,6 @@
+-module(docs_memo).
+
+-type type() -> any().
+
+-spec function() -> ok.
+function() -> ok.
--- a/apps/els_lsp/src/els_db.erl
+++ b/apps/els_lsp/src/els_db.erl
@@ -31,7 +31,8 @@ tables() ->
         els_dt_document,
         els_dt_document_index,
         els_dt_references,
-        els_dt_signatures
+        els_dt_signatures,
+        els_docs_memo
     ].
 
 -spec delete(atom(), any()) -> ok.
--- a/apps/els_lsp/src/els_docs.erl
+++ b/apps/els_lsp/src/els_docs.erl
@@ -105,26 +105,7 @@ docs(_M, _POI) ->
 function_docs(Type, M, F, A) ->
     case edoc_parse_enabled() of
         true ->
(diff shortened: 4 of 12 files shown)


jdamanalo:
Fixed dialyzer checks on older OTP.

plux:
Hi @jdamanalo , thanks for submitting this PR and sorry for the long response time.
This looks like an excellent addition, however it seems like there's conflicts that needs to be resolved.

Could you please rebase on main and resolve the conflicts? Thank you!

plux:
Thanks for your contribution @jdamanalo !

Seems like CI is a bit salty, I will merge this anyway.

PR #1418, opened 2023-02-01
Respect appdirs when prioritizing module candidates

keynslug:
### Description

This might be helpful in situations when tooling (e.g. `rebar3`) makes "shadow copies" in the indexable project directories. This might happen when underlying filesystem lacks sound symlink support, or (in my case) when slightly patched `rebar3` is used.

Before this change files having `_build/` in filepath were preferred over those having `apps/` or `src/` for example.


plux:
Thanks for your contribution @keynslug !

PR #1421, opened 2023-02-21
[DAP] Prioritize information from module_info when available

robertoaloi:
### Description

Certain build systems (most notoriously [Buck2)](https://buck2.build/) store build artifacts (including a copy of the source files) in custom folders (e.g. `buck-out` in Buck2). This can lead to an incorrect path to be returned by the debugger to the client whenever a breakpoint triggers. In turn, this result in the editor opening a buffer from the build system custom folder rather than the original source file and confusion among the users.

This PR changes the logic how the location of source files is identified. The current logic is:

1. Ask the interpreter first (can return an incorrect path)
2. If not found, ask the code server
4. Only as a last resort, try reading the `compile` info from `module_info`.

The `module_info` is supposed to contain the most accurate information, but - in the case of non hermetic builds - it can contain a path to a non-existing file (e.g. if the code was built on a machine and ported to a different one). That's the main reason the current logic looks like the above (see [this PR](https://github.com/erlang-ls/erlang_ls/pull/1084) for context).

Instead:

1. Check whether the `compile` info from `module_info` points to a regular file
2. If it does, use it (eventually converting it to an absolute path)
3. If it does not, ask the interpreter
4. If not found, ask the code server

The logic has been refactored into separate functions to ease reading. Debug logs have also been added to simplify eventual troubleshooting.

### DIFF
--- a/apps/els_dap/src/els_dap_rpc.erl
+++ b/apps/els_dap/src/els_dap_rpc.erl
@@ -74,26 +74,41 @@ eval(Node, Input, Bindings) ->
         {badrpc, Error} -> Error
     end.
 
--spec file(node(), module()) -> file:filename().
+-spec file(node(), module()) -> {ok, file:filename()} | {error, not_found}.
 file(Node, Module) ->
-    MaybeSource =
-        case rpc:call(Node, int, file, [Module]) of
-            {error, not_loaded} ->
-                BeamName = atom_to_list(Module) ++ ".beam",
-                case rpc:call(Node, code, where_is_file, [BeamName]) of
-                    non_existing -> {error, not_found};
-                    BeamFile -> rpc:call(Node, filelib, find_source, [BeamFile])
-                end;
-            IntSource ->
-                {ok, IntSource}
-        end,
-    case MaybeSource of
-        {ok, Source} ->
-            Source;
+    case file_from_int(Node, Module) of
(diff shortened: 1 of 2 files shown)


apps/els_dap/src/els_dap_rpc.erl
```diff
@@ -74,25 +76,71 @@ eval(Node, Input, Bindings) ->
...
+                        unsafe ->
+                            %% File is already absolute
+                            {ok, Source};
+                        RelativePath ->
+                            Joined = filename:join(Cwd, RelativePath),
+                            ?LOG_DEBUG("Composed Absolute Path as: [~p]", [Joined]),
+                            {ok, Joined}
+                    end;
+                false ->
+                    %% Build is not hermetic
```
TheGeorge, line 141: it might be hermetic, e.g. with +deterministic you get only the base name and is_regular will fail.

TheGeorge (approved):
Looks good, there is just a nit in one of the comments

apps/els_dap/src/els_dap_rpc.erl
robertoaloi: When I tried the `+determinsitc` option on OTP 25 I didn't get a `source` entry at all, so that case would be handled in the `undefined` clause above. But I can remove the comment.

Issue #1044, opened 2021-06-23
Cowboy behaviour undefined


Katur7:
Erlang LS gives me a warning that behaviour cowboy_rest is undefined.
Even if I include "_build/default/lib" in include_dirs in erlang_ls.config.

If I also include "_build/default/lib/cowboy/src" in include_dirs in erlang_ls.config I get the warning:
Issue in included file (18): behaviour cowboy_sub_protocol undefined

I made this small repo to reproduce the issue: [Repo](https://github.com/Katur7/erlang-ls-cowboy-issue)

**Expected behavior**
I expected the cowboy_rest behaviour to be found by default.

**Actual behavior**
The cowboy_rest behaviour is not found

**Context**
 - `erlang_ls` version (tag/sha): 0.17.0
 - Editor used: VS Code 1.57.1
 - LSP client used: erlang-ls.erlang-ls v0.0.27

plux:
It's actually complaining about the `cowboy_sub_protocol` not being found in `cowboy_rest`.
This is similar to this issue:
https://github.com/erlang-ls/erlang_ls/issues/1024

dweinstein:
I worked around this for now by adding:

```
code_path_extra_dirs:
  - "./_build/default/lib/*/ebin/"
```

to my `erlang_ls.config`.

Katur7:
Thank you @dweinstein. That works for me also :)

dweinstein:
hi @plux was this issue fixed by some code change or is the workaround still best to use? I noticed you marked this ticket as completed so I was just curious if I should try updating my `erlang_ls.config`. Thanks for the work you're doing!

plux:
I think this issue still exists and we should investigate it further.
I will open it again and.

eproxus:
Unfortunately that workaround doesn't work for me with the behavior `cowboy_middleware`.

PR #1423, opened 2023-03-04
Make code_reload act regardless of diagnostics

### LINKED ISSUE #1409: code reload should be triggered despite compilation warnings
I noticed that code reload isn't attempted when there are compilation warnings. Given hot-reload is (hopefully!) a development workflow, I think warnings should be ignored, at least as an option. The most common situation, at least in my case, is that I simply have some unused code. I'd like to still be able to quickly test the parts that are not unused. :)

Would you agree with this?

Thanks a lot for your work!

fabjan:
### Description

See #1409 for background. Given that hot-reloading from an editor is a development feature it should not matter if there are diagnostic errors while iterating.

The reason I moved the code out from `els_compiler_diagnostics` is that it seemed a bit misplaced in there and looking at commit history it looks like code reload was added before the `els_diagnostics` behavior existed, and it was just inside `els_text_synchronization_provider`. If it is to be detached from diagnostics results it seems there is little reason to keep it in `els_compiler_diagnostics`.

Fixes #1409 

### Extras

Stop trying to compile and reload files that are not ".erl" files. It doesn't work and just gives popup spam.

Skärmavbild 2023-03-05 kl  08 59 08

### Testing

The moved test cases work fine. I also manually tested that code reloading still works.

### Notes

If the remote node has a cookie it is not possible to configure code reload to work with it. Perhaps the "code_reload" section in the config file needs to get a "cookie" attribute (but for a future PR I guess).

### DIFF
--- a/apps/els_lsp/src/els_code_reload.erl
+++ b/apps/els_lsp/src/els_code_reload.erl
@@ -1,0 +1,53 @@
+-module(els_code_reload).
+
+-include("els_lsp.hrl").
+-include_lib("kernel/include/logger.hrl").
+
+-export([
+    maybe_compile_and_load/1
+]).
+
+-spec maybe_compile_and_load(uri()) -> ok.
+maybe_compile_and_load(Uri) ->
+    Ext = filename:extension(Uri),
+    case els_config:get(code_reload) of
+        #{"node" := NodeStr} when Ext == <<".erl">> ->
+            Node = els_utils:compose_node_name(
+                NodeStr,
+                els_config_runtime:get_name_type()
+            ),
+            Module = els_uri:module(Uri),
+            case rpc:call(Node, code, is_sticky, [Module]) of
+                true -> ok;
+                _ -> handle_rpc_result(rpc:call(Node, c, c, [Module]), Module)
+            end;
+        _ ->
+            ok
+    end.
+
+-spec handle_rpc_result(term() | {badrpc, term()}, atom()) -> ok.
+handle_rpc_result({ok, Module}, _) ->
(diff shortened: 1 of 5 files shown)


dweinstein:
I tend to agree with @fabjan here. It's weird to get spurious errors like "code_reload swap crashed for: erlang_ls.config with: error" when opening/saving the `erlang_ls.config` file or `*.app.src` files. 

I'd say either this change or an option in `erlang_ls.config` so that if I hit save on a file it tries to reload regardless of anything else such as warnings. I've found that it's quite confusing that, e.g., I have this issue https://github.com/erlang-ls/erlang_ls/issues/1044 and due to that it wouldn't `code_reload` my cowboy handler.

image

dweinstein:
separately @fabjan 

> If the remote node has a cookie it is not possible to configure code reload to work with it. Perhaps the "code_reload" section in the config file needs to get a "cookie" attribute (but for a future PR I guess).

putting the cookie into the `erlang_ls.config` doesn't sound like a good idea since that would be committed? Seems like something to pick up from another file or environment variable instead?

fabjan:
> putting the cookie into the erlang_ls.config doesn't sound like a good idea since that would be committed? Seems like something to pick up from another file or environment variable instead?

That sounds like a better idea. There is already a cookie setting in the config, so I figured this might have been considered already: https://github.com/erlang-ls/erlang_ls/blob/229175ec35afddbb5c5a0ac2cf25423d7aa0b6ab/apps/els_core/src/els_config_runtime.erl#L71

plux:
Thanks for your contribution @fabjan !

PR #1426, opened 2023-03-06
Idea: hrl don't have to warning unuse

DominicGame:
normally, include another `.hrl` in an `.hrl` file, don't have to warning unuse in this situation.
a.hrl
```
-include("aa.hrl").
-include("aaa.hrl").
```
a.erl
```
-include("a.hrl").
```


plux:
Thanks for your contribution @DominicGame !

PR #1430, opened 2023-03-30
Add -hidden flag

mkuratczyk:
We saw situations where erlang_ls interfered with the application being developed through `global`. We had a locking issue that only occurred when the editor with erlang_ls was running. Adding `-hidden` resolves the problem for us.


plux:
Thanks for your contribution @mkuratczyk !

PR #1433, opened 2023-05-26
[DAP] Use cwd from launch config, not from project node

robertoaloi:
### Description

* When extracting the file path in the `file_from_module_info` function, use `cwd` from launch config, not the one from the project node (that could be started in a different dir)
* When checking for existence of relative paths, ensure you use the correct `cwd`


alanz:
LGTM

PR #1441, opened 2023-06-29
Lowercase drive letter in URIs for Windows

corben2:
### Description

If a LSP client on Windows sends a capital drive letter in a URI, then the compiler diagnostics end up sending the wrong position/range. 

This is because of [this line.](https://github.com/erlang-ls/erlang_ls/blob/e91c9519996e0ff701edc19f308524906a3df131/apps/els_lsp/src/els_compiler_diagnostics.erl#L195) It checks that the path from the compiler matches the URI path. The path from the compiler has a lowercase drive letter, so they don't match and the diagnostics are treated as if they are coming from an include.

I've fixed this in els_uri:path, but there might be a better place to do it, I'm not sure.


plux:
Thanks for your contribution @corben2 !

PR #1446, opened 2023-08-04
Add github release.yml to publish binary artifacts

alanz:
So they can be downloaded by external clients, such as lsp-erlang.

See https://github.com/erlang-ls/erlang_ls/actions/runs/5762933437 for a successful run, and
https://github.com/erlang-ls/erlang_ls/releases/tag/0.48.1-test3 for the release with artefacts.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,193 @@
+name: Release
+on:
+  release:
+    types:
+      - created
+
+# Test trigger. Uncomment to test basic flow.
+#   NOTE: it will fail on trying to get the release url
+# on:
+#   push:
+#     branches:
+#       - '*'
+
+jobs:
+  linux:
+    strategy:
+      matrix:
+        platform: [ubuntu-latest]
+        otp-version: [23, 24, 25]
+    runs-on: ${{ matrix.platform }}
+    container:
+      image: erlang:${{ matrix.otp-version }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,196 @@
+name: Release
+on:
+  release:
+    types:
+      - created
+      # Temporary, for testing
+      - edited
+      - published
+
+# Test trigger. Uncomment to test basic flow.
```
robertoaloi, line 10: Leftover?

.github/workflows/release.yml
```diff
@@ -0,0 +1,196 @@
+name: Release
+on:
+  release:
+    types:
+      - created
+      # Temporary, for testing
```
robertoaloi, line 6: Leftover?

robertoaloi (approved):
Looks good apart from the leftovers.

.github/workflows/release.yml
alanz: No, Kept as a prompt for the next unlucky person who needs to modify this flow, to get the basics right without having to disturb other people with test releases

.github/workflows/release.yml
```diff
@@ -0,0 +1,193 @@
+name: Release
+on:
+  release:
+    types:
+      - created
```
robertoaloi, line 5: For VS Code, I had `published` here:

https://github.com/erlang-ls/vscode/blob/main/.github/workflows/publish.yml#L4

PR #1449, opened 2023-08-22
[LSP] Fix initialize on OTP 23.0/23.1

danielfinke:
### Description

`els_uri:path/2` was calling `uri_string:percent_decode/1`, which is unavailable until OTP 23.2
(https://www.erlang.org/doc/man/uri_string#percent_decode-1)

### DIFF
--- a/apps/els_core/src/els_uri.erl
+++ b/apps/els_core/src/els_uri.erl
@@ -5,6 +5,10 @@
 %%==============================================================================
 -module(els_uri).
 
+-if(?OTP_RELEASE =:= 23).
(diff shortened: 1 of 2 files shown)


robertoaloi:
Hi @danielfinke and thanks for this. It seems the OTP 23 CI still breaks on a XRef error.

danielfinke:
@robertoaloi I think I have addressed the xref error (as well as a Dialyzer warning when on OTP 23.0/23.1)

plux:
Thanks for your contribution @danielfinke !

PR #1454, opened 2023-10-23
added nonempty_binary and nonempty_bitstring

MarkoMin:
### Description

Added types were introduced in OTP24, but since macro FEATURE_AVAILABLE (introduced in OTP25) is available, I think those types should be too.


plux:
Thanks for you contribution @MarkoMin !

PR #1460, opened 2023-11-06
Makes 'find_references/2' faster

shuying2244:
Fixed the way Kind equals `record`/`record_def_field`/`define` works in `find_references/2`, this `uri_pois_to_locations(find_scoped_references_for_def(Uri, Poi));` is really too slow, now they can use `find_references_for_id(Kind, Key);` like a function.

### DIFF
--- a/apps/els_lsp/src/els_dt_document.erl
+++ b/apps/els_lsp/src/els_dt_document.erl
@@ -286,6 +286,8 @@ get_words(Text) ->
             Fun = fun
                 ({atom, _Location, Atom}, Words) ->
                     sets:add_element(Atom, Words);
+                ({var, _Location, Var}, Words) ->
+                    sets:add_element(Var, Words);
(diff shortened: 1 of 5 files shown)


plux (changes requested):
Thanks for your contribution, however I can't accept this PR as it will make erlang_ls behave incorrectly.

Records and macros can be specified in more than one file and are not globally unique, so we can't simply index them without taking into account in what file they are located and where that file is included.

Take for example the common use case of having a `#state{}` record, this can potentially be specified in many places, but when we want to look up references for a certain `#state{}` record we only want to find the places that can actually see that specific `#state{}` record.

I however like the idea to try to add some kind of indexing to speed this up, maybe we can figure out a way to make this fast while also being correct.

plux:
This PR should make find_references faster, without sacrificing correct behavior.
https://github.com/erlang-ls/erlang_ls/pull/1476

PR #1463, opened 2023-11-19
Add goto definition for implemented behaviour callbacks

fridayy:
### Description

Currently `erlang_ls` supports navigation from a behaviour callback to its implementation. It is not possible to navigate back from the implementation to the definition to the callback using goto definition. This PR adds this support.

Looking forward hearing some feedback on this. Have a nice day!


fridayy:
Hi, 
although I haven't touched `apps/els_core/src/els_dodger.erl` shall I run the formatter and fix this?

fridayy:
Used erlfmt 1.2.0 locally but CI already runs with v.1.3.0. That's the issue. The file should be properly formatted now.
TIL: Update local plugins regularly

plux (approved):
Thanks for your contribution!

Nice idea, code looks good!

PR #1475, opened 2024-01-08
`all` target - building as dap not required

MarkoMin:
### Description

I think that it's not necessary anymore to have this build step.


plux:
@robertoaloi Seems ok?

PR #1499, opened 2024-02-16
Cleanup unused includes/macros (as identified by ELP)

robertoaloi:
### Description

This is the result of running:

```
elp lint --diagnostic-filter W0020 --apply-fix --in-place
elp lint --diagnostic-filter unused_macro --apply-fix --in-place
```


apps/els_lsp/src/els_group_leader_server.erl
```diff
@@ -38,7 +38,6 @@
 %%==============================================================================
 %% Macro Definitions
 %%==============================================================================
--define(SERVER, ?MODULE).
```
plux, line 41: I think we should update the code to use ?SERVER.

plux (approved):

plux:
Good stuff! :+1:

apps/els_lsp/src/els_group_leader_server.erl
robertoaloi: Fair point, fixed.

PR #1505, opened 2024-04-24
nifs attribute completion

MarkoMin:
### Description

Adds completion for [-nifs attribute](https://www.erlang.org/doc/reference_manual/modules#pre-defined-module-attributes) officially supported from OTP25

EDIT: dont merge yet, I have to fix completion inside "-nifs" attribute to behave similarly as inside "-export"

### DIFF
--- a/apps/els_core/src/els_poi.erl
+++ b/apps/els_core/src/els_poi.erl
@@ -45,6 +45,8 @@
     | include_lib
     | macro
     | module
+    | nifs
+    | nifs_entry
     | parse_transform
     | record
     | record_def_field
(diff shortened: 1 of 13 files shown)


MarkoMin:
I pushed the solution, but tests are missing ATM. It's not as elegant as I hoped, but did my best. It doesn't exclude functions already defined as `-nifs`. I tried to do it, but ended up too deeply in els code. If you have an idea how do to this more elegant, I'm all ears.

Will add tests when I find more time

plux:
I tried out the feature locally and it works fine.

Seems like there some formatting issues.
Please run `rebar3 fmt` and commit the changes.

If you find time to write some tests that would also be awesome :)

plux:
I have opened a new PR for this here: https://github.com/erlang-ls/erlang_ls/pull/1537

PR #1507, opened 2024-04-29
Upgrade elvis_core to 3.2.2

rlipscombe:
The version of elvis_core included in erlang_ls is kinda old, and doesn't support some of the newer rules. So I upgraded it.


plux:
Seems like it fails to compile on OTP 23.
I think it is probably time to drop support for OTP 23, this means updating the vscode-extension to use a newer erlang version as well.

plux:
We have now dropped OTP 23 support. Would you please rebase on latest main?

plux:
This is now handled in https://github.com/erlang-ls/erlang_ls/pull/1522

PR #1513, opened 2024-05-04
Avoid running get_words twice when opening new file

ztion:
### Description

Small fix to make a small performance improvement when opening a file the first time.
Avoids running get_words twice from did_open, and also makes the set created in get_words use version 2 which is a bit more efficient, supported from OTP24.

On a big file like erlfmt_parse.erl on my computer the time spent went from about 960ms to ~850ms.


plux:
Thanks for you contribution @ztion !

PR #1515, opened 2024-05-05
Change linum-mode to display-line-numbers-mode

### LINKED ISSUE #1514: Error void-function linum-mode on Emacs
**Describe the bug**
After a lot of warnings when installing from the description of https://erlang-ls.github.io/getting-started/emacs/
(don't know if the many warnings while are considered a bug too, let me know if I should report them).

After opening the first Erlang file I get the following error:

`File mode specification error: (void-function linum-mode)`

It looks like it works but I don't know if some things didn't run because of this error.

**To Reproduce**

On MacOS 14.1 MacBook Pro with M3 chip with a pretty recent official Emacs for MacOS: Version 29.2 (9.0)

Copy the .emacs content into a otherwise empty one from here: https://github.com/erlang-ls/erlang_ls/blob/main/misc/dotemacs

Start emacs: it downloads and compiles everything with a ton of warnings.

Then open first *.erl file in a project.

**Expected behavior**

No error message.

**Actual behavior**

Error message:  `File mode specification error: (void-function linum-mode)`

**Context**
 - `erlang_ls` version (tag/sha):  ? see below
 - Editor used: Emacs 29.2
 - LSP client used: ? see below

```
  erlang                         20240429.1402  installed             Erlang major mode
  helm-lsp                       20210419.2014  installed             LSP helm integration
  lsp-mode                       20240502.1852  installed             LSP mode
  lsp-origami                    20230815.704   installed             origami.el support for lsp-mode
  lsp-ui                         20240424.412   installed             UI modules for lsp-mode
```

peerst:
### Description

link-mode no longer supported and replaced by a built in functionality.

Fixes #1514

### DIFF
--- a/misc/dotemacs
+++ b/misc/dotemacs
@@ -41,7 +41,7 @@
 (setq lsp-log-io t)
 
 ;; Show line and column numbers
-(add-hook 'erlang-mode-hook 'linum-mode)
+(add-hook 'erlang-mode-hook #'display-line-numbers-mode)
 (add-hook 'erlang-mode-hook 'column-number-mode)
 
 ;; Enable and configure the LSP UI Package


plux:
Thanks for your contribution @peerst

PR #1528, opened 2024-08-22
Add 'TEST' macro to options when compile for reload

zsoci:
### Description

Handle {d, 'TEST'} format when recompiling.


plux:
Thanks for your contribution @zsoci

PR #1530, opened 2024-09-13
OTP27 support

### LINKED ISSUE #1489: OTP 27 support
There will (very probably) be new attributes in OTP27: `-doc` and -`moduledoc`. They'll accept string, tuple, atom or map (with some reserved keys) which we could autocomplete.

There will also be rules like: `-doc` with string can only appear once

I'd want to work on this because I want to get into this project more :D

Release is still 4 months away, but I put this here as a remainder

MarkoMin:
### Description

Things done:
1. Completion for `-doc` and `-moduledoc` (+ tests)
2. Bumped to new version of `erlfmt` which fully supports OTP27
3. Removed some OTP23-compatibility code

Things to be done:
1. Sigils completion
2. Include OTP27 in CI

Things I need help with:
1. Fixing `els_diagnostics_SUITE` - those even fail for me when I run them on current `main` branch with OTP25 (tescases fail due to "can't find include file" error, while test cases expect them to be found)
2. Fixing `els_completion_SUITE` - some functions fail when comparing documentation (e.g. `application_remote_otp`- Seems like only signature is fetched as a markdown in OTP27, dunno exactly why. Any help here is very welcome

Fixes #1489.


plux:
@MarkoMin 

Fix for the workflows is merged now. 
Please re-base on main so we can re-run the tests.

MarkoMin:
seems like there is some issue with artifact caching in the workflow. Can you restart it?

plux:
Sorry, can you rebase again? It should be fixed now.

plux:
Thanks for you contribution @MarkoMin

PR #1531, opened 2024-09-13
Fix false positive in maybe expression

### LINKED ISSUE #1527: Maybe/else pattern erroneously triggering 'bound variable in pattern' error
**Describe the bug**
A `Bound variable in pattern: Var` error is generated for maybe expressions that bind a variable in their `else` patterns:

```erlang
foo() ->
    maybe
        foo ?= bar()
    else
        e = Error -> Error
          % ^- Bound variable in pattern: Error
    end.
```

**To Reproduce**
Use the code above.

**Expected behavior**
No 'bound variable in pattern' hint is generated.

**Actual behavior**
A 'bound variable in pattern' hint is generated even though the variable is not already bound.

**Context**
 - `erlang_ls` version (tag/sha): `Version: 0.52.0+build.1987.ref8700e96`
 - Editor used: Zed v0.139.3
 - LSP client used: Zed v0.139.3

MarkoMin:
### Description

Here is just the test case that confirms it. Hope I'd manage to get some time to write a solution.

Fixes #1527


MarkoMin:
Just discovered that not only `else` variables are false positives, but variables in `maybe` body are false negatives. I've updated an existing test to cover that.

MarkoMin:
Turns out we'll have to wait for new OTP release to fix it. How should we this in lower OTP versions?

MarkoMin:
PR got merged in OTP so I will test this one the new release is out. This should work for OTP > 27.2. I have no idea if we can fix this for earlier versions of OTP

PR #1537, opened 2024-09-18
nifs attribute completion

plux:
### Description

Adds completion for [-nifs attribute](https://www.erlang.org/doc/reference_manual/modules#pre-defined-module-attributes) officially supported from OTP25

### DIFF
--- a/apps/els_lsp/src/els_rename_provider.erl
+++ b/apps/els_lsp/src/els_rename_provider.erl
@@ -122,7 +122,8 @@ workspace_edits(Uri, [#{kind := Kind} = POI | _], NewName) when
     Kind =:= export_entry;
     Kind =:= import_entry;
     Kind =:= export_type_entry;
-    Kind =:= type_application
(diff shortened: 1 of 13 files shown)


apps/els_lsp/src/els_rename_provider.erl
```diff
@@ -205,7 +206,8 @@ editable_range(#{kind := Kind, data := #{name_range := Range}}, function) when
     Kind =:= export_type_entry;
     Kind =:= import_entry;
     Kind =:= type_application;
-    Kind =:= type_definition
+    Kind =:= type_definition,
```
MarkoMin, line 209:
```suggestion
    Kind =:= type_definition;
```
took me a while...

plux: Oh wow haha! that was a subtle bug. Good thing we got tests :)

plux:
Thanks for your contribution @MarkoMin

